PR #32, opened 2025-01-29
[e2e] Run e2e tests parallel

v-scharf:
cucumber can run tests parallel using `--parallel`

We ran the tests consistently because we used the same users (Alice, Brian...), had conflicts when creating users/sharing resources with users. I now create users and groups with a unique prefix for each test to avoid name conflict. example: `Brian-04f` 

how to tests: run opencloud with `OC_SHOW_USER_EMAIL_IN_RESULTS=true`

run command: `PARALLEL=4 OC_BASE_URL=opencloud-server:9200 RETRY=1 HEADLESS=true ./run-e2e.sh --suites admin-settings,file-action,journeys,navigation,search,shares,spaces,user-settings`


tests/e2e/support/objects/app-files/share/collaborator.ts
```diff
@@ -98,9 +98,13 @@ export default class Collaborator {
     } = args
     const collaboratorInputLocator = page.locator(Collaborator.inviteInput)
     await collaboratorInputLocator.click()
+
+    // use unique email for user and unique display name for group
+    const collaboratorIdentifier =
+      'email' in collaborator ? collaborator.email : collaborator.displayName
```
v-scharf, line 104: I cannot search users by `onPremisesSamAccountName` as it turns out, it doesn't work https://github.com/opencloud-eu/opencloud/issues/144

tests/e2e/config.js
```diff
@@ -19,8 +19,7 @@ export const config = {
     return withHttp(this.keycloakHost + '/admin/master/console')
   },
   // ocm config
-  federatedbaseUrlOpenCloud:
-    process.env.FEDERATED_BASE_URL_OPEN_CLOUD ?? 'federation-open-cloud:10200',
+  federatedbaseUrlOpenCloud: process.env.OC_FEDERATED_BASE_UR ?? 'federation-open-cloud:10200',
```
JammingBen, line 22:
```suggestion
  federatedbaseUrlOpenCloud: process.env.OC_FEDERATED_BASE_URL ?? 'federation-open-cloud:10200',
```

Where is this URL coming from anyway? 🤔 I couldn't find any occurrences.

tests/e2e/cucumber/environment/index.ts
```diff
@@ -77,6 +78,8 @@ Before(async function (this: World, { pickle }: ITestCaseHookParameter) {
       }
     }
   }
+  // this.uniquePrefix = uuidv4().substring(0, 8)
```
JammingBen: Can we remove this?

tests/e2e/package.json
```diff
@@ -5,6 +5,7 @@
     "join-path": "1.1.1",
     "lodash-es": "4.17.21",
     "luxon": "3.5.0",
-    "xml-js": "^1.6.11"
+    "xml-js": "^1.6.11",
+    "uuid": "10.0.0"
```
JammingBen: Please make it a version range (also there is version 11 available).

```suggestion
    "uuid": "^11.0.0"
```

tests/e2e/support/types.ts
```diff
@@ -26,6 +26,7 @@ export interface User {
    * actual id, that will be exposed by the graph api
    */
   uuid?: string
+  userName: string
```
JammingBen: Ultra nitpicking: Technically, `username` would be correct (no capital `N`), because it's one word (`username` and not `user name`) 🤓  Feel free to leave it though, I don't have a strong opinion on this.

tests/e2e/cucumber/environment/index.ts
v-scharf: sure

tests/e2e/config.js
v-scharf: I didn't found it in drone.star or testing docs
I suppose it might be necessary to run tests locally. I haven't tested the ocm yet. I suggest we leave it at that.

tests/e2e/support/types.ts
v-scharf: fixed

PR #37, opened 2025-02-03
Fix: Hidden tooltip in collapsed sidebar

JammingBen:
Removes the tooltip in the collapsed sidebar as it was almost completely hidden, leading to a visual glitch on hover. Replaces it with a proper `aria-label` to make it a11y conform.

Before:

Pasted Graphic

After:

Pasted Graphic 1

### DIFF
--- a/packages/web-runtime/src/components/SidebarNav/SidebarNavItem.vue
+++ b/packages/web-runtime/src/components/SidebarNav/SidebarNavItem.vue
@@ -1,13 +1,15 @@
 <template>
   <li class="oc-sidebar-nav-item oc-pb-xs oc-px-s" :aria-current="active ? 'page' : null">
     <oc-button
-      v-oc-tooltip="toolTip"
       :type="handler ? 'button' : 'router-link'"
       appearance="raw"
       :variation="active ? 'primary' : 'passive'"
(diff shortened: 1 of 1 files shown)


packages/web-runtime/src/components/SidebarNav/SidebarNavItem.vue
```diff
@@ -1,13 +1,15 @@
...
   <li class="oc-sidebar-nav-item oc-pb-xs oc-px-s" :aria-current="active ? 'page' : null">
     <oc-button
-      v-oc-tooltip="toolTip"
       :type="handler ? 'button' : 'router-link'"
       appearance="raw"
       :variation="active ? 'primary' : 'passive'"
       :class="['oc-sidebar-nav-item-link', 'oc-oc-width-1-1', { active: active }]"
       :data-nav-id="index"
       :data-nav-name="navName"
+      :aria-label="
```
kulmann, line 10: Why don't we keep the same aria-label for both the collapsed and the expanded state?

JammingBen: Because with the expanded state we have a `span` with the label visible. And having both a visible label and an `aria-label` are bad practice if I recall correctly, no? 🤔

Issue #42, opened 2025-02-05
App Token UI


kulmann:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

As a user, I want to easily manage my app tokens so that I am enabled to use non-oidc third party applications easily.

### User Value

Makes app token management available to non-tech-savvy users (= no system / cli access and API usage too complicated).

### Acceptance Criteria

- new extension point for account page panels `preferences-panel`
- extensions of this extension point get rendered between `Extensions` and `GDPR` panels
- new extension (registered by the `runtime` package) with an app token preferences panel
- panel:
  - button for creating a new app token (via modal). modal shows expiration input field and cancel and submit buttons.
  - lists all existing app tokens, ordered by creation date. no pagination needed (at least not in the first iteration of the feature)
  - each app token has a delete button for deleting the respective token. Approving via modal is required.
  - each app token has a copy button for copying the respective token into the system clipboard.
- unit tests for creation (incl. test for validation failure on negative expiration durations), listing, deletion, error handling (creation or deletion failed)
- new extension point is documented in the dev docs

PR #43, opened 2025-02-05
App tokens preferences panel

JammingBen:
### Description

Introduces a new extension point for user preference panels and implements a panel for app tokens.

image

### Setup

Add the following config to your `docker-compose.override.yml`:

```
OC_ADD_RUN_SERVICES: auth-app
PROXY_ENABLE_APP_AUTH: true
```

closes https://github.com/opencloud-eu/web/issues/42

### DIFF
--- a/packages/web-runtime/src/components/Account/AppTokens.vue
+++ b/packages/web-runtime/src/components/Account/AppTokens.vue
@@ -51,7 +51,7 @@
             <span v-text="formatDateFromISO(item.created_date, currentLanguage)" />
           </div>
         </template>
-        <template #exprationDate="{ item }">
+        <template #expirationDate="{ item }">
           <div class="oc-width-1-1 oc-text-truncate">
             <span v-text="formatDateFromISO(item.expiration_date, currentLanguage)" />
           </div>
@@ -214,7 +214,7 @@ const tableFields = computed(() => {
       title: $gettext('Created at')
     },
     {
-      name: 'exprationDate',
+      name: 'expirationDate',
       type: 'slot',
       wrap: 'truncate',
       title: $gettext('Expires at')


packages/web-runtime/src/pages/account.vue
```diff
@@ -357,6 +367,21 @@ export default defineComponent({
...
+    const extensions: Extension[] = [
+      {
+        id: 'com.github.opencloud-eu.web.runtime.preferences-panels.app-tokens',
+        type: 'customComponent',
+        extensionPointIds: [preferencesPanelExtensionPoint.id],
+        content: AppTokens
+      } as CustomComponentExtension
+    ]
+
+    extensionRegistry.registerExtensions(computed(() => extensions))
```
kulmann: What's the reasoning to register the extension when visiting the account page (and deregistering it when leaving the account page), instead of doing that once during runtime initialization? All other apps and extensions get registered during app bootstrap at the moment, don't they?

packages/web-runtime/src/pages/account.vue
```diff
@@ -255,6 +255,11 @@
           </oc-td>
         </oc-tr>
       </account-table>
+      <component
```
kulmann: Made up my mind, could you please move the extensions below the gdpr export?

Also, unrelated but would still be helpful: could you add some margin bottom to all "panels" on the account page? I'm missing some white space above the headlines / in between panels.

packages/web-runtime/src/components/Account/AppTokens.vue
```diff
@@ -0,0 +1,248 @@
...
+        <oc-icon name="add" size="small" />
+        <span v-text="$gettext('New')" />
+      </oc-button>
+    </div>
+    <span
+      v-if="authAppServiceDisabled"
+      data-testid="auth-service-unavailable"
+      v-text="
+        $gettext(
+          'App tokens are not available because the auth app service is not running. Please contact an administrator of your instance.'
```
kulmann:
```suggestion
          'App tokens are not available because the »auth-app« service is not running. Please contact an administrator.'
```

packages/web-runtime/src/pages/account.vue
JammingBen: In the runtime, we seem to register extensions in `Application.vue` (https://github.com/opencloud-eu/web/blob/feat/app-tokens-preferences-panel/packages/web-runtime/src/layouts/Application.vue#L201). I can move it there, but I still need to unregister the extension on unmount.

I don't know why we're not doing this during the bootstrap process. Do you have any idea?

JammingBen: I now added the registration to the runtime bootstrap process, see latest commit. We could maybe do the same for the progress bars (https://github.com/opencloud-eu/web/blob/feat/app-tokens-preferences-panel/packages/web-runtime/src/layouts/Application.vue#L190)? Or is there any reason why this has been implemented in `Applications.vue`?

JammingBen: As discussed, all runtime extensions are now being registered during the app bootstrap.

packages/web-runtime/src/components/Account/AppTokens.vue
```diff
@@ -0,0 +1,250 @@
...
+      title: $gettext('App Token')
+    },
+    {
+      name: 'creationDate',
+      type: 'slot',
+      wrap: 'truncate',
+      title: $gettext('Created at')
+    },
+    {
+      name: 'exprationDate',
```
kulmann:
```suggestion
      name: 'expirationDate',
```

packages/web-runtime/src/components/Account/AppTokens.vue
```diff
@@ -0,0 +1,250 @@
...
+              <oc-icon name="file-copy" fill-type="line" />
+            </oc-button>
+          </div>
+        </template>
+        <template #creationDate="{ item }">
+          <div class="oc-width-1-1 oc-text-truncate">
+            <span v-text="formatDateFromISO(item.created_date, currentLanguage)" />
+          </div>
+        </template>
+        <template #exprationDate="{ item }">
```
kulmann:
```suggestion
        <template #expirationDate="{ item }">
```

PR #145, opened 2025-02-11
Add emblem, only use background image when set

AlexAndBear:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

### DIFF
--- a/packages/design-system/src/styles/theme/login.scss
+++ b/packages/design-system/src/styles/theme/login.scss
@@ -20,10 +20,9 @@
     bottom: -2vh;
     width: 20vw;
     height: 20vw;
-    z-index: 0;
 
     @media (max-width: 640px) {
-      display: none !important;
+      display: none;
     }
   }
 


packages/design-system/src/styles/theme/login.scss
```diff
@@ -14,6 +14,19 @@
     transform: translate(-50%, -150%);
   }
 
+  &-emblem {
+    position: fixed;
+    right: -4vw;
+    bottom: -2vh;
+    width: 20vw;
+    height: 20vw;
+    z-index: 0;
```
AlexAndBear: z-index can be removed

packages/design-system/src/styles/theme/login.scss
```diff
@@ -14,6 +14,19 @@
...
+
+    @media (max-width: 640px) {
+      display: none !important;
```
AlexAndBear: check if we need important

PR #158, opened 2025-02-12
chore: clean up design-system and and remove CollapsibleTable

JammingBen:
Cleans up the design-system dependencies after switching to VitePress, removes unused doc tokens and also removes the `CollapsibleTable` since it's also not being used anymore.


kulmann (approved):
nice, finally 🤩

PR #163, opened 2025-02-13
chore: remove component doc blocks

JammingBen:
Also refactors `OcTable` and `OcTableSimple` to script setup because those were to last two remaining components in the design system.


kulmann (approved):
Thanks for picking up the last bits - I missed those before 😅

Issue #137, opened 2025-02-10
Design System Documentation


kulmann:
## Initial Checklist

- [x] I understand this is a feature request and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

## Motivation / context

For the documentation of the design system we're currently using `vue-styleguidist`. This is mostly unmaintained and also fills up our Security issue tracker. Since the documentation is not published anywhere we want to get rid of it (edit: of `vue-styleguidist`, not of the documentation).

## Requirements

We have a very limited set of requirements at the moment. There are two main use cases for the documentation:
1. Component development within rendered examples (dev server)
2. Documentation for developers within the extension eco-system with the following main points of interest:
  a. Component features and how to use them
  b. CSS utility classes and how to use them
  c. Look up available icons
  d. (later on: how to use material color roles... if we ever switch to Material)

## Alternatives

There are several alternatives for our small set of requirements. The following is a non-exhaustive list of alternatives including some pros and cons respectively.

#### Pros

- actively maintained
- MIT licensed

#### Cons

- might be too bloated (e.g. visual tests are far beyond what we need)
- open core? what's the enterprise-featureset?

#### Pros
- actively maintained and officially supported by Vue
- seems easy to use as you can write simple markdown and use components within (https://vitepress.dev/guide/using-vue#using-components)

#### Cons

- needs some digging into to check if it fulfills all our needs and how to make it work

#### Pros

- dog-fooding our own design system
- quick start (-> no need to gain expert-knowledge in a documentation tool)
- 

#### Cons

- nothing auto-generated

JammingBen:
Closing here, the 2 remaining items have separate issues.

PR #172, opened 2025-02-13
chore(design-system): include icons in docs bundle

JammingBen:
As per title.


kulmann (approved):
nice 💪

PR #176, opened 2025-02-14
feat(runtime): rework app tokens

JammingBen:
Reworks the app tokens so a newly created token gets only shown once to the user. The user then can copy the token.

Follow-up after https://github.com/opencloud-eu/web/pull/43

refs https://github.com/opencloud-eu/web/issues/42


kulmann (approved):
Nice, thank you for the quick and flawless followup 🥇

PR #179, opened 2025-02-14
chore(design-system): add live doc block plugin

JammingBen:
Adds a live doc plugin. It can either include a local script or an external component:

~~~
::: livecode
```vue
<oc-application-icon icon="home" color-primary="#e2baff" color-secondary="#20434f" />
```
:::
~~~

~~~
::: livecode {path=/components/OcApplicationIcon/example.vue}
<<< @/components/OcApplicationIcon/example.vue
:::
~~~

Also moves each doc component file into its dedicated folder. This allows for a better structure since related files such as code examples can be placed in the same folder.


kulmann (approved):
so awesome! 🤩 🎖️

PR #184, opened 2025-02-17
fix(ods): model and event binding on textarea and input

kulmann:
## Description
Turns out that moving everything to `v-model` is too strict / breaking with our existing usage. We need `input` events from the text input component, so I added them back to the `OcTextInput` and `OcTextarea` components. We might also want to bind `value` props to the model as well... 🤔 

## Types of changes

- [x] Bugfix

## Open tasks
- [ ] fine tune OcTextInput docs
- [ ] come up with OcTextarea docs


JammingBen (approved):
Nice 🚀

PR #189, opened 2025-02-18
docs(design-system): add docs for components c-f

JammingBen:
As per title.

### DIFF
--- a/packages/design-system/docs/components/OcCheckbox/OcCheckbox.md
+++ b/packages/design-system/docs/components/OcCheckbox/OcCheckbox.md
@@ -1,0 +1,37 @@
+# OcCheckbox component
+
+## Description
+
+The `OcCheckbox` component represents a simple checkbox input element that can either be checked or unchecked.
+
+## Accessibility
+
+The `label` is required and represents the name of the input. If `label-hidden` is set to `true`, the label will be hidden from the screen but still be available for screen readers via the `aria-label` attribute.
+
+## Examples
+
+### Default
+
+The default use case needs a `label` and a `v-model` to bind the value of the checkbox.
+
+::: livecode {path=/components/OcCheckbox/default.vue}
+<<< @/components/OcCheckbox/default.vue
+:::
+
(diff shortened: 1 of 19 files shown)


packages/design-system/docs/components/OcCheckbox/OcCheckbox.md
```diff
@@ -0,0 +1,37 @@
...
+## Examples
+
+### Default
+
+The default use case needs a `label` and a `v-model` to bind the value of the checkbox.
+
+::: livecode {path=/components/OcCheckbox/default.vue}
+<<< @/components/OcCheckbox/default.vue
+:::
+
```
kulmann, line 20: We should have two `v-model` examples, one for a singular value model and one for an array model. Should that be another example section?

kulmann (approved):
So nice to see more and more components documented. And the live code block component really pays off... thank you! 🏅

packages/design-system/src/components/OcDropzone/OcDropzone.vue
```diff
@@ -5,6 +5,10 @@
   </div>
 </template>
 
+<script setup lang="ts">
+defineOptions({ name: 'OcDropzone' })
```
kulmann, line 9: why do we need the name here?

JammingBen: Because the component doesn't have a `script` tag otherwise, meaning it's `__name` and `name` properties will both be empty.

kulmann: but can't you add an empty script tag? as in `<script setup lang="ts"></script>`

JammingBen: As I wrote you in chat, unfortunately it still results in `__name` being `undefined`.

kulmann: thanks for trying it!

PR #190, opened 2025-02-18
docs(design-system): add docs for components g-p

JammingBen:
As per title.

### DIFF
--- a/packages/design-system/docs/components/OcGrid/OcGrid.md
+++ b/packages/design-system/docs/components/OcGrid/OcGrid.md
@@ -18,7 +18,7 @@ The component provides a default slot for any content that should be displayed i
 	<div style="width: 80px; height: 80px; background-color: var(--oc-color-swatch-warning-default)" class="oc-height-small"></div>
 </oc-grid>
 
-<oc-grid :flex="true" class="oc-mt-m">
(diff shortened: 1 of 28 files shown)


packages/design-system/docs/components/OcGrid/OcGrid.md
```diff
@@ -0,0 +1,27 @@
...
+
+::: livecode
+```html
+<oc-grid>
+	<div style="width: 80px; height: 80px; background-color: var(--oc-color-swatch-passive-default)" class="oc-height-small"></div>
+	<div style="width: 80px; height: 80px; background-color: var(--oc-color-swatch-success-default)" class="oc-height-small"></div>
+	<div style="width: 80px; height: 80px; background-color: var(--oc-color-swatch-warning-default)" class="oc-height-small"></div>
+</oc-grid>
+
+<oc-grid :flex="true" class="oc-mt-m">
```
kulmann: I don't like the `flex` prop we're using here... would be a great chance to change that, since it's only used in one location. Can we replace the `flex` prop with a `direction` prop which is `row` by default and `column` as the second option? That's what flexbox is doing and what this component also does internally, just that `flex` is a super weird prop name for that, especially since it's a boolean.

kulmann (approved):
💪

PR #201, opened 2025-02-20
feat(design-system): implement parser for emits, props and slots

JammingBen:
Adds a doc parser for component emits, props and slots that can be used in the design-system docs. It automatically parses the data so there is no need to manually add this information to the docs.

### Usage

The documented component needs to have the interfaces `Emits`, `Props` and/or `Slots`, depending on what should be documented. Each property can have a doc block that currently supports the `@docs` annotation (= `description`). The `Props` interface additionally supports the `@default` annotation to specify the default value of a property, which will be appended at the end of the description.

To embed the documentation into any component `md` file, you simply need to add the following:

```md
::: component-api
```

This will e.g. result in the following:

image

### How it works

The data gets parsed during build time via a vite plugin. It searches all component `vue` files and extracts the information into separate `json` files for emits, props and slots. These land in the vitepress `public` folder to ensure they are accessible by the browser in form of static assets.

Docs then can use a dedicated component to load and display the generated json files.


kulmann (approved):
This is pure gold 💰 thank you! 😍  as discussed, I'd like to see `Props`, `Emits` and `Slots` on each and every component docs page - just with e.g. a `This component doesn't define emits` note if a component has none defined.

kulmann (approved):
🥳

PR #204, opened 2025-02-20
Sanitize for md editor 

AlexAndBear:
bugfix: sanitize content from md-editor

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

### DIFF
--- a/packages/web-pkg/src/components/TextEditor/TextEditor.vue
+++ b/packages/web-pkg/src/components/TextEditor/TextEditor.vue
@@ -50,7 +50,7 @@
 import { computed, defineComponent, unref, PropType } from 'vue'
 import { Resource } from '@opencloud-eu/web-client'
 
-import { config, MdEditor, MdPreview } from 'md-editor-v3'
+import { config, MdEditor, MdPreview, XSSPlugin } from 'md-editor-v3'
 import 'md-editor-v3/lib/style.css'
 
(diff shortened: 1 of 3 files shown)


packages/web-pkg/src/components/TextEditor/TextEditor.vue
```diff
@@ -110,11 +114,14 @@ export default defineComponent({
       }
     })
 
+    const sanitize = (html) => dompurify.sanitize(html)
```
kulmann: this doesn't seem to sanitize on initial rendering when I enter the malicious html... e.g `<a href="http://example.com/page.html?userInput=<script>alert('Injected!');</script>">test</a>`
only on subsequent renderings after page reload

AlexAndBear: Bad, then let's sanitize the currentContent(prop) as well.  


Or rather use

https://imzbf.github.io/md-editor-v3/en-US/demo#%F0%9F%94%92%20Add%20XSS%20extension

AlexAndBear: Can you pls check again ?

kulmann (approved):
thank you 🛡️

PR #205, opened 2025-02-20
docs(design-system): add docs for component api

JammingBen:
Adds docs for the API of design system components: Properties, Emits & Slots.


packages/design-system/src/components/OcAvatar/OcAvatar.vue
```diff
@@ -20,9 +20,22 @@ import OcImage from '../OcImage/OcImage.vue'
...
+  /**
+   * @docs The accessible label for the avatar. Only needed in case the avatar is used alone. If not specified, the avatar will get `aria-hidden="true"`.
+   */
   accessibleLabel?: string
+  /**
+   * @docs The source of the image to be displayed.
+   */
   src?: string
+  /**
+   * @docs The user name to extract the initials from.
```
kulmann: ... if the `src` prop is empty

packages/design-system/src/components/OcButton/OcButton.vue
```diff
@@ -18,21 +18,83 @@ import { RouteLocationRaw } from 'vue-router'
 import { getSizeClass } from '../../helpers'
 
 export interface Props {
+  /**
+   * @docs The appearance of the button.
+   * @default outline
+   */
   appearance?: 'filled' | 'outline' | 'raw' | 'raw-inverse'
+  /**
+   * @docs The aria label of the button. Needs to be present if the icon doesn't have a visible label.
```
kulmann:
```suggestion
   * @docs The aria label of the button. Needs to be present if the button doesn't have a visible label.
```

packages/design-system/src/components/OcButton/OcButton.vue
```diff
@@ -18,21 +18,83 @@ import { RouteLocationRaw } from 'vue-router'
...
+   */
   ariaLabel?: string
+  /**
+   * @docs Determines if the button is disabled.
+   * @default false
+   */
   disabled?: boolean
+  /**
+   * @docs The gap size between the button and its content elements.
```
kulmann:
```suggestion
   * @docs The gap size between content elements of the button.
```

packages/design-system/src/components/OcButton/OcButton.vue
```diff
@@ -18,21 +18,83 @@ import { RouteLocationRaw } from 'vue-router'
...
+   * @default medium
+   */
   size?: 'small' | 'medium' | 'large'
+  /**
+   * @docs The type of the button element. Only takes effect if the `type` is set to `button`.
+   * @default button
+   */
   submit?: 'null' | 'button' | 'submit' | 'reset'
+  /**
+   * @docs The target of the button if the `type` is set to `a'.
```
kulmann:
```suggestion
   * @docs The target of the button if the `type` is set to `a`.
```

packages/design-system/src/components/OcContextualHelper/OcContextualHelper.vue
```diff
@@ -16,10 +16,25 @@ import OcInfoDrop from '../OcInfoDrop/OcInfoDrop.vue'
...
+  /**
+   * @docs Text to be displayed at the end.
+   */
   endText?: string
+  /**
+   * @docs List of items to be displayed. Please refer to the component source for the `ContextualHelperDataListItem` type definition.
+   */
   list?: ContextualHelperDataListItem[]
+  /**
+   * @docs Link to be displayed at the end.
```
kulmann:
```suggestion
   * @docs Link for external references. Gets labeled with `Read more`.
```

packages/design-system/src/components/OcDrop/OcDrop.vue
```diff
@@ -44,10 +73,38 @@ export interface Props {
     | 'bottom-end'
     | 'left-end'
     | 'auto-end'
+  /**
+   * @docs Element selector that can used as a target of the drop. This takes precedence over the `toggle` property.
```
kulmann:
```suggestion
   * @docs Element selector that can be used as a target of the drop. This takes precedence over the `toggle` property.
```

kulmann: I'm confused about the second sentence. That is wrong and should be removed, right?

packages/design-system/src/components/OcIcon/OcIcon.vue
```diff
@@ -28,12 +28,37 @@ import { getSizeClass, uniqueId } from '../../helpers'
...
+   * @docs Name of the icon. Please refer to `Remixicon` for a list of available icons.
+   */
   name?: string
+  /**
+   * @docs Size of the icon.
+   * @default medium
+   */
   size?: 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge' | 'xxlarge' | 'xxxlarge'
+  /**
+   * @docs HTML type of the icon.
```
kulmann:
```suggestion
   * @docs HTML element to be used for the icon.
```

packages/design-system/src/components/OcInfoDrop/OcInfoDrop.vue
```diff
@@ -57,13 +57,38 @@ import { uniqueId } from '../../helpers'
...
+  /**
+   * @docs Text at the end of the info drop.
+   */
   endText?: string
+  /**
+   * @docs List of items to display in the info drop. Please refer to the component source for the `ContextualHelperDataListItem` type definition.
+   */
   list?: ContextualHelperDataListItem[]
+  /**
+   * @docs Asserts the event that triggers the info drop.
```
kulmann:
```suggestion
   * @docs Event that triggers the info drop.
```

packages/design-system/src/components/OcInfoDrop/OcInfoDrop.vue
```diff
@@ -57,13 +57,38 @@ import { uniqueId } from '../../helpers'
...
+   */
   dropId?: string
+  /**
+   * @docs Text at the end of the info drop.
+   */
   endText?: string
+  /**
+   * @docs List of items to display in the info drop. Please refer to the component source for the `ContextualHelperDataListItem` type definition.
+   */
   list?: ContextualHelperDataListItem[]
```
kulmann, line 75: A bit weird, that the type name includes the `ContextualHelper` prefix, although the contextual helper wraps the `OcInfoDrop` component. Wouldn't it make sense to rename the type to `InfoDropDataListItem`?

packages/design-system/src/components/OcLoader/OcLoader.vue
```diff
@@ -4,7 +4,14 @@
 
 <script setup lang="ts">
 export interface Props {
+  /**
+   * @docs Arial label to desribe the loader's purpose for screen readers.
```
kulmann:
```suggestion
   * @docs Aria label to describe the loader's purpose for screen readers.
```

packages/design-system/src/components/OcLogo/OcLogo.vue
kulmann, line 1: The component is unused, correct? Can we please delete the component and the docs page for it?

packages/design-system/src/components/OcModal/OcModal.vue
```diff
@@ -90,27 +90,102 @@ import { ContextualHelperData } from '../../helpers'
...
+  /**
+   * @docs Contextual helper data. Please refer to the component source for the `ContextualHelperData` type definition.
+   */
   contextualHelperData?: ContextualHelperData
+  /**
+   * @docs Label of the contextual helper.
+   */
   contextualHelperLabel?: string
+  /**
+   * @docs Additional class to be added to the modal.
```
kulmann:
```suggestion
   * @docs Additional class(es) to be added to the modal.
```

kulmann: Do we need to state that if you want to add multiple classes that has to happen in one string (separated by blanks)? Or is that self-explanatory (honestly I think so...)?

packages/design-system/src/components/OcModal/OcModal.vue
```diff
@@ -90,27 +90,102 @@ import { ContextualHelperData } from '../../helpers'
...
+  /**
+   * @docs Error message to be displayed below the input field.
+   */
   inputError?: string
+  /**
+   * @docs Label of the input field.
+   */
   inputLabel?: string
+  /**
+   * @docs Selection range of the input field.
```
kulmann:
```suggestion
   * @docs Selection range of the input field in case parts of the input content should be selected on first render.
```

packages/design-system/src/components/OcInfoDrop/OcInfoDrop.vue
JammingBen: Not sure If I understand 100%, but the type is used at various places across the code base, it's just being re-used here. So we would need to add a new type `InfoDropDataListItem`, which would be a 1:1 copy, just for this use case.

packages/design-system/src/components/OcModal/OcModal.vue
JammingBen: IMO that's self-explanatory because of the type and because it's "common Vue knowledge".

packages/design-system/src/components/OcInfoDrop/OcInfoDrop.vue
kulmann: meh, ok. then let's leave it like it is.

kulmann (approved):
Awesome ❤️

PR #238, opened 2025-02-25
fix: change docker image name

ScharfViktor:
- change docker image name to `${OC_IMAGE:-opencloudeu/opencloud-rolling:latest}`
- downgraded collabora version to `CODE 24.04.12.2`. the last 24.04.12.4 https://www.collaboraonline.com/code-24-04-release-notes/?utm_source=chatgpt.com doesn;t work (checked also with opencloud_full)


kulmann (approved):
thank you 🥇

PR #248, opened 2025-02-26
Adjust icons for disabled spaces

AlexAndBear:
## Description
Before:
image

image

After:
image

image

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

### DIFF
--- a/packages/web-app-files/src/views/spaces/Projects.vue
+++ b/packages/web-app-files/src/views/spaces/Projects.vue
@@ -542,6 +542,16 @@ export default defineComponent({
   .tile-default-image > svg {
     filter: grayscale(100%);
     opacity: 80%;
+    height: 100%;
+    max-height: 100%;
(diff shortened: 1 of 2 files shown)


AlexAndBear:
Not 100% aligned but still...

kulmann (commented):
Why is the icon size of a disabled and enabled space different at all? Should be the same size I think... also, the fact that the list view doesn't indicate disabled spaces with an icon or label is weird. Could be a followup though.

AlexAndBear:
It's not that the icon size change, rather the problem originates from having preset images on enabled spaces (the templating stuff) that do also resize with tile size. 

With disabled spaces or spaces without images we don't have an actual image and do render icons, that don't align so much. 

Without having preset images, we wouldn't have that problem.

Issue #212, opened 2025-02-21
Prevent Sidebar from Snapping


tbsbdr:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

> As a user with a screen at ca 900pw width I want the right sidebar not to snap to the left so that I don't loose orientation.

### User Value

- improve demo scenarios (side-by-side browsers)
- improve usage on small / windowed screens

### Acceptance Criteria

- right sidebar snaps on 1 or 2 smaller breakpoints
- collapse left sidebar when space is tight - define breakpoints eg. via hands-on with @tbsbdr 

## Visuals (problem description)

![Image](https://github.com/user-attachments/assets/281230cc-83cd-4918-90e9-f8a32d217e2a)

[Snap](https://www.youtube.com/watch?v=nm6DO_7px1I)!
expanded left sidebar
![Image](https://github.com/user-attachments/assets/d0308e31-6b12-47ee-a15a-37c11cf8535e)
collapsed left sidebar
![Image](https://github.com/user-attachments/assets/02192834-cf7c-4ed4-803c-54eb190a7127)

PR #251, opened 2025-02-28
feat: mobile sidebars

JammingBen:
Reworks how the sidebars on the left and right behave on smaller screens to improve the mobile experience. The left sidebar now toggles automatically with smaller screen sizes. The right sidebar on the other hand only snaps on screens <580px.

As discussed with @tbsbdr , we currently don't care about the user setting for the left sidebar. That means resizing the screen could potentially expand the sidebar despite the user having it toggled in general. We can tackle this as soon as we correctly persist this setting.

Also fixes the search bar appearance in the mobile view and refactors 2 components to use script setup. So this PR is best reviewed by commits.

closes https://github.com/opencloud-eu/web/issues/212

### DIFF
--- a/packages/web-app-admin-settings/src/components/AppTemplate.vue
+++ b/packages/web-app-admin-settings/src/components/AppTemplate.vue
@@ -1,67 +1,74 @@
 <template>
   <main class="oc-flex oc-height-1-1 app-content oc-width-1-1">
-    <app-loading-spinner v-if="loading" />
-    <template v-else>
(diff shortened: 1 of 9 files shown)


AlexAndBear (approved):
Damn, love it 😍 

There is still a little bug, that the main content disappears when the sidebar is opened in mobile view and the viewport will be extended. This also occurs on the main branch.

JammingBen:
> There is still a little bug, that the main content disappears when the sidebar is opened in mobile view and the viewport will be extended. This also occurs on the main branch.

Yes, will tackle that in a separate PR 👍

PR #254, opened 2025-02-28
fix: empty file list

JammingBen:
Fixes an issue where the file list would be blank if the sidebar is open and snapped (<580px screen width) and then increasing the window size.


AlexAndBear (approved):
🚀

PR #266, opened 2025-03-03
chore: add format command

JammingBen:
Adds a `format` command that runs prettier. It checks if all files are compliant with the prettier config rules. Run `pnpm format:check` check the formatting and `pnpm format:write` to fix all the issues.

We did this to keep both `eslint` and `prettier` separated from each other because they are supposed to tackle different things. Having `eslint` checking both code quality *and* formatting results in bad performance and conflicts between both tools. They can now run in CI to ensure that both formatting and code quality matches our standards.

I also ran `prettier` to fix alle the remaining issues in files that we rarely touch so we have a clean state. Also, `prettier` is now being hoisted to ensure it's globally available for all packages. That didn't work before, so each IDE was probably running different `prettier` instances instead of the one we have as dependency.


.github/ISSUE_TEMPLATE/1-bug-report.yml
```diff
@@ -7,12 +7,10 @@ body:
     attributes:
       label: Initial Checklist
       options:
-        - label:
-            I understand this is a bug report and questions should be posted in
+        - label: I understand this is a bug report and questions should be posted in
```
kulmann, line 10: I thought that starting text on the `- label:` line is only allowed if it fits in a single line... was that assumption wrong?

JammingBen: I think so, this seems to be valid yaml: https://yaml-multiline.info/

PR #281, opened 2025-03-05
fix test when admin creates spaces parallel

ScharfViktor:
when we run tests parallel `RETRY=1 HEADLESS=true PARALLEL=4 ./run-e2e.sh --suites admin-settings` test was flaky because admin user had 2 similar` team A` spaces. fixed it

### DIFF
--- a/tests/e2e/cucumber/features/admin-settings/spaces.feature
+++ b/tests/e2e/cucumber/features/admin-settings/spaces.feature
@@ -125,8 +125,8 @@ Feature: spaces management
       | Alice | Space Admin |
     And "Admin" creates the following project spaces using API
       | name   | id     |
-      | team A | team.a |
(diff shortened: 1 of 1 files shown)


JammingBen (approved):
LGTM 👍 

That generally means we need to be careful not to re-use the same space ids across multiple tests?

individual-it:
This might fix it for now, but is really hard to maintain as the test writer has to be careful what user/groups/spaces to use.
In theory every test should be independent of all the others, but if we run tests in parallel on the same system that is a big issue.

ScharfViktor:
> That generally means we need to be careful not to re-use the same space ids across multiple tests?

No, now we create for each test `Alice_uniqueName` `Brian_uniqueName`. when these unique users create a space with the same name - there are no conflicts. But only one non-unique user is the admin. There was one fake here, when admin created the same space in parallel.

PR #284, opened 2025-03-06
chore: remove hoisted dependencies that do not need to be hoisted

JammingBen:
Removes `sass` and `@opencloud-eu/design-system` from our hoisted dependencies because it's not necessary to hoist them. Seems to be a left over from the past.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,18 +1,13 @@
 auto-install-peers=true
 shared-workspace-lockfile=true
 
-# @vue/component-compiler-utils
-hoist-pattern[]=sass
+# for md-editor-v3
 hoist-pattern[]=*@codemirror*
 hoist-pattern[]=*@lezer*
-
-# eslint (public so they are picked up by IDEs)
-public-hoist-pattern[]=*eslint*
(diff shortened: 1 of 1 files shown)


.npmrc
```diff
@@ -1,18 +1,13 @@
...
-public-hoist-pattern[]=*eslint*
-public-hoist-pattern[]=*prettier*
-
-# Make ODS available to tests, even if the main package does not require it
-public-hoist-pattern[]=@opencloud-eu/design-system
-
 # c.f. https://github.com/vuejs/test-utils/pull/2197
 hoist-pattern[]=@vue/server-renderer
 
+# eslint and prettier (public so they are picked up by IDEs)
```
kulmann, line 11: Aaaah! Ooooh! Thank you ❤️

kulmann (approved):

.npmrc
JammingBen: I just moved the code block and clarified the comment though 😅 So you probably still need the manual config in your IDE for whatever reason.

Issue #2, opened 2025-01-14
Material Design Guideline


kulmann:
# OpenCloud Material Design Guideline

This guide aims at making the OpenCloud adaption of material design easy to 
understand and usable across **all client platforms**.

## Material Design Concepts

The base concept of material design is that components are being layered on
top of each other, where the highest layer is the one closest to the users eye,
while the furthest layer is the one in the background.
The background is called `Surface`.

The main design elements in material design are shapes, fonts and color roles.

## Color System

The material design color system consists of color roles. These roles are supposed
to express the different layers of components.

### Pairing and Layering Colors

Use matching `On` roles for text and icons on top of surfaces. This ensures visual accessibility (color contrast of 3:1 or better). 
Examples:
- Use `OnPrimary` for text and icons on `Primary` colored elements.
- Use `OnSurfaceVariant` for text and icons on `SurfaceVariant` colored elements.

### Primary

Use primary roles for most prominent components (important buttons, active states).
- `Primary` for important fills, text and icons against `Surface`
- `PrimaryContainer` for standout fills against `Surface` (rarely or never)

We only use `PrimaryContainer` for hover and active/focussed states.

### Secondary

Use secondary roles for less prominent components (anything that's not a CTA).
- `Secondary` for regular fills, text and icons against `Surface`
- `SecondaryContainer` for recessive components like tonal buttons against `Surface`

### Tertiary

Use tertiary roles for contrasting accents.
- `Tertiary` for complementary fills, text and icons against `Surface`
- `TertiaryContainer` for complementary container fills against `Surface`, like input fields (e.g. the active state of a toggle-switch).

### Error

Use error roles for error states (incorrect password, error message, etc).
- `Error` for attention-grabbing fills, text and icons against `Surface` (-> urgency!)
- `ErrorContainer` for attention-grabbing fills against `Surface` (-> less urgent)

### Surface

Use surface roles for backgrounds, container roles for components like cards, sheets, dialogs.
- `Surface` default color for backgrounds. Note: In OpenCloud Web the `background` layer is the app content area. 
- Five container roles expressing different level of emphasis:
  - `SurfaceContainerLowest` lowest-emphasis container
  - `SurfaceContainerLow` low-emphasis container
  - `SurfaceContainer` default container
  - `SurfaceContainerHigh` high-emphasis container
  - `SurfaceContainerHighest` highest-emphasis container
  - While `SurfaceContainer` is the default, others are helpful for creating hierarchy and nested containers.
[...]

PR #291, opened 2025-03-07
feat: material design

JammingBen:
Replaces our custom color framework with material design color roles.

#### To-Dos

- [x] Button variations
- [x] Add button hovers
- [x] Quota varation?
- [ ] Create color for our theme and update it in the server repo
- [x] Use theme as example in dev docs

refs https://github.com/opencloud-eu/web/issues/2


packages/design-system/src/components/OcDefinitionList/OcDefinitionList.vue
```diff
@@ -41,5 +41,10 @@ const { items } = defineProps<Props>()
   dd {
     margin-inline-start: var(--oc-space-medium);
   }
+
+  dd:last-of-type,
+  dt:last-of-type {
```
AlexAndBear, line 46: Doubled

packages/design-system/src/components/OcDefinitionList/OcDefinitionList.vue
```diff
@@ -41,5 +41,10 @@ const { items } = defineProps<Props>()
   dd {
     margin-inline-start: var(--oc-space-medium);
   }
+
+  dd:last-of-type,
```
AlexAndBear, line 45:
```suggestion
```

packages/design-system/src/components/OcProgressPie/OcProgressPie.vue
```diff
@@ -77,8 +77,8 @@ $default-size: 64px;
 
     &::before,
     &::after {
-      border: calc($default-size / 10) solid var(--oc-color-swatch-passive-default);
-      border-color: var(--oc-color-swatch-passive-default);
+      border: calc($default-size / 10) solid var(--oc-role-secondary);
+      border-color: var(--oc-role-secondary);
```
AlexAndBear, line 81: Maybe rather outline role?

AlexAndBear (changes requested):

packages/design-system/src/components/OcDefinitionList/OcDefinitionList.vue
kulmann: nope, the two lines are `dd` and `dt`, so we need both.

packages/design-system/src/components/OcProgressPie/OcProgressPie.vue
kulmann: Weird that we set the `border-color` after it was already set in the `border` attribute...

packages/design-system/src/components/OcDefinitionList/OcDefinitionList.vue
AlexAndBear: 🤦‍♀️

packages/design-system/src/components/OcProgressPie/OcProgressPie.vue
AlexAndBear: true

AlexAndBear (approved):

packages/design-system/src/components/OcProgressPie/OcProgressPie.vue
kulmann: Checked it, the border is the progress pie fill, so `secondary` is correct.

kulmann (approved):
Great result! ❤️

PR #319, opened 2025-03-13
allow space activities be checked by regex

individual-it:
## Description

since #32 the activity will contain a random suffix to the user-name. This fixes the issue by allowing to check the string by using regular expressions

## Related Issue

needed for https://github.com/opencloud-eu/qa/issues/3

## How Has This Been Tested?

ran test cases locally

## Types of changes

- [ ] Bugfix

### DIFF
--- a/tests/e2e/cucumber/features/smoke/activity.feature
+++ b/tests/e2e/cucumber/features/smoke/activity.feature
@@ -72,6 +72,9 @@ Feature: Users can see all activities of the resources and spaces
       |Alice Hansen shared team via link|
       |Alice Hansen added brian as member of team|
       | Alice Hansen added readme.md to .space |
+    And "Brian" should see activities of the space mathing the following regex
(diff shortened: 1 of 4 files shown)


individual-it:
In the activities the actor is shown with the full name and the user that was added with the user-name
image

ScharfViktor (approved):
👍

tests/e2e/cucumber/features/smoke/activity.feature
```diff
@@ -69,9 +69,11 @@ Feature: Users can see all activities of the resources and spaces
     And "Brian" navigates to the project space "team.1"
     Then "Brian" should see activity of the space
       | activity                               |
-      |Alice Hansen shared team via link|
-      |Alice Hansen added brian as member of team|
```
v-scharf, line 73: This step will be checked using regex

PR #347, opened 2025-03-19
chore(prettier-config): make prettier a peer dependency

JammingBen:
Declares `prettier` as a peer dependency because it's supposed to be provided by the user of this package. The package should only provide the config for it.


kulmann (approved):
good catch 👍

Issue #357, opened 2025-03-19
Deletion date is invalid


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

The deletion date in the right sidebar is invalid.

### Reproduction Steps

* Delete a file
* Go to the trash bin and open the right sidebar for that file

### Expected Outcome

The deletion date is being displayed.

### Actual Outcome

The deletion date is invalid.

PR #358, opened 2025-03-19
fix(files): deletion date in file details

JammingBen:
Fixes the invalid deletion date in the file details and refactors the component to script setup (separated into 2 commits to ease review).

fixes https://github.com/opencloud-eu/web/issues/357


kulmann (approved):
thank you! 💪

Issue #354, opened 2025-03-19
Unable to select all spaces in tiles view


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

It's not possible to select all spaces in tiles view when at least on space is disabled.

### Reproduction Steps

* Create 2 spaces
* Disable one
* Click on the checkbox on the top left to select all spaces

### Expected Outcome

All spaces are selected.

### Actual Outcome

The disabled space does not get selected.

PR #363, opened 2025-03-19
fix: select all checkbox in spaces tiles view

JammingBen:
Fixes the select all checkbox in the spaces tiles view so it also selects disabled spaces. We need to handle disabled spaces separately from disabled resources because they differ in behaviour.

Also refactors the `ResourceTiles` component to script setup.

fixes https://github.com/opencloud-eu/web/issues/354

### DIFF
--- a/packages/web-pkg/src/components/FilesList/ResourceTiles.vue
+++ b/packages/web-pkg/src/components/FilesList/ResourceTiles.vue
@@ -129,16 +129,14 @@
   </div>
 </template>
 
-<script lang="ts">
+<script setup lang="ts">
 import {
   computed,
   ComponentPublicInstance,
-  defineComponent,
   nextTick,
   onBeforeUnmount,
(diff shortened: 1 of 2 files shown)


packages/web-pkg/src/components/FilesList/ResourceTiles.vue
```diff
@@ -378,11 +378,17 @@ export default defineComponent({
...
       }
 
-      if (isSpaceResource(resource) && resource.disabled) {
+      return resource.processing === true
+    }
+
+    const isSpaceResourceDisabled = (resource: Resource) => {
+      // a disabled space behaves a bit different than a disabled resource because
+      // it still allows certain actions, hence we need to handle them separately.
+      if (unref(isResourceDisabled)(resource)) {
```
kulmann: Why the `unref`?

Issue #371, opened 2025-03-19
Resource can't be pasted into same folder


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

A resource can't be pasted into same folder via the UI. It only works via keyboard shortcuts.

### Reproduction Steps

* Create a file
* Copy it
* Try to paste it via the "Paste here" button

### Expected Outcome

The button is enabled and lets the user paste the file into the current folder, resulting in the conflict dialog being opened.

### Actual Outcome

The button is disabled.

PR #381, opened 2025-03-20
fix(files): copy pasting items into current folder

JammingBen:
Allows copy pasting items into the current folder and refactors the component to script setup on the way. Also fixes an issue where disabled secondary buttons would have a hover effect.

fixes https://github.com/opencloud-eu/web/issues/371

### DIFF
--- a/packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
+++ b/packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
@@ -180,6 +180,7 @@
 
 <script lang="ts">
 import {
+  ClipboardActions,
   FileAction,
   isLocationPublicActive,
   isLocationSpacesActive,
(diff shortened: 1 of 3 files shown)


packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
```diff
@@ -402,6 +403,10 @@ export default defineComponent({
     })
 
     const isPastingIntoSameFolder = computed(() => {
```
kulmann: The name of the `computed` is not good anymore with this change. Can you think of another one?

kulmann: Same for the user facing error message on line 424:
`return $gettext('You cannot paste resources into the same folder.')`

That's more like `You cannot cut&paste resource within the same folder.` or something like that.

kulmann (approved):
works like a charm 🥇

Issue #380, opened 2025-03-20
Delete queue is missing in tiles view


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

The delete queue is missing in the tiles view, it's only being used in the table.

### Reproduction Steps

* Create a bunch of files
* Switch to the tiles view
* Switch to a slow network mode via the browser console
* Select all of the files and delete them

### Expected Outcome

The delete queue should work and indicate that files are being deleted.

### Actual Outcome

The delete queue doesn't work. Files are just being de-selected and disappear eventually.

PR #389, opened 2025-03-20
fix(pkg): add missing delete queue to tiles view

JammingBen:
Adds the missing delete queue to the tiles view by also adding a loading state to the component. Refactors it to script setup on the way. This PR is best reviewed by its individual commits.

fixes https://github.com/opencloud-eu/web/issues/380

### DIFF
--- a/packages/web-pkg/src/components/FilesList/ResourceTile.vue
+++ b/packages/web-pkg/src/components/FilesList/ResourceTile.vue
@@ -86,8 +86,8 @@
   </div>
 </template>
 
-<script lang="ts">
-import { computed, defineComponent, PropType } from 'vue'
+<script setup lang="ts">
+import { computed, customRef, ref, unref } from 'vue'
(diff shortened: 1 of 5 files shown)


packages/web-pkg/src/components/FilesList/ResourceTile.vue
```diff
@@ -93,137 +96,111 @@ import { useGettext } from 'vue3-gettext'
...
+})
+const resourceDescription = computed(() => {
+  if (isSpaceResource(resource)) {
+    return resource.description
+  }
+  return ''
+})
+
+const shouldDisplayThumbnails = (resource: Resource) => {
+  return resource.thumbnail
```
AlexAndBear: Maybe remove this helper fnc since it doesn't add any value ?

AlexAndBear (changes requested):

AlexAndBear (approved):

Issue #356, opened 2025-03-19
Drag hover in file list is not visible


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

The drag hover in the file list is not visible.

### Reproduction Steps

* Create a folder and a file
* Drag the file into the folder

### Expected Outcome

The folders gets highlighted when the file is being dragged over it.

### Actual Outcome

The highlight is missing.

Issue #348, opened 2025-03-19
Wrong hovers


JammingBen:
Some hovers look out of place and should be removed.

* Name and icon of files that get downloaded on click

<img width="216" alt="Image" src="https://github.com/user-attachments/assets/20c9ba5d-ad31-49a9-bbfc-ef477a957640" />

* Names of spaces in the trash overview

* Parent folder indicators (e.g. in search results)

<img width="164" alt="Image" src="https://github.com/user-attachments/assets/33b8e020-ac0f-4140-975d-532333744dca" />

* Role select in the "Create Link" modal (should not have a hover since it can't make padding work here)

<img width="167" alt="Image" src="https://github.com/user-attachments/assets/327fe458-cdb8-4019-90e7-70bfafaff13d" />

* Table sort arrow (either no hover or the white surface background with a bit more padding)

<img width="129" alt="Image" src="https://github.com/user-attachments/assets/d6f933a1-1e57-41fe-b55f-876b97e6b7c9" />

* The "Mark as read" button on the top right of the notifications drop

<img width="398" alt="Image" src="https://github.com/user-attachments/assets/814e5fb4-77e4-40f7-8f76-ff364272b3cd" />

* The "hide" action should have a the surface background, not the grayish one.

<img width="116" alt="Image" src="https://github.com/user-attachments/assets/31dd8c31-ce03-4265-a03f-d1df71f6ee9d" />

Issue #345, opened 2025-03-18
Relogin Button looks broken on hover


kulmann:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

After the material design implementation the hover state of the re-login button on the `access denied` page looks broken.

### Reproduction Steps

1. Visit the `/access-denied` page
2. Hover over the `Log in again` button

### Expected Outcome

Button looks nice

### Actual Outcome

Button looks awful, see screenshot.

<img width="489" alt="Image" src="https://github.com/user-attachments/assets/42e1d99d-c1b8-4ca4-8c1d-a21713fcd1b3" />

PR #395, opened 2025-03-21
fix: various hovers and small visual glitches

JammingBen:
fixes https://github.com/opencloud-eu/web/issues/348
fixes https://github.com/opencloud-eu/web/issues/345
fixes https://github.com/opencloud-eu/web/issues/356

### DIFF
--- a/packages/web-pkg/src/components/ViewOptions.vue
+++ b/packages/web-pkg/src/components/ViewOptions.vue
@@ -298,7 +298,7 @@ export default defineComponent({
 .oc-range {
   -webkit-appearance: none;
   border-radius: 0.3rem;
-  background: var(--oc-role-secondary-container);
+  background: var(--oc-role-surface-container-high);
   height: 0.5rem;
   outline: none;
   width: 100%;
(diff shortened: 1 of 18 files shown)


packages/web-pkg/src/components/ViewOptions.vue
```diff
@@ -298,7 +298,7 @@ export default defineComponent({
 .oc-range {
   -webkit-appearance: none;
   border-radius: 0.3rem;
-  background: var(--oc-role-secondary-container);
+  background: var(--oc-role-surface-container-high);
```
kulmann, line 301: According to material design documentation we could (or should?) use `tertiary-container` for the slider and also for OcSwitch enabled backgrounds in general... Played around with it, looks like this with our current color tokens:

Screenshot 2025-03-21 at 08 24 06

With your state:

Screenshot 2025-03-21 at 08 25 46

What do you think about that?

kulmann: eh... kind of unrelated to this PR, since you made the slider more neutral. Sorry, didn't get that before... If you think the tertiary-container might be a thing for the switch -> ping me or open an issue please. :-)

kulmann (approved):
Checked all of them in the ui now 😎 thanks for the attention to detail, really good improvement! 🥳

packages/web-pkg/src/components/ViewOptions.vue
JammingBen: Hmm to me the tertiary color looks a bit off... maybe because it's one of the generated colors that doesn't really go well with our custom lilac.

PR #440, opened 2025-03-26
Fix: don't show version edition in left sidebar

AlexAndBear:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/web-runtime/src/container/versions.ts
+++ b/packages/web-runtime/src/container/versions.ts
@@ -15,6 +15,5 @@ export const getBackendVersion = ({
   }
   const product = backendStatus.product || 'OpenCloud'
   const version = backendStatus.productversion || backendStatus.versionstring
-  const edition = backendStatus.edition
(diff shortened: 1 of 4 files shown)


packages/web-runtime/src/container/versions.ts
```diff
@@ -15,6 +15,5 @@ export const getBackendVersion = ({
   }
   const product = backendStatus.product || 'OpenCloud'
   const version = backendStatus.productversion || backendStatus.versionstring
-  const edition = backendStatus.edition
```
kulmann, line 18: `const edition = backendStatus.edition || ''` should already do the trick, right? We want the edition to be shown if it's set. We just don't want a nullish string to be printed. 🙈 The rest of the PR can be reverted, we do want to keep the functionality.

kulmann (changes requested):

PR #444, opened 2025-03-27
refactor: simplify status indicators handling

JammingBen:
Status indicators were stored/cached on the resource objects, that required updating them in a lot of places spread over the whole codebase. This change moves computation into the view so status indicators for resources not visible aren't computed at all.


kulmann (approved):
Good catch 👍  Thank you!

Issue #449, opened 2025-03-28
favicon does not work with Firefox


individual-it:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

On the demo site and also on my private installation started from https://github.com/opencloud-eu/opencloud/tree/16071354882880c4ff22eddfdada25ee15aaf20f/deployments/examples/opencloud_full
the favicon is not shown in Firefox, but works fine in Chrome
trying to access https://demo.opencloud.eu/signin/v1/static/favicon.ico shows me a message that it contains errors.

Firefox 136.0.4 (64-Bit) on Ubuntu 24.04

### Reproduction Steps

access https://demo.opencloud.eu/signin/v1/static/favicon.ico  wit Firefox

### Expected Outcome

icon should be shown

### Actual Outcome

![Image](https://github.com/user-attachments/assets/8a49f804-4e7d-45bf-9e3a-c55daabf8900)

kulmann:
@individual-it do you have the same issue with the web ui itself? The error you're showing stems from the IDP, so I'd move this to the [opencloud repo](https://github.com/opencloud-eu/opencloud) if there is no issue in web itself.

individual-it:
Not sure what you mean, but if I just run the whole thing it happens:
FF: ![Image](https://github.com/user-attachments/assets/dfe3a240-dca3-4d43-99b3-eea4f33d41c4)
Chrome: ![Image](https://github.com/user-attachments/assets/e44f1b36-270f-4dd1-9209-3e271e184212)

kulmann:
Ok, thank you... then we need to fix it for both web itself and for the idp. But we can keep the issue here in the web repo.

PR #457, opened 2025-03-31
feat: support svg as icon

AlexAndBear:
## Related Issue

- Part of #449 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/web-runtime/src/composables/head/useHead.ts
+++ b/packages/web-runtime/src/composables/head/useHead.ts
@@ -22,7 +22,9 @@ export const useHead = () => {
               .join(', ')
           }
         ],
-        ...(unref(favicon) && { link: [{ rel: 'icon', href: unref(favicon) }] })
+        ...(unref(favicon) && {
+          link: [{ rel: 'icon', type: 'image/svg+xml', href: unref(favicon) }]
+        })
       }
     })
   )


packages/web-runtime/src/composables/head/useHead.ts
```diff
@@ -22,7 +22,9 @@ export const useHead = () => {
               .join(', ')
           }
         ],
-        ...(unref(favicon) && { link: [{ rel: 'icon', href: unref(favicon) }] })
+        ...(unref(favicon) && {
+          link: [{ rel: 'icon', type: 'image/svg+xml', href: unref(favicon) }]
```
kulmann, line 26: since `unref(favicon)` is the url of the favicon... could you please only use the `type: 'image/svg+xml'` if the url ends with `.svg`? 🤔

kulmann (changes requested):

packages/web-runtime/src/composables/head/useHead.ts
AlexAndBear: Good point 👍

AlexAndBear: quadro-checked, not really necessary with modern browsers, closing that pr

PR #471, opened 2025-04-01
chore: pin opencloud version to v2.1 in dev stack

kulmann:


JammingBen (commented):
Good catch 👍 It's missing for the `collaboration` container though.

kulmann:
> Good catch 👍 It's missing for the `collaboration` container though.

Thank you, changed that as well now. I already fell into that trap before... -.-

AlexAndBear (approved):

JammingBen (approved):

Issue #478, opened 2025-04-02
long names for public links break the UI


individual-it:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

A long name for a public link overflows the window and makes it impossible to click the copy button or the "more-functions" button

### Reproduction Steps

1. create a new public link
2. edit the name and give a really long one

### Expected Outcome

at least the buttons should be usable

### Actual Outcome

![Image](https://github.com/user-attachments/assets/1e4fea62-40f4-4420-b0f9-7ffaa665624c)

PR #481, opened 2025-04-02
fix(files): truncation on long link names

JammingBen:
fixes https://github.com/opencloud-eu/web/issues/478

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Shares/Links/ListItem.vue
+++ b/packages/web-app-files/src/components/SideBar/Shares/Links/ListItem.vue
@@ -3,11 +3,7 @@
     <div class="oc-flex oc-flex-middle">
       <oc-avatar-item :width="36" icon-size="medium" icon="link" name="link" />
       <div class="files-links-name-wrapper oc-pl-s">
-        <span
-          aria-hidden="true"
-          class="files-links-name oc-text-truncate"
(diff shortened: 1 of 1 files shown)


packages/web-app-files/src/components/SideBar/Shares/Links/ListItem.vue
```diff
@@ -1,13 +1,13 @@
...
-      <oc-avatar-item :width="36" icon-size="medium" icon="link" name="df" />
+      <oc-avatar-item :width="36" icon-size="medium" icon="link" name="link" />
       <div class="files-links-name-wrapper oc-pl-s">
-        <div class="oc-flex oc-flex-middle">
-          <div class="oc-text-truncate">
-            <span aria-hidden="true" class="files-links-name" v-text="linkShare.displayName" />
-          </div>
-        </div>
+        <span
+          aria-hidden="true"
```
kulmann: hm... why is the link displayname aria-hidden at all? 🤔

JammingBen: Well that's a good question 😅 I can't think of any good reason to be honest... should I remove it?

kulmann: Yes, please 👍

PR #486, opened 2025-04-02
[full-ci]use Playwright api instead of node fetch

ScharfViktor:
This PR replaces the usage of node-fetch with Playwright’s request API
We no longer need to save cookies and send them in another request because playwright saves cookies in each context and reuses them

### DIFF
--- a/tests/e2e/support/api/keycloak/openCloudUserToken.ts
+++ b/tests/e2e/support/api/keycloak/openCloudUserToken.ts
@@ -1,7 +1,7 @@
-import fetch from 'node-fetch'
 import { TokenEnvironmentFactory } from '../../environment'
 import { config } from '../../../config'
 import { User } from '../../types'
+import { request, APIRequestContext } from '@playwright/test'
 
 interface openCloudTokenForKeycloak {
   access_token: string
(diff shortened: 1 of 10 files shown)


tests/e2e/support/api/keycloak/openCloudUserToken.ts
```diff
@@ -147,3 +138,66 @@ export const refreshAccessTokenForKeycloakOpenCloudUser = async (user: User) =>
...
+  const refreshToken = tokenEnvironment.getToken({ user }).refreshToken
+  const context = await request.newContext()
+
+  const refreshResponse = await context.post(tokenMasterEndpoint, {
+    form: {
+      client_id: 'admin-cli',
+      grant_type: 'refresh_token',
+      refresh_token: refreshToken
+    }
+    // header: { 'Content-Type': 'application/x-www-form-urlencoded' },
```
JammingBen: Can this be removed?

JammingBen (approved):
Great change 👍

tests/e2e/support/api/keycloak/openCloudUserToken.ts
v-scharf: Yes, I forgot to do it

PR #487, opened 2025-04-02
chore: design system component types

JammingBen:
Adds proper types for the design-system components and fixes type issues afterwards.

Also removes the `vue` peer dependency from the design-system because it's redundant and causes issues with the vue language server in VSCode.


kulmann (approved):
Woohoo, thank you for looking into this! 🥇

Issue #488, opened 2025-04-02
Live code blocks in design-system docs broken


JammingBen:
The live code blocks in the design-system docs are broken. In dev mode everything is fine, with a production build however there is an issue with bundling the vue files.

PR #493, opened 2025-04-03
fix(design-system): design system live doc blocks

JammingBen:
There was in issue with dynamically importing vue files for the live code blocks. They need to be imported via a glob import for vite to bundle them correctly.

Also removes redundant and irrelevant stuff and moves the `LiveCodeBlock` component to the `components` folder.

fixes https://github.com/opencloud-eu/web/issues/488


kulmann (approved):
Thanks for the fix and cleanup! 🥳

Issue #45, opened 2025-02-06
Datepicker clear button is off


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

The clear button of the datepicker input seems to be off when a date has been selected.

### Reproduction Steps

1. Add a public link for a resource
2. Add an expiration date to the link 
3. In the modal, enter a valid expiration date

### Expected Outcome

The datepicker button and the clear button on the right sit side by side.

### Actual Outcome

The icons kinda overlap:

<img width="483" alt="Image" src="https://github.com/user-attachments/assets/11edb163-2d92-4f88-b727-4c166f46f320" />

PR #502, opened 2025-04-04
do not overwrite release when uploading artifacts

individual-it:
## Description

This pipeline runs after ready-release-go, which generates the title and changelogs in the notes.
We don't want to overwrite the existing notes

## Related Issue

- Fixes [#45](https://github.com/opencloud-eu/qa/issues/45)

## How Has This Been Tested?

@kulmann I think for testing you would have to create a new release with ready-release-go

## Types of changes

- [ ] Bugfix


kulmann (approved):
Thank you! Let's give it a try.

PR #520, opened 2025-04-08
fix: use native fetch for downloading archives

JammingBen:
Axios seems to have problems with requests that exceed a certain size which results in data streams that can't be read. Use fetch instead to ensure big archives can still be downloaded.

This is a follow-up of https://github.com/opencloud-eu/web/pull/465.

refs https://github.com/opencloud-eu/opencloud/issues/537


AlexAndBear (approved):

kulmann (approved):
works like a charm 👍

PR #532, opened 2025-04-09
download big archive

ScharfViktor:
test to download a large archive. By default we can download an archive with size not more than 1 GB
If the archive is bigger - I check that the button is disabled and has tooltip.

I was thinking how to upload big test datas so here are 2 identical tests but with different file uploads: 
- via api - (works for different storages)

Screenshot 2025-04-09 at 16 35 26

- via posix directly into the file systems. `STORAGE_USERS_POSIX_WATCH_FS=true`. (This way is really quick, but works only for posix) I suggest keeping it to test this new feature

Screenshot 2025-04-09 at 16 36 48


tests/e2e/cucumber/features/file-action/download.feature
```diff
@@ -62,3 +62,48 @@ Feature: Download
...
+      """
+    And "Alice" logs out
+
+  
+  @posix
+  Scenario: download large archive using upload with posix 
+    Given "Alice" creates the following files into personal space using API
+      | pathToFile | content     |
+      | new.txt | new content |
+    And "Alice" creates following filess using CLI:
```
individual-it:
```suggestion
    And "Alice" creates following files using CLI:
```

tests/e2e/cucumber/features/file-action/download.feature
```diff
@@ -62,3 +62,48 @@ Feature: Download
...
+      | resource         | size |
+      | testfile500Mb.bin| 500  |
+      | testfile400Mb.bin| 400  |
+      | testfile300Mb.bin| 300  |
+    When "Alice" logs in
+    And "Alice" downloads the following resources using the batch action
```
individual-it: There is no "Then" step for this action. How do we know the download was successful?

tests/e2e/cucumber/features/file-action/download.feature
```diff
@@ -62,3 +62,48 @@ Feature: Download
       | resource       | type |
       | testavatar.jpg | file |
     And "Brian" logs out
+
+  @posix
```
individual-it: this test should work on any storage

tests/e2e/cucumber/features/file-action/download.feature
```diff
@@ -62,3 +62,48 @@ Feature: Download
...
+    When "Alice" selects all files
+    Then the download button should be disabled for user "Alice" with the tooltip:
+      """
+      The selection exceeds the allowed archive size (max. 1.1 GB)
+      """
+    And "Alice" logs out
+
+  
+  @posix
+  Scenario: download large archive using upload with posix
```
individual-it:
```suggestion
  Scenario: download large archive using adding files with posix-watch-fs
```

tests/e2e/cucumber/steps/cli.ts
```diff
@@ -0,0 +1,25 @@
...
+import { World } from '../environment'
+import { execSync } from 'child_process'
+
+Given(
+  '{string} creates following files(s) using CLI:',
+  function (this: World, stepUser: string, stepTable: DataTable) {
+    const id = this.usersEnvironment.getCreatedUser({ key: stepUser }).uuid
+
+    const containerIdOrName = 'opencloud'
+    const storagePath = `/var/lib/opencloud/storage/users/users/${id}`
```
individual-it: can we make these values variable?

individual-it (dismissed):

tests/e2e/cucumber/features/file-action/download.feature
v-scharf: Yeah, I know. we cannot check that file was download on UI. we check that file was downloaded as archiver `expect(name).toBe('download')` inside the method

tests/e2e/cucumber/features/file-action/download.feature
v-scharf: are you sure?

tests/e2e/cucumber/steps/cli.ts
v-scharf: Probably you mean `const containerIdOrName = 'opencloud'` ?
if yes, I'll do it. first I want to adjust this test in the CI

tests/e2e/cucumber/features/file-action/download.feature
v-scharf: deleted this test

JammingBen (approved):
👍

Issue #540, opened 2025-04-11
Group member filter outline is cut off


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

The outline of the group member input field is cut off when focussed.

### Reproduction Steps

- go to the admin-settings via the app switcher
- go to "Groups"
- right-click a group and click "Show details"
- open the "Members" panel
- click in the input at the top

### Expected Outcome

The outline is not being cut off.

### Actual Outcome

The outline if cut off, probably because of some hidden overflow or `oc-text-truncate` class sitting on the input.

<img width="441" alt="Image" src="https://github.com/user-attachments/assets/fbab0288-a6ec-4da0-bcac-2c7059bc01f9" />

PR #551, opened 2025-04-14
fix: removed unnecessary oc-text-truncate to avoid a cut off

tammi-23:
Removed unnecessary oc-text-truncate to avoid the cut off inside the outline of the group member input field, when it has focus.

fixes: #540


AlexAndBear (approved):

JammingBen (changes requested):
It looks like the commit from https://github.com/opencloud-eu/web/pull/550 sneaked in here 🤔 

Also, 2 snapshot tests fail. Sometimes we have snapshots of the DOM in our units tests, and since you removed the  `oc-text-truncate` class, the snapshots need to be updated. You can do so by running `pnpm test:unit -u`.

JammingBen (approved):
Nice 👍

PR #553, opened 2025-04-15
fix: added oc-text-truncate to avoid line overflows in search preview

tammi-23:
Added oc-text-truncate to avoid line overflows into other columns in search preview

fixes: #546


JammingBen (approved):
Works nicely 👍

PR #611, opened 2025-04-29
chore(design-system): bundle css & externalize vue

JammingBen:
* Makes sure the css styles get bundled correctly so they can be imported via `@opencloud-eu/design-system/dist/design-system.css`.
* Externalize vue during build. This ensures that the components can be used properly when importing the design-system as a module in other projects. Vue is a peer dependency and needs to be provided by the host application.

### DIFF
--- a/packages/design-system/README.md
+++ b/packages/design-system/README.md
@@ -6,7 +6,7 @@ The **OpenCloud Design System** provides components and utilities for applicatio
 OpenCloud Web ecosystem. It can be developed standalone via the design system documentation. The documentation is
 built with [VitePress](https://vitepress.dev/).
 
-Head over to the [hosted docs](https://design.opencloud.eu/) for more information!
+Head over to the [hosted docs](https://docs.opencloud.eu/design-system/) for more information!
 
 ## Running the docs locally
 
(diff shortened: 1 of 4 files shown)


packages/design-system/README.md
```diff
@@ -24,14 +24,34 @@ pnpm docs:dev
...
-To use the components and utilities of the design-system in your application, you first need to install the package. Depending on your package manager, run one of the following commands:
+### Installation
+
+To use the design-system in your application, you first need to install the package. Depending on your package manager, run one of the following commands:
 
 ```
-$ npm install @opencloud-eu/design-system --save-dev
+$ npm install @opencloud-eu/design-system
 
 $ pnpm add -D @opencloud-eu/design-system
```
kulmann:
```suggestion
$ pnpm add @opencloud-eu/design-system
```

packages/design-system/README.md
```diff
@@ -24,14 +24,34 @@ pnpm docs:dev
...
 
-$ yarn add @opencloud-eu/design-system --dev
+$ yarn add @opencloud-eu/design-system
+```
```
kulmann, line 37: Thinking out loud, do you think it makes sense to add a note somewhere, installing the design system and importing the styles is only necessary for standalone apps, not for apps embedded into OpenCloud?

packages/design-system/README.md
```diff
@@ -24,14 +24,34 @@ pnpm docs:dev
...
+In order to use the provided CSS classes and to ensure the components are styled correctly, you need to import the styles like so:
+
+```
+import '@opencloud-eu/design-system/dist/design-system.css'
 ```
 
-It's recommended to install this package as a dev dependency because it's only really needed for providing autocompletion in your IDE and unit tests. In a runtime context, the OpenCloud Web runtime provides the actual implementation.
+### Components
+
+To use the components, you need to import the component you want to use. For example, to use the `OcButton` component:
```
kulmann: Is it possible to register all components globally?

packages/design-system/README.md
JammingBen: Yes good point, I added the info here and in the styling section.

packages/design-system/README.md
```diff
@@ -24,14 +24,62 @@ pnpm docs:dev
...
-$ npm install @opencloud-eu/design-system --save-dev
+import '@opencloud-eu/design-system/dist/design-system.css'
+```
+
+Again, this is not needed if you're using the design-system in an OpenCloud Web app because the styles are already available via the Web runtime.
+
+### Components
 
-$ pnpm add -D @opencloud-eu/design-system
+To use the components, you need to import the component you want to use. For example, to use the `OcButton` component:
```
kulmann:
```suggestion
To use a component, you need to import it. For example, to use the `OcButton` component:
```

kulmann (approved):
Thank you! Especially for adding so much docs content 🤓

PR #649, opened 2025-05-07
upload folder via dragAndDrop

ScharfViktor:
- upload folder via drag and drop
- create separate scenario for insufficient quota


AlexAndBear (approved):
🥳

Issue #674, opened 2025-05-12
[design-system] add missing features to text input


JammingBen:
### Initial Checklist

- [x] I understand this is a feature request and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Idea

The text input field is missing some feature we'd like to have:

* add option for a required mark (*)
* add support for a file upload field that returns a `File` object
* add support for a color picker

It might make sense to have the latter two as separate components. They probably require quite a bit of custom styling.

PR #675, opened 2025-05-12
feat(design-system): add required mark to text input component

JammingBen:
refs https://github.com/opencloud-eu/web/issues/674


kulmann (approved):
💯

PR #678, opened 2025-05-12
feat(design-system): add file input component

JammingBen:
refs https://github.com/opencloud-eu/web/issues/674

image


packages/design-system/src/components/OcFileInput/OcFileInput.vue
```diff
@@ -0,0 +1,262 @@
...
+const inputRef = useTemplateRef<HTMLInputElement>('inputRef')
+const inputBtnRef = useTemplateRef<HTMLElement>('inputBtnRef')
+
+const focus = () => {
+  unref(inputRef).focus()
+}
+defineExpose({ focus })
+
+const fileNames = computed(() => {
+  if (unref(inputRef)?.files) {
```
kulmann, line 216: This computed doesn't get re-evaluated:
- when I clear the selected file and add another selected file
- when I don't clear the selected file but just select another file when a file was already selected before

kulmann (changes requested):

PR #686, opened 2025-05-13
fix: minor design system issues and styling

JammingBen:
* fix attr inheritance on file input
* remove focus expose on file input component
* fix input description and error message styling
* add missing type definitions for color and file inputs
* fix file input value binding

### DIFF
--- a/packages/design-system/src/components/OcColorInput/OcColorInput.vue
+++ b/packages/design-system/src/components/OcColorInput/OcColorInput.vue
@@ -42,10 +42,9 @@
     >
       <oc-icon
         v-if="!!errorMessage"
-        name="information"
+        name="error-warning"
         size="small"
         fill-type="line"
-        accessible-label="info"
         aria-hidden="true"
         class="oc-mr-xs"
       />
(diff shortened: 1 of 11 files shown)


packages/design-system/src/styles/theme/oc-text.scss
```diff
@@ -13,7 +13,7 @@ html * {
   font-family: var(--oc-font-family);
 }
 
-html, button {
+html {
```
JammingBen, line 16: Button size is always being controlled inside the button component, no need to handle it here. It seems to interfere when using the design-system as library.

packages/design-system/src/components/OcColorInput/OcColorInput.vue
```diff
@@ -34,19 +34,20 @@
...
+      class="oc-color-input-message oc-text-small"
       :class="{
         'oc-color-input-description': !!descriptionMessage,
         'oc-color-input-danger': !!errorMessage
       }"
     >
       <oc-icon
-        v-if="messageText !== null && !!descriptionMessage"
+        v-if="!!errorMessage"
         name="information"
```
kulmann: What do you think about using `error-warning` (line?) as icon?

kulmann (approved):
Nice! Thank you for the attention to detail 😍

PR #691, opened 2025-05-14
fix(design-system): text input error message icon position

JammingBen:
Follow-up of https://github.com/opencloud-eu/web/pull/686.


kulmann (approved):
👍

Issue #600, opened 2025-04-28
Show Calendar URL in Web UI


tbsbdr:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

> As a user, I want to look up my personal calendar URL so that I can connect a third party application to it (e.g. Thunderbird, Apple Calendar).

### User Value

Promote Calendar availability

### Acceptance Criteria

- account page has a new panel (same architecture as app token panel)
- panel is only visible if a OPTIONS to the server URL (from config) urlJoined with path `/.well-known/caldav` responds with 301 and a location header set to `/caldav/`
- panel title is `Calendar` and shows a small badge with text `NEW` (technical note: `oc-tag` with attribute `:rounded="true"`)
- panel content consists of:
  - a short message which explains that the user has their own calendar available for integration into third party apps (thunderbird, apple calendar, etc)
  - configuration:
    - url: render the server url, can be copied to clipboard via button
    - username = render the opencloud username, can be copied to clipboard via button
    - password = show a hint that an app token needs to be generated and then can be used.

rhafer:
In theory with https://github.com/opencloud-eu/opencloud/pull/773 it should be enough if the carddav/caldav clients would configure `http://<my.cloud.domain>/caldav` or  `http://<my.cloud.domain>/carddav` as their base url and discover calenders and address books from there.

However, at least in the current Thunderbird this does not work for address books (https://bugzilla.mozilla.org/show_bug.cgi?id=1925006)

As a workaround we could display the full URL for the defaults:

Addressbook: `https://<my.cloud.domain>/carddav/<user-id>/def-addressbook/`
Calendar: `https://<my.cloud.domain>/caldav/<user-id>/def-calendar/`

tammi-23:
>panel is only visible if a OPTIONS to the server URL (from config) urlJoined with path /.well-known/caldav responds with 301 and a location header set to /caldav/

After thorough evaluation, this requirement cannot be reliably fulfilled using browser-based JavaScript due to platform limitations.
- The browser automatically follows 301 redirects.
- Even with fetch or using GET directly the browser does not expose the Location header of a redirect due to security and CORS restrictions. 
- As a result, client-side code cannot detect whether a redirect happened or where it pointed to.

[...]

PR #693, opened 2025-05-14
feat: Added CalDAV URL to the Accountpage

tammi-23:
Added a Calendar Panel to the User Account Page.
This includes: CalDAV URL, username and an information about the Calendar-Implementaion.

This implements: https://github.com/opencloud-eu/web/issues/600

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -107,7 +107,6 @@ services:
       #traefik.http.middlewares.redirect-caldav.redirectregex.replacement: '/caldav/'
       #traefik.http.middlewares.redirect-caldav.redirectregex.permanent: true
 
-
     volumes:
       - ./dev/docker/opencloud/csp.yaml:/etc/opencloud/csp.yaml
       - ./dev/docker/opencloud/proxy.yaml:/etc/opencloud/proxy.yaml
@@ -301,11 +300,11 @@ services:
     volumes:
       - ./dev/docker/radicale/data:/data
     #labels:
-      #traefik.enable: true
-      #traefik.http.routers.radicale.tls: true
-      #traefik.http.routers.radicale.rule: PathPrefix(`/`)
-      #traefik.http.routers.radicale.entrypoints: websecure
-      #traefik.http.services.radicale.loadbalancer.server.port: 5232
+    #traefik.enable: true
+    #traefik.http.routers.radicale.tls: true
+    #traefik.http.routers.radicale.rule: PathPrefix(`/`)
+    #traefik.http.routers.radicale.entrypoints: websecure
+    #traefik.http.services.radicale.loadbalancer.server.port: 5232
     restart: unless-stopped
 
 volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -101,9 +101,15 @@ services:
       traefik.http.routers.opencloud.rule: Host(`host.docker.internal`) && PathPrefix(`/`)
       traefik.http.routers.opencloud.entrypoints: opencloud
       traefik.http.services.opencloud.loadbalancer.server.port: 9200
-      traefik.http.routers.opencloud.middlewares: cors
+      traefik.http.routers.opencloud.middlewares: cors,redirect-caldav
+
+      traefik.http.middlewares.redirect-caldav.redirectregex.regex: "^/.well-known/caldav$"
+      traefik.http.middlewares.redirect-caldav.redirectregex.replacement: "/caldav/"
+      traefik.http.middlewares.redirect-caldav.redirectregex.permanent: true
+
     volumes:
       - ./dev/docker/opencloud/csp.yaml:/etc/opencloud/csp.yaml
+      - ./dev/docker/opencloud/proxy.yaml:/etc/opencloud/proxy.yaml
       - ./dev/docker/opencloud/password-policy-banned-passwords.txt:/etc/opencloud/password-policy-banned-passwords.txt
       - ./dev/docker/opencloud.idp.config.yaml:/etc/opencloud/idp.yaml
       - ./dev/docker/opencloud-ca:/var/lib/opencloud/proxy
@@ -249,6 +255,9 @@ services:
       traefik.http.middlewares.cors.headers.accesscontrolexposeheaders: '*'
       traefik.http.middlewares.cors.headers.accesscontrolmaxage: 100
       traefik.http.middlewares.cors.headers.addvaryheader: true
+      traefik.http.middlewares.radicale-redirect.redirectregex.regex: "^http://(.*):5232(.*)"
+      traefik.http.middlewares.radicale-redirect.redirectregex.replacement: "http://$1:9201$2"
+      traefik.http.middlewares.radicale-redirect.redirectregex.permanent: true
     ports:
       - '80:80'
       - '8090:8080'
@@ -277,6 +286,25 @@ services:
     container_name: web_tika
     ports:
       - 9998:9998
+    restart: unless-stopped
+
+  radicale:
+    image: ${RADICALE_DOCKER_IMAGE:-opencloudeu/radicale}:${RADICALE_DOCKER_TAG:-latest}
+    container_name: web_radicale
+    environment:
+      RADICALE_LOG_LEVEL: ${RADICALE_LOG_LEVEL:-info}
+      RADICALE_STORAGE_BACKEND: ${RADICALE_STORAGE_BACKEND:-filesystem} # Map  9201 to 5232
+      RADICALE_STORAGE_FILESYSTEM_FOLDER: /data
+    ports:
+      - '5232:5232'
+    volumes:
+      - ./dev/docker/radicale/data:/data
+    labels:
+      traefik.enable: true
+      traefik.http.routers.radicale.tls: true
+      traefik.http.routers.radicale.rule: PathPrefix(`/`)
+      traefik.http.routers.radicale.entrypoints: websecure
+      traefik.http.services.radicale.loadbalancer.server.port: 5232
     restart: unless-stopped
 
 volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -102,11 +102,6 @@ services:
       traefik.http.routers.opencloud.entrypoints: opencloud
       traefik.http.services.opencloud.loadbalancer.server.port: 9200
       traefik.http.routers.opencloud.middlewares: cors
-      #traefik.http.routers.opencloud.middlewares: cors,redirect-caldav
-      #traefik.http.middlewares.redirect-caldav.redirectregex.regex: '^/.well-known/caldav$'
-      #traefik.http.middlewares.redirect-caldav.redirectregex.replacement: '/caldav/'
-      #traefik.http.middlewares.redirect-caldav.redirectregex.permanent: true
-
     volumes:
       - ./dev/docker/opencloud/csp.yaml:/etc/opencloud/csp.yaml
       - ./dev/docker/opencloud/proxy.yaml:/etc/opencloud/proxy.yaml
@@ -255,9 +250,7 @@ services:
       traefik.http.middlewares.cors.headers.accesscontrolexposeheaders: '*'
       traefik.http.middlewares.cors.headers.accesscontrolmaxage: 100
       traefik.http.middlewares.cors.headers.addvaryheader: true
-      traefik.http.middlewares.radicale-redirect.redirectregex.regex: '^http://(.*):5232(.*)'
-      traefik.http.middlewares.radicale-redirect.redirectregex.replacement: 'http://$1:9201$2'
-      traefik.http.middlewares.radicale-redirect.redirectregex.permanent: true
+
     ports:
       - '80:80'
       - '8090:8080'
@@ -299,12 +292,6 @@ services:
       - '5232:5232'
     volumes:
       - ./dev/docker/radicale/data:/data
-    #labels:
-    #traefik.enable: true
-    #traefik.http.routers.radicale.tls: true
-    #traefik.http.routers.radicale.rule: PathPrefix(`/`)
-    #traefik.http.routers.radicale.entrypoints: websecure
-    #traefik.http.services.radicale.loadbalancer.server.port: 5232
     restart: unless-stopped
 
 volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -101,11 +101,12 @@ services:
       traefik.http.routers.opencloud.rule: Host(`host.docker.internal`) && PathPrefix(`/`)
       traefik.http.routers.opencloud.entrypoints: opencloud
       traefik.http.services.opencloud.loadbalancer.server.port: 9200
-      traefik.http.routers.opencloud.middlewares: cors,redirect-caldav
+      traefik.http.routers.opencloud.middlewares: cors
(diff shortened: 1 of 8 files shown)


packages/web-runtime/src/components/Account/CalDavUrl.vue
```diff
@@ -0,0 +1,121 @@
...
+const { user } = storeToRefs(userStore)
+const configStore = useConfigStore()
+const clientService = useClientService()
+
+const calDavAvailable = ref(false)
+const copiedIcon = 'check'
+const copyIcon = 'file-copy'
+
+const calDavUrl = computed(() => {
+  return `${configStore.serverUrl}/caldav/${user.value.id}/def-calendar/`
```
AlexAndBear:
```suggestion
  return `${configStore.serverUrl}/caldav/${unref(user).id}/def-calendar/`
```

packages/web-runtime/src/components/Account/CalDavUrl.vue
```diff
@@ -0,0 +1,121 @@
...
+})
+
+const copyCalDavUrlIcon = ref(copyIcon)
+const copyCalDavUsernameIcon = ref(copyIcon)
+
+const copyCalDavUrlToClipboard = () => {
+  navigator.clipboard.writeText(calDavUrl.value)
```
AlexAndBear:
```suggestion
  navigator.clipboard.writeText(unref(calDavUrl))
```

packages/web-runtime/src/components/Account/CalDavUrl.vue
```diff
@@ -0,0 +1,121 @@
...
+    const response = await clientService.httpUnAuthenticated.get(wellKnownUrl, {
+      method: 'OPTIONS',
+      maxRedirects: 0
+    })
+
+    if (response.status === 301 && response.headers['Location'] === '/caldav/') {
+      calDavAvailable.value = true
+    }
+  } catch (error) {
+    console.error('CalDAV check failed:', error)
```
AlexAndBear:
```suggestion
    console.info('CalDAV check failed:', error)
```

Rather an info, it's valid that the caldav server is not existing and therefore should not announce errors to the console

packages/web-runtime/src/components/Account/CalDavUrl.vue
```diff
@@ -0,0 +1,121 @@
+<template>
+  <div v-if="calDavAvailable">
+    <account-table
+      :title="$gettext('Calendar')"
+      :fields="[$gettext('CalDAV information name'), $gettext('CalCAD information value')]"
```
AlexAndBear:
```suggestion
      :fields="[$gettext('CalDAV information name'), $gettext('CalDAV information value')]"
```

tbsbdr:
forwarding screenies from @tammi-23 
image

image

packages/web-runtime/src/components/Account/CalDavUrl.vue
```diff
@@ -0,0 +1,131 @@
...
+onMounted(async () => {
+  try {
+    const wellKnownUrl = `${configStore.serverUrl}/.well-known/caldav`
+
+    try {
+      const response = await clientService.httpAuthenticated.get(wellKnownUrl, {
+        method: 'OPTIONS'
+      })
+
+      if (response.request.responseURL.includes('/caldav/')) {
```
AlexAndBear: Why not severurl/caldav/ seems more like a safe condition to me.

packages/web-runtime/src/extensions.ts
```diff
@@ -2,6 +2,7 @@ import { computed, markRaw } from 'vue'
 import { preferencesPanelExtensionPoint, progressBarExtensionPoint } from './extensionPoints'
 import AppTokens from './components/Account/AppTokens.vue'
 import { CustomComponentExtension, LoadingIndicator } from '@opencloud-eu/web-pkg'
+import CalDavUrl from './components/Account/CalDavUrl.vue'
```
AlexAndBear, line 5: Not 100 percent sure about importing this, I’d rather have an extension point not be dependent on a component, maybe you find a better place for the constant.

packages/web-runtime/src/components/Account/CalDavUrl.vue
```diff
@@ -0,0 +1,131 @@
...
+        calDavAvailable.value = true
+      }
+    } catch (error) {
+      console.error('CalDAV check failed:', error)
```
AlexAndBear: As noted before, I think a console.info is the better option.

AlexAndBear (changes requested):

tammi-23:
With the latest push, I updated the Radicale/CalDAV check to verify whether the redirect URL includes /caldav/.
→ https://github.com/opencloud-eu/web/issues/600

I also extended OcTag.vue by adding a color prop with a default value and implemented our website's highlight color (#ca8df5) as a custom option.

Additionally, I added a new configuration to our Docker development stack to enable Radicale.

Still missing: Unit Tests

docker-compose.yml
```diff
@@ -101,9 +101,15 @@ services:
       traefik.http.routers.opencloud.rule: Host(`host.docker.internal`) && PathPrefix(`/`)
       traefik.http.routers.opencloud.entrypoints: opencloud
       traefik.http.services.opencloud.loadbalancer.server.port: 9200
-      traefik.http.routers.opencloud.middlewares: cors
+      traefik.http.routers.opencloud.middlewares: cors,redirect-caldav
+
+      traefik.http.middlewares.redirect-caldav.redirectregex.regex: '^/.well-known/caldav$'
+      traefik.http.middlewares.redirect-caldav.redirectregex.replacement: '/caldav/'
+      traefik.http.middlewares.redirect-caldav.redirectregex.permanent: true
```
JammingBen: Since we're not accessing radicale via traefik, is this stuff really needed? Same with all the other traefik related lines below.

packages/design-system/src/components/OcTag/OcTag.vue
```diff
@@ -30,6 +30,10 @@ export interface Props {
    * @default false
    */
   rounded?: boolean
+  /**
+   * @docs The color of the tag.
+   */
+  color?: 'default' | 'warning' | 'custom-purple'
```
JammingBen: I'd stick with the color role schema as we know it from material: `primary`, `secondary` (default) and `tertiary`. Or even make this property take a css instruction as string and then you'd have an excuse to actually use `style` on the component 😄

packages/design-system/src/components/OcTag/OcTag.vue
```diff
@@ -113,4 +126,9 @@ function $_ocTag_click(event: MouseEvent) {
     }
   }
 }
+
+.oc-tag-color-custom-purple {
+  background-color: #ca8df5;
+  color: white;
```
JammingBen: `background-color` would then e.g. be `var(--oc-role-primary)` and `color` would be `var(--oc-role-on-primary)`.

packages/web-runtime/src/components/Account/CalDavUrl.vue
```diff
@@ -0,0 +1,132 @@
...
+const clientService = useClientService()
+
+const calDavAvailable = ref(false)
+const copiedIcon = 'check'
+const copyIcon = 'file-copy'
+const checked = ref(false)
+
+const calDavUrl = computed(() => {
+  return `${configStore.serverUrl}`
+})
```
JammingBen: This doesn't need to be a `computed`. Computed properties are only necessary when values as expected to change, which is not the case here. You could just use `configStore.serverUrl` directly where needed.

packages/web-runtime/src/components/Account/CalDavUrl.vue
```diff
@@ -0,0 +1,132 @@
...
+
+const copyCalDavUsernameToClipboard = () => {
+  navigator.clipboard.writeText(user.value.onPremisesSamAccountName)
+  copyCalDavUsernameIcon.value = copiedIcon
+  setTimeout(() => (copyCalDavUsernameIcon.value = copyIcon), 1500)
+}
+
+onMounted(async () => {
+  try {
+    const wellKnownUrl = `${configStore.serverUrl}.well-known/caldav`
```
JammingBen: No need to have the server url here, the client service already does this for you.

```suggestion
    const wellKnownUrl = '.well-known/caldav'
```

packages/web-runtime/src/components/Account/CalDavUrl.vue
```diff
@@ -0,0 +1,132 @@
...
+
+    try {
+      const response = await clientService.httpAuthenticated.get(wellKnownUrl, {
+        method: 'OPTIONS'
+      })
+      console.info('CalDAV check response:', response.request.responseURL)
+      if (response.request.responseURL.includes(`${configStore.serverUrl}caldav/`)) {
```
JammingBen: The `web-client` package provides a helper for building urls like this. It's a bit less error prone since you don't need to manually care about the slashes.

```suggestion
      if (response.request.responseURL.includes(urlJoin(configStore.serverUrl, 'caldav')) {
```

packages/web-runtime/tests/unit/components/Account/CalDavUrl.spec.ts
```diff
@@ -0,0 +1,118 @@
+import { defaultComponentMocks, defaultPlugins, mount } from '@opencloud-eu/web-test-helpers'
+import { flushPromises } from '@vue/test-utils'
+import { nextTick } from 'vue'
+import CalDavUrl from '../../../../src/components/Account/CalDavUrl.vue'
+import { vi } from 'vitest'
```
JammingBen: This is available globally, no need to import it.

```suggestion
```

packages/web-runtime/tests/unit/components/Account/CalDavUrl.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+          authStore: {
+            userContextReady: true
+          }
+        }
+      },
+      props: {
+        checked: true,
+        calDavAvailable: true
+      }
+    })
```
JammingBen: I'd suggest having one method per test file that returns the wrapper. See e.g. how it's done the unit tests for the app tokens: https://github.com/opencloud-eu/web/blob/main/packages/web-runtime/tests/unit/components/Account/AppTokens.spec.ts

packages/web-runtime/tests/unit/components/Account/CalDavUrl.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+      }
+    }),
+    useUserStore: () => ({
+      user: ref({
+        onPremisesSamAccountName: 'test-user'
+      })
+    }),
+    useConfigStore: () => ({
+      serverUrl: 'https://example.com/'
+    })
```
JammingBen: Are these mocks really needed? If so, please mock them via the plugins like e.g. here: https://github.com/opencloud-eu/web/blob/main/packages/web-runtime/tests/unit/components/Account/AppTokens.spec.ts#L96

JammingBen (commented):
A few annotations from my side, but overall it looks quite good already!

docker-compose.yml
```diff
@@ -102,8 +102,14 @@ services:
       traefik.http.routers.opencloud.entrypoints: opencloud
       traefik.http.services.opencloud.loadbalancer.server.port: 9200
       traefik.http.routers.opencloud.middlewares: cors
+      #traefik.http.routers.opencloud.middlewares: cors,redirect-caldav
+      #traefik.http.middlewares.redirect-caldav.redirectregex.regex: '^/.well-known/caldav$'
+      #traefik.http.middlewares.redirect-caldav.redirectregex.replacement: '/caldav/'
+      #traefik.http.middlewares.redirect-caldav.redirectregex.permanent: true
+
```
JammingBen:
```suggestion
```

docker-compose.yml
```diff
@@ -243,6 +249,9 @@ services:
       traefik.http.middlewares.cors.headers.accesscontrolexposeheaders: '*'
       traefik.http.middlewares.cors.headers.accesscontrolmaxage: 100
       traefik.http.middlewares.cors.headers.addvaryheader: true
+      traefik.http.middlewares.radicale-redirect.redirectregex.regex: '^http://(.*):5232(.*)'
+      traefik.http.middlewares.radicale-redirect.redirectregex.replacement: 'http://$1:9201$2'
+      traefik.http.middlewares.radicale-redirect.redirectregex.permanent: true
```
JammingBen: I think these can be removed as well?

```suggestion
```

docker-compose.yml
```diff
@@ -273,6 +282,25 @@ services:
...
+    ports:
+      - '5232:5232'
+    volumes:
+      - ./dev/docker/radicale/data:/data
+    #labels:
+    #traefik.enable: true
+    #traefik.http.routers.radicale.tls: true
+    #traefik.http.routers.radicale.rule: PathPrefix(`/`)
+    #traefik.http.routers.radicale.entrypoints: websecure
+    #traefik.http.services.radicale.loadbalancer.server.port: 5232
```
JammingBen:
```suggestion
```

packages/web-runtime/src/components/Account/CalDavUrl.vue
```diff
@@ -0,0 +1,128 @@
...
+const { $gettext } = useGettext()
+const userStore = useUserStore()
+const { user } = storeToRefs(userStore)
+const configStore = useConfigStore()
+const clientService = useClientService()
+
+const calDavAvailable = ref(false)
+const copiedIcon = 'check'
+const copyIcon = 'file-copy'
+const checked = ref(false)
```
AlexAndBear: Maybe we can give the variable a more descriptive naming

packages/web-runtime/src/components/Account/CalDavUrl.vue
```diff
@@ -0,0 +1,128 @@
...
+    try {
+      const response = await clientService.httpAuthenticated.get(wellKnownUrl, {
+        method: 'OPTIONS'
+      })
+
+      if (response.request.responseURL.includes(urlJoin(configStore.serverUrl, 'caldav'))) {
+        calDavAvailable.value = true
+      }
+    } catch (error) {
+      console.error('CalDAV check failed:', error)
```
AlexAndBear:
```suggestion
      console.info('CalDAV check failed:', error)
```

Since CalDav is not required for a working OpenCloud Instance, I wouldn't announce an error

AlexAndBear (changes requested):

AlexAndBear (approved):
🎉

packages/web-runtime/src/components/Account/CalDavUrl.vue
```diff
@@ -0,0 +1,128 @@
...
+            )
+          "
+        />
+      </template>
+
+      <oc-table-tr class="account-page-info-caldav-integration"></oc-table-tr>
+      <oc-table-tr class="account-page-info-caldav-url">
+        <oc-table-td>{{ $gettext('CalDAV URL') }}</oc-table-td>
+        <oc-table-td>
+          <span class="oc-text-truncate">{{ configStore.serverUrl }}</span>
```
AlexAndBear, line 31: Did this change ? I thought we need a full url with /caldav slug, now its only the server url ?

kulmann: Only the server URL is correct, because calendar apps will use the .well-known endpoints automagically.

PR #724, opened 2025-05-20
chore(design-system): make translations usable in library mode

JammingBen:
See additions in README.md.


kulmann (approved):
thank you for thinking this through 👍

PR #774, opened 2025-05-30
fix(admin-settings): broken update user space quota for users with sp…

fschade:
…ecial characters in name

## Description
Under certain circumstances, it was not possible to change the user space quota. For example, it failed if the username contained special characters...

The error occurred because Web had always sent the username when changing the quota... In the end, it all comes down to the fact that the wrong interface was used in the client, which required a name

## Related Issue
- Fixes https://github.com/opencloud-eu/opencloud/issues/789

## How Has This Been Tested?

- local installation

## Types of changes

- [X] Bugfix

### DIFF
--- a/packages/web-app-admin-settings/src/components/Users/SideBar/EditPanel.vue
+++ b/packages/web-app-admin-settings/src/components/Users/SideBar/EditPanel.vue
@@ -230,7 +230,6 @@ export default defineComponent({
       const updateSpace = await client.drives.updateDrive(
         editUser.drive.id,
         {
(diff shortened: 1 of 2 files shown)


fschade:
@AlexAndBear please have a look, its not clear to me if `onUpdateUserDrive` should take care of multiple property updates or just thwe quota as the condition indicates `... if (!isEqual(user.drive?.quota?.total, editUser.drive?.quota?.total)) { ...`

kulmann (approved):
LGTM 👍

PR #767, opened 2025-05-28
feat: show avatars in shares view

AlexAndBear:
## Related Issue

- Part of #598 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix
- [ ] Enhancement (a change that doesn't break existing code or deployments)
- [ ] Breaking change (a modification that affects current functionality)
- [ ] Technical debt (addressing code that needs refactoring or improvements)
- [ ] Tests (adding or improving tests)
- [ ] Documentation (updates or additions to documentation)
- [ ] Maintenance (like dependency updates or tooling adjustments)

PR #742, opened 2025-05-23
e2e tests. user profile photo

ScharfViktor:
related: #582 

test for: 
- set profile photo
- change profile photo
- delete profile photo
- check that profile photo exist/not exist

PR #792, opened 2025-06-03
e2e-tests. Check avatar tests in shares view

ScharfViktor:
related #767
addition to #742 

I check user's avatar in: 
- [x] notification
- [x] the shares view
- [x] the share with me page
- [x] the share with other page
- [x] activity panel
- [x] set user avatar using api (Given step)


ScharfViktor:
> As a followup we could also check the avatars in the activities panel

Right! for this place I forgot to add test. I'll do it in this PR

AlexAndBear (approved):

PR #793, opened 2025-06-03
feat: change visual representation of stacked avatars

AlexAndBear:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix


packages/design-system/src/components/OcAvatars/OcAvatars.vue
```diff
@@ -100,8 +106,10 @@ const {
   width = 30
 } = defineProps<Props>()
 
-const isOverlapping = computed(() => maxDisplayed && maxDisplayed < items.length)
+const avatarsRef = useTemplateRef('avatarsRef')
+const observer = null
```
AlexAndBear:
```suggestion
```

packages/design-system/src/components/OcAvatars/OcAvatars.vue
```diff
@@ -157,6 +164,18 @@ const getAvatarComponentForItem = (item: Item) => {
...
 })
+
+onMounted(() => {
+  if (!unref(avatarsRef)) {
+    return
+  }
+
+  const avatarElements = Array.from(unref(avatarsRef).children)
+
+  avatarElements.forEach((child, index) => {
```
kulmann, line 189: We can avoid the repaint. You can give the desired z-index as property to all the avatars individually and then inside oc-avatar use the prop for a css variable.

AlexAndBear: I am not sure if this is possible, since we have a slot for #userAvatars 🤔

AlexAndBear:
@kulmann I think it's finally time to close the chapter avatars ;)

packages/design-system/src/components/OcAvatars/OcAvatars.vue
```diff
@@ -167,16 +167,20 @@ const getAvatarComponentForItem = (item: Item) => {
...
     `oc-avatars-gap-${getSizeClass(gapSize)}`,
     ...(stacked ? ['oc-avatars-stacked'] : []),
-    ...(stacked && hoverEffect ? ['oc-avatars-hover-effect'] : [])
+    ...(unref(hasHoverEffect) ? ['oc-avatars-hover-effect'] : [])
   ]
 })
 
 onMounted(() => {
-  if (!unref(avatarsRef) || !stacked) {
+  if (!unref(avatarsRef) || !unref(hasHoverEffect)) {
```
kulmann, line 183: hm, this is not correct. If you want to have the avatars stacked, there are 5 of them but you don't want the hover effect, then the z-index won't be correct.

AlexAndBear: They do stack naturally without giving them a z index manually. 

Only if you want to interfere via the hover effect and want to change the z index (or change direction), this will be necessary

kulmann: aaaah, got it. ok, thank you & nevermind then 😅

kulmann (approved):
Awesome! :unicorn:

PR #796, opened 2025-06-04
[with-tracing] Upload tracing results of UI tests to public S3

individual-it:
## Description

When a PR contains `with-tracing` in the heading or a test fails, the tracing results will be uploaded to a public S3 bucket, so that they can be used for debugging.

The cleanup will be done in #838

## Related Issue

- Fixes https://github.com/opencloud-eu/qa/issues/24

## How Has This Been Tested?

- passing tests with `with-tracing` in the PR header
- failed tests without `with-tracing` in the PR header

## Types of changes

- [ ] Bugfix

### DIFF
--- a/tests/e2e/cucumber/environment/index.ts
+++ b/tests/e2e/cucumber/environment/index.ts
@@ -183,8 +183,8 @@ function filterTracingReports(status: string) {
     reports.forEach((report) => {
       fs.renameSync(`${traceDir}/${report}`, `${failedDir}/${report}`)
     })
-  } else {
-    // clean up the tracing directory
+  } else if (!defaults.reportTracing) {
+    // clean up the tracing directory if the report tracing was not set explicitly
     fs.rmSync(traceDir, { recursive: true })
   }
 }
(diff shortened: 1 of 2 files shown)


tests/e2e/cucumber/environment/index.ts
```diff
@@ -183,8 +187,8 @@ function filterTracingReports(status: string) {
     reports.forEach((report) => {
       fs.renameSync(`${traceDir}/${report}`, `${failedDir}/${report}`)
     })
-  } else {
-    // clean up the tracing directory
+  } else if (!defaults.reportTracing) {
+    // clean up the tracing directory if the report tracing was not set explicitly
     fs.rmSync(traceDir, { recursive: true })
```
individual-it, line 192: don't remove the traces of passed tests if we have enabled tracing explicitly

ScharfViktor (approved):
Nice to have this feature again 👍 
I already tested it with [with-tracing] tag. works fine. I can get and open trace-result

tests/e2e/cucumber/environment/index.ts
saw-jan: tracings are generated only if `config.reportTracing` is true. we have to introduce a new config to implement this properly

saw-jan: but yeah, this works fine for retried scenarios

saw-jan: let's keep the current implementation. I was just confused and getting many unwanted tracings while trying to debug only the failed tests

PR #803, opened 2025-06-06
fix: status column was shown in multiple places, it shouldn't

### LINKED ISSUE #801: Two "status" columns in space listing
### Initial Checklist

### Bug Description

With the latest code from web I get two "Status" columns in the space list

Image

### Reproduction Steps

Click on "Spaces" in the files app

### Expected Outcome

No duplicated columns

### Actual Outcome

duplicated columns

AlexAndBear:
## Related Issue

- Fixes #801 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/web-app-files/src/views/spaces/GenericSpace.vue
+++ b/packages/web-app-files/src/views/spaces/GenericSpace.vue
@@ -69,6 +69,7 @@
               :header-position="fileListHeaderY /* table */"
               :sort-fields="sortFields /* tiles */"
               :view-size="viewSize /* tiles */"
+              :fields-displayed="['name', 'size', 'indicators', 'tags', 'mdate']"
               :style="folderViewStyle"
               v-bind="folderView.componentAttrs?.()"
               @file-dropped="fileDropped"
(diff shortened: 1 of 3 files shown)


kulmann (changes requested):
Could we make the `enabled/disabled` status a resource indicator instead and then get rid of this `status` column entirely? Limiting the displayed fields doesn't sound very maintainable to me.

AlexAndBear:
> Could we make the `enabled/disabled` status a resource indicator instead and then get rid of this `status` column entirely? Limiting the displayed fields doesn't sound very maintainable to me.

I don't think so, the only place were status should be shown is in the files list and the spaces table. 

Currently the status column is also shown in places, where it shouldn't (or is buggy as the columns are empty):
- Shared with me table
- Shared with others table
- Shared via link table
- Deleted files -> space xyz table

AlexAndBear:
But maybe they should be there? IDK.

Iam opening up a new pr, where I merge space status into status indicators and we can decide later on in which table they should be displayed

Issue #625, opened 2025-05-02
UI for ranged selections has limited click areas


kulmann:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

After selecting one file and holding the shift-key I want to modify the selection range without worrying where exactly to click.

### Reproduction Steps

- create 5 files
- select file no. 2
- hold shift-key

now I want to select file no. 4 as end of my selection range.

### Expected Outcome

list view: anywhere I click in a list row the click is interpreted as selection-range modification
tiles view: anywhere I click on a tile the click is interpreted as selection-range modification

### Actual Outcome

list view:
- click on checkbox toggles only the selection of the clicked file, doesn't modify the selection range
- click on filename opens the file in viewer or editor, doesn't modify the selection range

tiles view:
- click on checkbox modifies the selection range
- click on icon (or preview, if the file has one) or file name opens the file in viewer or editor, doesn't modify the selection range
- click on whitespace in the tile doesn't do anything

PR #824, opened 2025-06-12
fix: Added advanced Shift-Click-Handling for Files

tammi-23:
Added advanced Shift-Click-Handling for Files

Related Issue #625


packages/web-app-files/src/components/FilesList/QuickActions.vue
```diff
@@ -8,7 +8,7 @@
       appearance="raw"
       class="oc-ml-xs quick-action-button oc-p-xs"
       :class="`files-quick-action-${action.name}`"
-      @click="action.handler({ space, resources: [item] })"
+      @click="(e: any) => action.handler({ space, resources: [item], event: e })"
```
kulmann: Why `any`? Is it not of type `MouseEvent`? 🤔

packages/web-pkg/src/components/FilesList/ResourceLink.vue
```diff
@@ -90,15 +93,17 @@ export default {
     }
   },
   methods: {
-    emitClick() {
+    emitClick(e: unknown) {
+      if (!e || typeof (e as MouseEvent).stopPropagation !== 'function') {
```
AlexAndBear: didn't checked the whole pr, but I would actually double check with mobile device bc we have different types of events here

packages/web-pkg/src/composables/actions/files/useFileActionsCopyPermanentLink.ts
```diff
@@ -22,15 +25,31 @@ export const useFileActionsCopyPermanentLink = () => {
...
+          eventBus.publish('app.files.list.clicked.shift', {
+            resource: resources[0],
+            skipTargetSelection: false
+          })
+          return
+        }
+
         const [resource] = resources
-        const permalink = resource.privateLink
+        const permalink = (resource as any).privateLink
```
kulmann: no `any` please

kulmann (commented):
Still some things off:
- tiles view: shift+click on preview opens the file in a new window but doesn't select the file (shift+click on filename works like that now, great!)
- cmd/ctrl+click triggers buttons while shift+click doesn't. We should be consistent, so the cmd/ctrl+click should also NOT trigger buttons. this is for tiles and list view.

packages/web-pkg/src/composables/keyboardActions/useInterceptModifierClick.ts
```diff
@@ -0,0 +1,31 @@
+import { eventBus } from '../../services'
+import type { Resource } from '@opencloud-eu/web-client'
+
+export function useInterceptModifierClick(event: MouseEvent, resource: Resource): boolean {
```
JammingBen: Generally, composables are more used a wrappers for functionality. So in this case, `useInterceptModifierClick` would expose a method (e.g. `interceptModifierClick`) that then takes these arguments and handles the logic.

You then use this composable in setup functions or inside other composables like this: `const { interceptModifierClick } = useInterceptModifierClick()`. Not directly inside methods, templates or computed properties, but in the root of a setup function or another composable. The method itself gets then called where you need to call it.

The same goes for the `useToggleTile` composable.

packages/web-pkg/src/components/FilesList/ResourceTile.vue
```diff
@@ -39,7 +39,12 @@
           :aria-label="tooltipLabelIcon"
         >
           <slot name="imageField" :item="resource">
-            <oc-image v-if="resource.thumbnail" class="tile-preview" :src="resource.thumbnail" />
+            <oc-image
+              v-if="resource.thumbnail"
+              class="tile-preview"
+              :src="resource.thumbnail"
+              @click="toggleTile([resource, $event], $event)"
+            />
```
JammingBen: Elements with a `@click` handler that usually are not supposed to have one should have a `role="button"` attribute. Or can we use a button here maybe? That would be the cleaner approach.

packages/web-pkg/src/components/FilesList/ResourceTile.vue
```diff
@@ -53,7 +58,7 @@
           </slot>
         </div>
       </resource-link>
-      <div class="oc-card-body oc-p-s">
+      <div class="oc-card-body oc-p-s" @click.passive="toggleTile([resource, $event], $event)">
```
JammingBen: Same here.

packages/web-pkg/src/components/FilesList/ResourceTile.vue
```diff
@@ -130,6 +137,10 @@ defineSlots<{
   selection?: (props: { item: Resource }) => unknown
 }>()
 
+function toggleTile(data: [Resource, MouseEvent | KeyboardEvent], event?: MouseEvent) {
```
JammingBen: Just for consistency inside setup functions:

```suggestion
const toggleTile = (data: [Resource, MouseEvent | KeyboardEvent], event?: MouseEvent) => {
```

packages/web-pkg/src/composables/actions/files/useFileActionsCopyPermanentLink.ts
```diff
@@ -15,23 +18,29 @@ export const useFileActionsCopyPermanentLink = () => {
...
       showErrorMessage({
         title: $gettext('Copy link failed'),
         errors: [e]
       })
     }
   }
-
+  interface FileActionOptionsWithEvent extends FileActionOptions<Resource> {
+    event?: MouseEvent
+  }
```
JammingBen: I would move this to the file where the `FileActionOptions` interface is located.

packages/web-pkg/src/composables/actions/files/useFileActionsShowShares.ts
```diff
@@ -8,14 +8,20 @@ import { useRouter } from '../../router'
...
+import { useInterceptModifierClick } from '../../keyboardActions'
 
 export const useFileActionsShowShares = () => {
   const router = useRouter()
   const { $gettext } = useGettext()
   const { canShare } = useCanShare()
   const resourcesStore = useResourcesStore()
 
-  const handler = ({ resources }: FileActionOptions) => {
+  const handler = ({ resources, event }: FileActionOptions & { event?: MouseEvent }) => {
```
JammingBen: You could use the `FileActionOptionsWithEvent` interface once you moved it.

packages/web-pkg/src/helpers/statusIndicators.ts
```diff
@@ -64,8 +65,10 @@ const getUserIndicator = ({
     category: 'sharing',
     type: isDirect ? 'user-direct' : 'user-indirect',
     fillType: 'line',
-    handler: () => {
-      eventBus.publish(SideBarEventTopics.openWithPanel, 'sharing#peopleShares')
+    handler: (resource: Resource, event?: MouseEvent) => {
+      if (event && useInterceptModifierClick(event, resource)) {
+        return
+      }
```
JammingBen: Did you intentionally remove `eventBus.publish(SideBarEventTopics.openWithPanel, 'sharing#peopleShares')` completely? 🤔

JammingBen (commented):
Works nicely! 🥳  Just a few things code-wise from my side.

packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -1024,8 +1055,19 @@ export default defineComponent({
     openTagsSidebar() {
       eventBus.publish(SideBarEventTopics.open)
     },
-    openSharingSidebar(file: Resource) {
-      let panelToOpen
+
+    handleFileClick(e: MouseEvent, resource: Resource) {
+      const { interceptModifierClick } = useInterceptModifierClick()
+      if (interceptModifierClick(e, resource)) return
```
JammingBen:
```suggestion
      if (this.interceptModifierClick(e, resource)) {
        return
      }
```

packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -1024,8 +1055,19 @@ export default defineComponent({
...
+      this.$emit('fileClick', { space: this.getMatchingSpace(resource), resources: [resource] })
+    },
+
+    openSharingSidebar(file: Resource, event?: MouseEvent) {
+      const { interceptModifierClick } = useInterceptModifierClick()
+      if (event instanceof MouseEvent && interceptModifierClick(event, file)) {
```
JammingBen:
```suggestion
      if (event instanceof MouseEvent && this.interceptModifierClick(event, file)) {
```

packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -1195,9 +1244,12 @@ export default defineComponent({
           .map((resource) => resource.id)
       )
     },
-    emitFileClick(resource: Resource) {
+    emitFileClick(resource: Resource, event?: MouseEvent) {
+      const { interceptModifierClick } = useInterceptModifierClick()
+      if (interceptModifierClick(event, resource)) {
```
JammingBen:
```suggestion
      if (this.interceptModifierClick(event, resource)) {
```

packages/web-pkg/src/components/FilesList/ResourceTile.vue
```diff
@@ -130,6 +142,10 @@ defineSlots<{
   selection?: (props: { item: Resource }) => unknown
 }>()
 
+const toggleTile = (data: [Resource, MouseEvent | KeyboardEvent], event?: MouseEvent) => {
+  useToggleTile(data, event)
```
JammingBen: Generally, composables should not be used inside functions. It can work, but it will not always and is considered bad practice. Ideally you do it the same like with `useInterceptModifierClick`, so it will eventually become:

```
const { toggleTile } = useToggleTile()
```

The `useToggleTile` composable then returns this function.

packages/web-pkg/src/composables/actions/files/useFileActionsCopyPermanentLink.ts
```diff
@@ -15,23 +18,28 @@ export const useFileActionsCopyPermanentLink = () => {
...
```
JammingBen: Can be removed now that you added it in `useFileActions.ts`.

```suggestion
```

packages/web-pkg/src/composables/actions/files/useFileActionsCopyPermanentLink.ts
```diff
@@ -15,23 +18,28 @@ export const useFileActionsCopyPermanentLink = () => {
...
       label: () => $gettext('Copy permanent link'),
-      handler: ({ resources }) => {
-        const [resource] = resources
-        const permalink = resource.privateLink
-        return copyLinkToClipboard(permalink)
+      handler: (options: FileActionOptionsWithEvent) => {
+        const { resources, event } = options
+        const resource = resources[0]
+
+        const { interceptModifierClick } = useInterceptModifierClick()
```
JammingBen: Same here, please no composable usage inside functions. You can just move this line to the top level of this composable and then use `interceptModifierClick` (like you already do).

packages/web-pkg/src/composables/actions/files/useFileActionsShowShares.ts
```diff
@@ -5,18 +5,26 @@ import { SideBarEventTopics } from '../../sideBar'
...
   const router = useRouter()
   const { $gettext } = useGettext()
   const { canShare } = useCanShare()
   const resourcesStore = useResourcesStore()
 
-  const handler = ({ resources }: FileActionOptions) => {
-    resourcesStore.setSelection(resources.map(({ id }) => id))
+  const handler = ({ resources, event }: FileActionOptionsWithEvent & { event?: MouseEvent }) => {
+    const resource = resources[0]
+    const { interceptModifierClick } = useInterceptModifierClick()
```
JammingBen: Same here. Can simply be moved outside of this function 🙂

packages/web-pkg/src/composables/selection/useToggleTile.ts
```diff
@@ -0,0 +1,30 @@
+import { Resource } from '@opencloud-eu/web-client'
+import { eventBus } from '../../services'
+import { useInterceptModifierClick } from '../../composables/keyboardActions'
+
+export function useToggleTile(data: [Resource, MouseEvent | KeyboardEvent], event?: MouseEvent) {
```
JammingBen: What you did with `useInterceptModifierClick` also applies here, meaning all the logic inside this composable is supposed to go into its own method. The composable then returns this method via an object.

Be careful though since you're using another composable here: `useInterceptModifierClick`. Which is totally fine, but keep in mind that it needs to be called at root level, not inside the method you're about to introduce.

packages/web-pkg/src/helpers/statusIndicators.ts
```diff
@@ -64,7 +65,10 @@ const getUserIndicator = ({
     category: 'sharing',
     type: isDirect ? 'user-direct' : 'user-indirect',
     fillType: 'line',
-    handler: () => {
+    handler: (resource: Resource, event?: MouseEvent) => {
+      const { interceptModifierClick } = useInterceptModifierClick()
+      if (event && interceptModifierClick(event, resource)) return
```
JammingBen: This is tricky... Technically, you shouldn't use a composable here for the reasons mentioned above. But you don't have a setup context here. So the proper (but also a bit ugly) way would be to pass this method down to this function and then call it.

packages/web-pkg/src/components/FilesList/ResourceTile.vue
```diff
@@ -120,6 +126,7 @@ const emit = defineEmits<{
   (e: 'click'): void
   (e: 'contextmenu', event: MouseEvent | KeyboardEvent): void
   (e: 'itemVisible'): void
+  (e: 'shift-select', resource: Resource): void
```
JammingBen: This seems like a left over from the past? I can't find this event firing.

```suggestion
```

packages/web-pkg/src/components/FilesList/ResourceTiles.vue
```diff
@@ -82,7 +82,7 @@
               class="oc-flex-inline oc-p-s"
               :disabled="isResourceDisabled(resource)"
               :model-value="isResourceSelected(resource)"
-              @click.stop.prevent="toggleTile([resource, $event])"
+              @click.stop.prevent="toggleTile([resource, $event], $event)"
```
JammingBen: `$event` is already being passed to `toggleTile()`, no need to pass it twice, no?

packages/web-pkg/src/composables/actions/files/useFileActionsShowShares.ts
```diff
@@ -5,18 +5,26 @@ import { SideBarEventTopics } from '../../sideBar'
...
+import { useInterceptModifierClick } from '../../keyboardActions'
 
 export const useFileActionsShowShares = () => {
   const router = useRouter()
   const { $gettext } = useGettext()
   const { canShare } = useCanShare()
   const resourcesStore = useResourcesStore()
+  const { interceptModifierClick } = useInterceptModifierClick()
+
+  const handler = ({ resources, event }: FileActionOptionsWithEvent & { event?: MouseEvent }) => {
```
JammingBen: This should be sufficient, `FileActionOptionsWithEvent` already includes `event`.

```suggestion
  const handler = ({ resources, event }: FileActionOptionsWithEvent) => {
```

packages/web-pkg/src/helpers/statusIndicators.ts
JammingBen, line 1: There is a `ResourceIndicator` interface inside this file. Please add your event to the signature of the `handler` method: `handler?: (resource: Resource, event?: MouseEvent) => void`.

packages/web-pkg/src/components/FilesList/ResourceLink.vue
```diff
@@ -90,15 +93,17 @@ export default {
     }
   },
   methods: {
-    emitClick() {
+    emitClick(e: unknown) {
+      if (!e || typeof (e as Event).stopPropagation !== 'function') {
```
JammingBen:
```suggestion
    emitClick(e: MouseEvent) {
      if (!e || typeof e.stopPropagation !== 'function') {
```

Also, why do you need to check for `stopPropagation` here?

packages/web-pkg/src/components/FilesList/ResourceLink.vue
```diff
@@ -8,7 +8,7 @@
     class="oc-resource-link"
     no-hover
     @dragstart.prevent.stop
-    @click="emitClick"
+    @click="(e: MouseEvent) => emitClick(e)"
```
JammingBen: It should work the way it was before.

```suggestion
    @click="emitClick"
```

JammingBen (commented):
A few final nitpick from my side.

packages/web-pkg/src/components/FilesList/ResourceTiles.vue
JammingBen: You also need to adjust `toggleTile` so it works with the one given event now, instead of the second one 😅

PR #820, opened 2025-06-11
feat: add open with context menu item

AlexAndBear:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix
- [ ] Enhancement (a change that doesn't break existing code or deployments)
- [ ] Breaking change (a modification that affects current functionality)
- [ ] Technical debt (addressing code that needs refactoring or improvements)
- [ ] Tests (adding or improving tests)
- [ ] Documentation (updates or additions to documentation)
- [ ] Maintenance (like dependency updates or tooling adjustments)

PR #843, opened 2025-06-23
[full-ci] fix app-provider tests

v-scharf:
related #820 and #835

we didn't run in `full-ci` so some tests from app-provider were broken. here is fix


ScharfViktor:
@tbsbdr in the in the right sidebar -> action,  user has `editorName`  previously it was called  `open in editorName`. Do you think this will cause confusion users? @AlexAndBear said that changing this is not so easy.

Screenshot 2025-06-23 at 10 38 44

ScharfViktor:
> Yes, that's good 👍

Thank you. I will do it in the next PR

ScharfViktor:
> > Yes, that's good 👍
> 
> Thank you. I will do it in the next PR

PR for that https://github.com/opencloud-eu/web/pull/845

PR #851, opened 2025-06-24
setup opencloud-keycloak-ldap setup. fix test after changing space template image

v-scharf:
- [x] fix test after changing space template. related https://github.com/opencloud-eu/opencloud/pull/1098
- [x] setup opencloud-keycloak-ldap setup. keycloak tests were failed. I added ldap 

I use Keycloak as IdP, and users are created there — they are stored in LDAP.
There are a couple of tests that currently create users or groups via the Graph API. I plan to rewrite those in the next PR. In my opinion, tests should not create users or groups through the Graph API — instead, we should rely on Keycloak api for that

### DIFF
--- a/.woodpecker.star
+++ b/.woodpecker.star
@@ -1422,15 +1422,6 @@ def postgresService():
 def ldapService():
     return [
         {
-            "name": "generate-ldap-certs",
-            "image": OC_CI_NODEJS,
-            "commands": [
-                "mkdir -p ldap-certs",
-                "openssl req -x509 -newkey rsa:2048 -keyout ldap-certs/openldap.key -out ldap-certs/openldap.crt -nodes -days 365 -subj '/CN=ldap-server'",
-                "chmod -R 777 ldap-certs",
-            ],
-        },
-        {
             "name": "ldap-server",
             "image": OPENLDAP,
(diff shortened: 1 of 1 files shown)


.woodpecker.star
```diff
@@ -1418,6 +1419,50 @@ def postgresService():
...
 
+def ldapService():
+    return [
+        {
+            "name": "generate-ldap-certs",
+            "image": OC_CI_NODEJS,
+            "commands": [
+                "mkdir -p ldap-certs",
+                "openssl req -x509 -newkey rsa:2048 -keyout ldap-certs/openldap.key -out ldap-certs/openldap.crt -nodes -days 365 -subj '/CN=ldap-server'",
+                "chmod -R 777 ldap-certs",
```
rhafer: Please never use `777` permissions anywhere. I guess `755` should do it here.

rhafer: Why do you need to create certificates here anyway. The custom entrypoint script you have for also creates a certificate if one does is not there.

One of the two openssl calls seems superfluous.

v-scharf: Yeah, that is duplicated. thank you, deleted it

Issue #858, opened 2025-06-25
Add A11Y statement URL to user menu


db-ot:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

As a visually impaired user, I want to have link to the official landingpage of the A11Y statement to see more details about a11y readyness of the website.

### Mockup

<img width="292" alt="Image" src="https://github.com/user-attachments/assets/0f79da4b-38cc-4846-9458-df0833c28924" />

### User Value

Visually impaired user, can easy reach the landingpage and inform about a11y readyness and contact

### Acceptance Criteria

- add a link into profilmenu bottom after "Privacy" link
- EN Link: https://opencloud.eu/en/accessibility-statement
- Translation: Barrierefreiheit / Accessibility
- Deploy on demo.opencloud.eu

db-ot:
updated AC with A11Y URL

AlexAndBear:
> updated AC with A11Y URL https://opencloud.eu/en/accessibility-statement

401

PR #861, opened 2025-06-25
feat: add accessibility config ling to user menu footer section

AlexAndBear:
## Related Issue

- Part of #858 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix


packages/web-runtime/src/components/Topbar/UserMenu.vue
```diff
@@ -121,6 +121,17 @@
         >
           <span v-text="$gettext('Privacy')" />
         </oc-button>
+        <span v-if="accessibilityUrl">·</span>
```
kulmann: What about a `<template v-if="accessibilityUrl">` as wrapper around the `span` and the `oc-button`?

AlexAndBear: Possible, but I don't wanted to break with the existing code, need to adjust it then too

packages/web-pkg/src/composables/piniaStores/theme.ts
```diff
@@ -15,7 +15,8 @@ const CommonSection = z.object({
     .object({
       accessDeniedHelp: z.string(),
       imprint: z.string(),
-      privacy: z.string()
+      privacy: z.string(),
+      accessibility: z.string().optional()
```
AlexAndBear, line 19:
```suggestion
      accessibility: z.string()
```

packages/web-pkg/src/composables/piniaStores/theme.ts
```diff
@@ -15,7 +15,8 @@ const CommonSection = z.object({
     .object({
       accessDeniedHelp: z.string(),
       imprint: z.string(),
-      privacy: z.string()
+      privacy: z.string(),
+      accessibility: z.string()
```
JammingBen: Shouldn't this be optional? Otherwise this would be a breaking change because it requires adjustments in the theme.json.

AlexAndBear: i had it optional, but was confused and removed it https://github.com/opencloud-eu/web/pull/861/commits/40a04b6964fd8281a7fe31c916cd987e0621cabc, bc the others ain't, any idea why this is the case ?

JammingBen: No clue 😅  IMO all of them could be optional, I don't see a reason for them to be required.

AlexAndBear: same, I go with that 👍

Issue #402, opened 2025-03-23
Add dark theme


BlitDev:
### Initial Checklist

- [x] I understand this is a feature request and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Idea

Currently only the light theme is available, so I'll leave the issue to keep an eye on when the dark theme is implemented/enabled, it's very important to me.

#137 #2

JammingBen:
Thanks for reporting this, we are aware of it and it's on our roadmap. Although it won't be part of the first stable release unfortunately.

Issue #513, opened 2025-04-07
Design finetuning incl. dark theme


kulmann:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

* > As a dark mode user, I want to have a dark theme so that my eyes don't hurt from all the light.
* > As OpenCloud we want to have the most beautiful and accessible design so that we can be proud of the design.

### User Value

Wow effect for all users

### Acceptance Criteria

- have a finetuned light theme
- have a dark theme (ping pong with agency)
- Creator of https://github.com/opencloud-eu/web/issues/402 is happy
- sharing sidebar spacings and sizings

tbsbdr:
further items that should get polished

![Image](https://github.com/user-attachments/assets/7dea2e27-9f64-4510-aadb-33f3e68ce4c3)

![Image](https://github.com/user-attachments/assets/5c524068-4242-41f0-945a-d6187fd7566c)

![Image](https://github.com/user-attachments/assets/87e6e297-bfe5-4425-89fc-55c28d58f4bc)

![Image](https://github.com/user-attachments/assets/6b169978-8a50-43ac-af61-a9b3f9291c3f)

db-ot:
What do you think of:

Explaination: using known elements and also direct on the place where user can change - without extra buttons left; bigger fontsize please ( password could be important )
<img width="556" alt="Image" src="https://github.com/user-attachments/assets/8191e2cb-04c3-49ac-ad01-4e4769485441" />

Explaination: alignment with spaces, more harmonizing 
<img width="550" alt="Image" src="https://github.com/user-attachments/assets/c862204a-4e3e-4bb6-915e-0af61916baf4" />

Explaination: a feeling says "use toggle" which user knows from other places
<img width="631" alt="Image" src="https://github.com/user-attachments/assets/9096c58a-f7c4-4d0e-b28f-e66229357722" />

db-ot:
fyi @tbsbdr Ticket is very blowing and become larger with all commented issue, we are agree the we should split here the issues also for estimation an working on from sprint to sprint, first agreed on "open" - context menu here:

https://github.com/opencloud-eu/web/issues/773

kulmann:
[...]

PR #912, opened 2025-07-07
feat: revert material design color hacks

kulmann:
## Description

Reverting color hacks that deviated our design away from material design. Download new theme.json and logos and test with a docker-compose overlay:

```yaml
services:
  opencloud:
    environment:
      WEB_ASSET_THEMES_PATH: ${WEB_ASSET_THEMES_PATH:-/web/themes}
      WEB_UI_THEME_PATH: ${WEB_UI_THEME_PATH:-/themes/opencloud/theme.json}
    volumes:
      - ../web-theme/:/web/themes/opencloud/
```

Download the following files and move them to the `web-theme` folder (or whichever folder you mount into the dev stack)

[theme.json](https://github.com/user-attachments/files/21103982/theme.json)

## Related Issue

Pre-requisite for merging our new theme in the opencloud repo, which will ultimately close https://github.com/opencloud-eu/web/issues/402 and works towards https://github.com/opencloud-eu/web/issues/513

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/web-pkg/src/components/FilesList/ResourceTile.vue
+++ b/packages/web-pkg/src/components/FilesList/ResourceTile.vue
@@ -1,7 +1,7 @@
 <template>
   <div
     ref="observerTarget"
-    class="oc-tile-card oc-card oc-card-default"
+    class="oc-tile-card oc-card"
     :data-item-id="resource.id"
     :class="{
       'oc-tile-card-selected': isResourceSelected,
@@ -196,8 +196,10 @@ if (!lazy) {
 </script>
 
 <style lang="scss">
+.oc-tile-card.oc-card {
+  background-color: var(--oc-role-surface-container);
+}
 .oc-tile-card {
-  background-color: var(--oc-role-surface-container) !important;
   box-shadow: none;
   height: 100%;
   display: flex;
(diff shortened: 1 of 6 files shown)


packages/web-pkg/src/components/FilesList/ResourceTile.vue
```diff
@@ -267,10 +269,10 @@ if (!lazy) {
     }
   }
   &:hover {
-    background-color: var(--oc-role-secondary-container) !important;
+    background-color: var(--oc-role-surface-container-highest);
```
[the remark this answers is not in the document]
kulmann: I explicitly removed the `!important` flag so that selected tiles won't have a hover color anymore. Made the selector for the tile background color more specific, so that it doesn't need an `!important` anymore as well.

packages/web-pkg/src/components/FilesList/ResourceTile.vue
```diff
@@ -196,8 +196,10 @@ if (!lazy) {
 </script>
 
 <style lang="scss">
+.oc-tile-card.oc-card {
+  background-color: var(--oc-role-surface-container);
+}
```
JammingBen, line 201: Nothing for now because it has always been like this, but having non-scoped and therefore global css instructions for the `oc-tile` class here feels a bit odd, we should improve that some day.

JammingBen (approved):
LGTM.

Though the left sidebar animation on click feels weird now because the nav highlighter gets white while moving and then settles behind the blue-ish current color.

packages/web-pkg/src/components/FilesList/ResourceTile.vue
kulmann: There's more to finetune on the technical side 😅 At this point it might also make sense to introduce an OcCard component, which then has a color prop for defining the background-color. We sometimes need that (most of the time we don't), and when we need it it always gets hacky.

AlexAndBear (approved):

PR #917, opened 2025-07-08
[full-ci] e2e tests. fix app-provider flaky test. Disable welcome popup

v-scharf:
- ~~Disabled the welcome pop-up window for tests. I tried different methods, but found only one that works: replace usr/share/coolwsd/browser/dist/welcome to empty folder~~

I found another way: param `--o:home_mode.enable=true` disables popup but limits number of concurrent open connections to 20 that I find not problem for our tests

here is from `coolwsd.xml`
```
<home_mode>
        <enable desc="Home users can enable this setting, which in turn disables welcome screen and user feedback popups, but also limits concurrent open connections to 20 and concurrent open documents to 10. The default means that number of concurrent open connections and concurrent open documents are unlimited, but welcome screen and user feedback cannot be switched off." type="bool" default="false">false</enable>
    </home_mode>
```
    
    
- second popup `our browser has very limited access to the clipboard, so please use the paste buttons on your on-screen keyboard instead` I handle with `tryCopyClipboard()`

how result: tests became faster 2m52 and, hopefully, more stable

### DIFF
--- a/.woodpecker.star
+++ b/.woodpecker.star
@@ -3,7 +3,7 @@ docker_repo_slug = "opencloudeu/web"
 
 ALPINE_GIT = "alpine/git:latest"
 APACHE_TIKA = "apache/tika:2.8.0.0"
-COLLABORA_CODE = "collabora/code:24.04.13.2.1"
+COLLABORA_CODE = "collabora/code:25.04.3.2.1"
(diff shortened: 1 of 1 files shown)


JammingBen (approved):
Thank you for taking care!

PR #931, opened 2025-07-10
alwaysUseVips

individual-it:
We are building openCloud for docker always with VIPS enabled.
Only the bare metal version is not using VIPS. So run the E2E tests always with VIPS enabled, because it renders more different image mimetypes.

fixes https://github.com/opencloud-eu/qa/issues/43


individual-it:
the `libdevs` package was enough. Sadly we have to do it for every run.
So here is the infrastructure question again: should we build and maintain an own fork of `docker.io/golang:1.24` or just suck the 64MB on every test run?
I will create an issue for that

ScharfViktor (approved):
Lgtm

Issue #865, opened 2025-06-26
Click on highlighted search result does not open the file unless clicked directly on the file element


tammi-23:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

In the search results, when a file is shown with a preview and the result is highlighted via mouse over, clicking anywhere within the highlighted area (but not directly on the file icon or name) does not trigger any action.

### Reproduction Steps

1. Open the search panel and perform a search that returns files with preview entries.

2. Hover over one of the results to highlight it.

3.  Click anywhere within the highlighted area, but not directly on the file icon or filename (e.g., click on the whitespace in the result row).

![Image](https://github.com/user-attachments/assets/18b820aa-461f-4e07-9be6-9f9d788d41bf)

### Expected Outcome

Clicking on the highlighted area of a search result should open the corresponding file, even if the click is not directly on the file icon or filename.

### Actual Outcome

Clicking on the non-interactive parts of the highlighted search result does nothing. The file only opens when the user clicks specifically on the file icon or filename.

kulmann:
Please note that the urls on file name and icon must stay. But I don't see a reason why a click on whitespace should do nothing. 

Technical question, maybe @db-ot knows this: is it a11y-conform to have a click handler on the whitespace which triggers the link as well? Or is it a requirement that the whitespace also is a link (= `<a href="...">` or `<router-link :to="...">`)?

db-ot:
Short answer: 

The area that appears as a link (including whitespace) must be contained in a real link element.

two source for A11Y-conform links are here:
https://www.w3.org/WAI/WCAG22/Understanding/link-purpose-in-context.html
https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html

JammingBen:
How should we proceed with this issue here? I don't see a solution that is technically doable while keeping the parent folder link within the list element that should be clickable. @kulmann what do you think?

PR #956, opened 2025-07-15
fix: ensure click on whitespace in search result opens file

tammi-23:
ensure click on whitespace in search result opens file.

Now also works for Keyboard select (ENTER)

Related to Issue #865

### DIFF
--- a/packages/web-app-search/src/portals/SearchBar.vue
+++ b/packages/web-app-search/src/portals/SearchBar.vue
@@ -84,6 +84,7 @@
                   active: isPreviewElementActive(providerSearchResultValue.id)
                 }"
                 class="preview oc-flex oc-flex-middle"
+                @click="openPreview($event)"
               >
                 <component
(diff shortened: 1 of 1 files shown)


packages/web-app-search/src/portals/SearchBar.vue
```diff
@@ -517,6 +518,17 @@ export default defineComponent({
...
+    },
+    openPreview(event: MouseEvent | KeyboardEvent) {
+      if ((event.target as HTMLElement).closest('a')) {
+        return
+      }
+
+      const container = event.currentTarget as HTMLElement
+      const link = container.querySelector<HTMLAnchorElement>('.oc-resource-link')
+      if (link) {
+        window.location.href = link.href
```
JammingBen, line 530: In general you want to avoid this because it circumvents our router. You notice that the whole page reloads when clicking on a search result, which defeats the purpose of our SPA.

PR #962, opened 2025-07-17
[with-tracing] add repo name in tracing path

individual-it:
## Description

As we also now uploading tracing results from the opencloud repo, it might be good to have also the repo name in the path of the trace result, because the pipeline number is only unique by repo

## Related Issue

follow up of https://github.com/opencloud-eu/qa/issues/28

## Types of changes

- [ ] Bugfix


ScharfViktor (approved):
👍

Issue #929, opened 2025-07-10
Mobile only: layout adjustments for bottom drawer


db-ot:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

As a user who uses the context menu mobile, I want to see the bottom drawer more modern, optimized and adjusted on mobilphone.

### User Value

- Touch-friendly – optimized for finger interaction
- Familiar layout pattern – matches standard mobile / app
- Space-saving – makes efficient use of limited screen space ( no white space )

### Acceptance Criteria

- The drawer opens from botton and closes with smooth, responsive animation bottom back.
- Remove also OS-Specific hints (shortcuts) as this is not needed mobile.
- Add border-radius to drawer elements
- Create a lightgrey background-basis for the drawer and 

please also check here some UI Experiences & Examples: 
https://m2.material.io/components/navigation-drawer#bottom-drawer

PR #985, opened 2025-07-21
Add bottom drawer for mobile devices

### LINKED ISSUE #904: Mobile only: Replace Context menu with bottom drawer
### User Story

> As a user who uses the context menu mobile, I want to see the name of the selected element(s), so that I'm reassured that Ive selected the right one(s).

> As a user who uses the context menu mobile, I want an explicit option to close the menu so that I know how to close it.

> As a user, I want a bottom drawer on my mobile device so that I can quickly access focus actions without leaving the current page, as this is a familiar and intuitive mobile experience.

> As a user, I want to easily open and close the bottom drawer by tapping or swiping, so that the interaction feels natural and comfortable.

> As a user, I want the bottom drawer to be responsive, well-organized and spaceoptimized, ensuring it is readable and usable across different mobile devices.

### User Value

- Touch-friendly – optimized for finger interaction
- Accessibility – screen reader and keyboard friendly
- Familiar interaction pattern – matches standard mobile / app UX behavior
- Space-saving – makes efficient use of limited screen space ( no surrounding white space )
- Contextual actions – displays only relevant options
- Easy to dismiss – can be closed quickly and intuitively

### Acceptance Criteria

- context menu gets replaced by bottom drawer on small mobile viewports (tablet as well?)
- Tapping the "X", click out or (swiping down -> will be a separate story) closes the entire drawer.
- Add File / Foldername on top of the drawer
- The drawer is a modal drawer (same as the current context menu actually)
- heigth of the drawer:
  - max. height is 66% of the screensize (unsure if this is a good value, lets play around and find out)
  - if there are only a few (2-3 options) in the first menu, the drawer height adapt to the height of these 2 options -> drawer is smaller than 66%
  - to reduce jumpyness:
    - submenus have at least the same height like its parents (usually 66%)  
- Only one menu level is visible at a time to reduce cognitive load.
- The drawer must follow touch-target size guidelines (min. 48×48dp). ( not 100% sure )
- If the content of the drawer requires scrolling, the last visible item must be cut off, so that the user knows that there's more.

please also check here some UI Experiences & Examples: 
https://m2.material.io/components/navigation-drawer#bottom-drawer

## Problem

resource name is not visible any more:

Image

### Mockups

will follow - direct examples for opencloud

Current place where we can show bottom drawer:
1. context menu for files / folder
2. context menu for usersettings in admin
3. context menu in spaces administration
4. context menu in groups administration
5. maybe also creating a new document behind the + button

Image

Image

Image

Image

Suggestion:

Image

With x and namefile

Image

AlexAndBear:
## Description

image

## Related Issue

- Fixes #904 
- Works towards #929 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/design-system/src/components/OcBottomDrawer/OcBottomDrawer.spec.ts
+++ b/packages/design-system/src/components/OcBottomDrawer/OcBottomDrawer.spec.ts
@@ -1,0 +1,136 @@
+import { defaultPlugins, mount } from '@opencloud-eu/web-test-helpers'
+import BottomDrawer from './OcBottomDrawer.vue'
+import { defineComponent, nextTick } from 'vue'
+import OcButton from '../OcButton/OcButton.vue'
+
+const selectors = {
+  toggle: '#button-drawer-toggle',
+  drawer: '.oc-bottom-drawer',
+  closeButton: '.oc-bottom-drawer-close-button',
+  background: '.oc-bottom-drawer-background',
+  actionButton: '#action-button'
+}
+
+describe('OcBottomDrawer', () => {
+  afterEach(() => {
+    document.body.innerHTML = ''
+  })
+
+  it('renders when toggle is clicked', async () => {
+    const { wrapper } = getWrapper()
+    await wrapper.find(selectors.toggle).trigger('click')
+    expect(wrapper.find(selectors.drawer).exists()).toBe(true)
+    expect(wrapper.html()).toMatchSnapshot()
+    wrapper.unmount()
+  })
+
+  it('does not render when toggle is not clicked', async () => {
+    const { wrapper } = getWrapper()
+    await nextTick()
+    expect(wrapper.find(selectors.drawer).exists()).toBe(false)
+  })
+
+  it('closes when close button is clicked', async () => {
+    const { wrapper } = getWrapper()
+
+    await wrapper.find(selectors.toggle).trigger('click')
+    expect(wrapper.find(selectors.drawer).exists()).toBe(true)
+
+    wrapper.find(selectors.closeButton).trigger('click')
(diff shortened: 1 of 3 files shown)


packages/web-app-files/src/components/SideBar/Shares/Collaborators/EditDropdown.vue
```diff
@@ -61,6 +63,7 @@ import { useConfigStore, useModals, DatePickerModal } from '@opencloud-eu/web-pk
 import { useGettext } from 'vue3-gettext'
 import { RouteLocationNamedRaw } from 'vue-router'
 import ContextMenuItem from './ContextMenuItem.vue'
+import { $gettext } from '@opencloud-eu/web-pkg/src/router/utils'
```
AlexAndBear: 🤦

packages/design-system/src/composables/useIsMobile/index.ts
```diff
@@ -0,0 +1,14 @@
+import { computed } from 'vue'
+import { useWindowSize } from '@vueuse/core'
+
+export const useIsMobile = () => {
+  const { width } = useWindowSize()
+
+  const isMobile = computed(() => {
+    return width.value < 768
```
JammingBen: Unsure if `768` is the right number here since we're using `640` as mobile breakpoint in the runtime. I think it makes sense to align that. But maybe we need more refinement of the breakpoints in general.

AlexAndBear: I think for now 640 is perfectly fine 👍
We can iterate on the breaks points if we are deeper into that mobile topic

PR #954, opened 2025-07-14
cross browser testing

ScharfViktor:
related #890

- [x] Run tests `suite1` (journeys,smoke) and `a11y` in Firefox, Chromium, and WebKit browsers
- [x] Adapt the test for uploading a folder in Firefox
- [x] Ignore the ResizeObserver console error in WebKit — it seems harmless
- [x] Update nodejs to version 22
- [x] `pnpm playwright install-deps webkit` I do this in the test pipeline (temporary solution). A good way to do this is to install all dependencies in the Docker image https://github.com/opencloud-eu/qa/issues/54
- [x] skip uploading empty folder for webkit browser https://github.com/opencloud-eu/web/issues/821
- [x] skip tests for webkit which contain this step
```
When "Alice" creates the following resources
      | resource    | type    | content   |	      | resource    | type    | content   |
      | example.txt | txtFile | some text |	      | example.txt | txtFile | some text |
```
in the CI I couldn't get a visible element on main page after creating txt file in the editor and after that closing editor
image

- [x] skipped checking console error for firefox and webkit

```
@sse @webkit-skip
  Scenario: sse events on file operations                                                                        # tests/e2e/cucumber/features/smoke/sse.feature:168
    ✔ When "Brian" navigates to the trashbin of the project space "marketing"                                    # file:/woodpecker/src/github.com/opencloud-eu/web/web/tests/e2e/cucumber/steps/ui/spaces.ts:130
[UNCAUGHT EXCEPTION] "uncaught exception: Object"
    ✘ And "Brian" restores the following resources from trashbin" restores the following resources from trashbin #
```
image
npx playwright show-trace https://s3.ci.opencloud.eu/public/web/tracing/3035/sse-events-on-file-operations-brian-2025-7-22-03-16-52.zip

PR #1006, opened 2025-07-24
Mobile view tests

ScharfViktor:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix


tests/e2e/support/objects/app-files/resource/actions.ts
```diff
@@ -576,6 +575,7 @@ export const editTextDocument = async ({
     page.locator(saveTextFileInEditorButton).click()
   ])
   await editor.close(page)
+  await page.reload()
```
v-scharf, line 578: need to avoid issue with webkit

JammingBen (approved):
LGTM.

We also should add tests for the recently added [bottom drawer](https://github.com/opencloud-eu/web/issues/904) at some point. But it makes sense to wait until https://github.com/opencloud-eu/web/pull/1017 is done.

tests/e2e/support/objects/app-files/resource/actions.ts
individual-it: will this cover a real bug? then we shouldn't do it

v-scharf: No, it is not possible to reproduce this locally, only in CI.
this issue: 
issue after closing text editor. similar was in the #954
Screenshot 2025-07-29 at 15 09 01

PR #1046, opened 2025-08-04
fix extension registry stable order

kulmann:
## Description

While reviewing https://github.com/opencloud-eu/web-extensions/pull/166 we found out that the order of applications during application loading is not stable. Instead of coming up with a sorting/weight concept we've decided to just maintain a stable order. This PR does that with the least possible performance impact:
- loading application scripts happens in parallel
- only building and registering the applications happens one by one, with the order defined by the config.json (this ran in parallel before, but produced a random order)
- calling the `initialize` hook on all applications happens in parallel again

## How Has This Been Tested?

Manually and with the existing unit tests.

## Types of changes

- [x] Bugfix

### DIFF
--- a/packages/web-runtime/src/container/bootstrap.ts
+++ b/packages/web-runtime/src/container/bootstrap.ts
@@ -1,5 +1,5 @@
 import { registerClient } from '../services/clientRegistration'
-import { buildApplication, NextApplication } from './application'
+import { buildApplication, loadApplication, NextApplication } from './application'
 import { RouteLocationRaw, Router, RouteRecordNormalized } from 'vue-router'
 import { App, computed, watch } from 'vue'
 import { loadTheme } from '../helpers/theme'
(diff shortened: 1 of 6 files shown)


AlexAndBear (approved):

packages/web-runtime/src/container/bootstrap.ts
```diff
@@ -233,30 +242,53 @@ export const initializeApplications = async ({
...
-    }
+      return acc
+    },
+    []
+  )
 
-    return acc
-  }, [])
+  // create applications in a stable order, one by one, so that apps and extensions get registered in the same order every time.
+  const applications = []
```
JammingBen: I guess it would make sense to type them?

JammingBen (approved):

packages/web-runtime/src/container/bootstrap.ts
kulmann: Yep, right. Done 👍

Issue #1037, opened 2025-07-31
Implement paste from Clipboard


TheOneRing:
### Initial Checklist

- [x] I understand this is a feature request and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Idea

When taking screenshots, I want to be able to paste them directly to OpenCloud.

JammingBen:
Yep, https://github.com/opencloud-eu/web/blob/main/packages/web-app-files/src/components/AppBar/CreateAndUpload.vue#L321 should handle this. Though it's been a long time since I've tried it, and it has some limitations if I remember correctly.

AlexAndBear:
Seems to be broken tough

PR #1047, opened 2025-08-04
fix: pasting files from local clipboard

JammingBen:
This didn't work before because the event listener for the ctrl/cmd + V shortcut was preventing the default. Adds an option to configure this and disables it for this shortcut.

fixes https://github.com/opencloud-eu/web/issues/1037


AlexAndBear:
@JammingBen or @ScharfViktor is it possible to write an e2e test for this feature ?

PR #1063, opened 2025-08-07
feat: change bottom drawer to less technical context menu label

kulmann:
change `Close the bottom drawer` label to a less technical `Close the context menu`.

### DIFF
--- a/packages/design-system/src/components/OcBottomDrawer/OcBottomDrawer.vue
+++ b/packages/design-system/src/components/OcBottomDrawer/OcBottomDrawer.vue
@@ -5,7 +5,7 @@
         v-show="show"
         class="oc-bottom-drawer-background"
         role="button"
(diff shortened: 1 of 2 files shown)


packages/design-system/src/components/OcBottomDrawer/OcBottomDrawer.vue
```diff
@@ -9,7 +9,6 @@
       ref="bottomDrawerRef"
       class="oc-bottom-drawer-background"
       role="button"
-      :aria-label="$gettext('Close the bottom drawer')"
```
kulmann, line 12: I removed the aria-label, because it would replace (!) the accessible content of the entire `<div>`. Don't know if it's needed at all, but if yes, we could use `aria-describedby` instead. I'd prefer to leave it out. Also not sure if we really need the `button` role here.

JammingBen (approved):

Issue #937, opened 2025-07-11
Introduce Tailwind CSS for layouting (viewport sizes, margins, paddings, etc)


kulmann:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

> * As a developer, I want to use a rock solid layouting system so that I don't need to fix all the edge cases myself.
> * As a user I want a fast UI so that I dont have to wait.
> * As a company, we want the web UI to be fast so that users can fully experience the performance of OpenCloud.

### User Value (developer facing)

- **Don't re-invent the wheel.** We want as little custom code as possible for being able to develop really fast with high quality for the upcoming groupware components.
- **Speed**: Experience the performance of OpenCloud.

### Acceptance Criteria

- Tailwind CSS is used throughout the design system and the web mono-repo as a whole
- No more oc-p (padding), oc-m (margin), oc-flex (flexbox) is used (this could be done incrementally)
- Adjust the design system docs
- Delete the respective design system css classes
- Delete the sizing tokens of the design system (also in the theme.json in the opencloud repo if there is something about it).

tbsbdr:
added **speed** as userstory/value

kulmann:
Followup issues:

## Activities sidebar panel

- [ ] activities are missing horizontal gap or margin between `timeline-dot` and activity content (use the same as with versions, which is also smaller than before, but fine IMO)
- [ ] activity date and avatar are not left-aligned

<img width="437" height="204" alt="Image" src="https://github.com/user-attachments/assets/d7e6296b-f4ff-4673-afef-bf97c6d9bd74" />

## App top bar within apps

- [ ] paddings are too small, especially the bottom padding on the parent folder name / distance to the bottom border of the app top bar file info thingy border

<img width="298" height="53" alt="Image" src="https://github.com/user-attachments/assets/0ecef9c4-fff4-4ac5-b83e-112f18a2b485" />

tammi-23:
### Followup issues (tested with FireFox):

## Generally
- [ ] border-radius is lower then before

<img width="1060" height="308" alt="Image" src="https://github.com/user-attachments/assets/c4002199-baaa-49bc-b508-2f845ce542bf" />

PR #1073, opened 2025-08-11
feat: add tailwindcss and replace oc-m/p classes

JammingBen:
Introduces Tailwind CSS.

Also replaces our margin and padding utility classes with Tailwind ones. The overall feel and look of the UI should be roughly the same as before. Maybe you'll find some minor differences in spacing due to them being more streamlined and many parts still missing.

The next step is to remove all margins and paddings from custom css and replace them with either Tailwind utility classes or by moving them to Tailwind's layer system.

The design-system docs also need some more work, I didn't get it to run well with Tailwind so far. Needs some more investigation and a follow-up PR.

#### Migration

- `oc-m-xs` > `m-1`
- `oc-m-s` > `m-2`
- `oc-m-m` > `m-4`
- `oc-m-l` > `m-6`
- `oc-m-xl` > `m-12`
- `oc-m-xxl` > `m-24`
- `oc-m` > `m-4`
- `oc-m-rm` > `m-0`

### DIFF
--- a/packages/design-system/src/components/OcButton/OcButton.vue
+++ b/packages/design-system/src/components/OcButton/OcButton.vue
@@ -176,7 +176,7 @@ const onClick = (event: MouseEvent) => {
 
 @layer components {
   .oc-button:not(.oc-button-raw, .oc-button-raw-inverse) {
-    @apply py-[0.5rem] px-[0.8rem];
+    @apply py-1.5 px-2.5;
   }
 }
 </style>


packages/web-runtime/src/pages/accessDenied.vue
```diff
@@ -2,7 +2,7 @@
   <div class="oc-height-viewport oc-flex oc-flex-column oc-flex-center oc-flex-middle">
     <div class="oc-login-card">
       <img class="oc-login-logo" :src="logoImg" alt="" :aria-hidden="true" />
-      <div class="oc-login-card-body oc-width-medium">
+      <div class="oc-login-card-body oc-width-medium p-6">
```
kulmann, line 5: why is the p-6 needed here? 🤔

packages/web-runtime/src/pages/logout.vue
```diff
@@ -2,19 +2,19 @@
   <div class="oc-height-viewport oc-flex oc-flex-column oc-flex-center oc-flex-middle">
     <div class="oc-login-card">
       <img class="oc-login-logo" :src="logoImg" alt="" :aria-hidden="true" />
-      <div class="oc-login-card-body oc-width-medium">
+      <div class="oc-login-card-body oc-width-medium p-6">
```
kulmann, line 5: Same here again, why the `p-6`? Didn't you add the padding via the tailwind.css file? Same for other pages that use the oc-card classes.

packages/web-runtime/src/pages/accessDenied.vue
JammingBen: Before this was set via css manually through the `oc-login-card-body` class.

packages/web-runtime/src/pages/logout.vue
JammingBen: Only the `oc-card` classes are styled globally, but this is `oc-login-card`. No idea why we differentiate here to be honest...

kulmann: ah, urgh. ok, nevermind then 🙈

PR #1081, opened 2025-08-14
refactor: rename defaultExtension to defaultAppId

kulmann:
## Description

Renames the `defaultExtension` option to `defaultAppId` since that's a more appropriate name for the option. The old name comes from a time where we thought that we'd ship extensions individually, not piggyback with an app.

## Related Issue

Idea came up in https://github.com/opencloud-eu/web-extensions/pull/177

## Types of changes

- [ ] Bugfix


kulmann:
Note: not marked as breaking change because the backend didn't offer the possibility to set the option at all so far (will do a PR next). Only custom config.json would have given the chance to set this option and I assume nobody did that.

JammingBen (approved):

PR #1093, opened 2025-08-19
refactor: font-weight and text-decoration to tailwind

JammingBen:

#### Migration

- `oc-font-semibold` > `font-semibold`
- `oc-text-bold` > `font-semibold`

See here for the weight values: https://tailwindcss.com/docs/font-weight

For `text-decoration` we didn't have a utility class. When a decoration is needed on hover, just use `hover:underline`. Disabling the underline is usually not needed since this is already done via the tailwind preflight.

### DIFF
--- a/tests/e2e/support/objects/app-files/resource/actions.ts
+++ b/tests/e2e/support/objects/app-files/resource/actions.ts
@@ -72,7 +72,7 @@ const globalSearchDirFilterDropdown =
 const globalSearchBarFilterAllFiles = '//*[@data-test-id="all-files"]'
 const globalSearchBarFilterCurrentFolder = '//*[@data-test-id="current-folder"]'
 const searchList =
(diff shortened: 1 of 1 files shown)


AlexAndBear (approved):

packages/design-system/src/components/OcInfoDrop/OcInfoDrop.vue
```diff
@@ -24,7 +24,11 @@
             :is="item.headline ? 'dt' : 'dd'"
             v-for="(item, index) in listItems"
             :key="index"
-            :class="{ 'ml-0': !item.headline, 'first:mt-0': item.headline }"
+            :class="{
```
kulmann, line 27: The text of contextual helpers is now bold. That looks weird. Was normal weight before.

PR #1105, opened 2025-08-21
chore: adjust flexbox css props to tailwind followup

AlexAndBear:
## Related Issue

- Works towards https://github.com/opencloud-eu/web/issues/937

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix


packages/web-pkg/src/components/FilesList/ResourceTiles.vue
```diff
@@ -673,14 +673,5 @@ onBeforeUnmount(() => {
 
 .ghost-tile {
   display: list-item;
-
-  div {
```
AlexAndBear, line 677: Doesn't even exist 🤔

packages/design-system/src/components/OcColorInput/OcColorInput.vue
```diff
@@ -34,9 +34,9 @@
     </div>
     <div
       v-if="showMessageLine"
-      class="oc-color-input-message text-sm mt-1"
+      class="oc-color-input-message flex items-center text-sm mt-1"
       :class="{
-        'oc-color-input-description': !!descriptionMessage,
+        'oc-color-input-description flex items-center': !!descriptionMessage,
```
JammingBen: No need to set it twice, once in `class` should be good.

packages/design-system/src/components/OcTextInput/OcTextInput.vue
```diff
@@ -43,9 +43,9 @@
     </div>
     <div
       v-if="showMessageLine"
-      class="oc-text-input-message text-sm mt-1"
+      class="oc-text-input-message flex align-center text-sm mt-1"
       :class="{
-        'oc-text-input-description': showDescriptionMessage,
+        'oc-text-input-description flex items-center': showDescriptionMessage,
```
JammingBen: Same here.

packages/web-pkg/src/components/ViewOptions.vue
```diff
@@ -298,6 +298,14 @@ export default defineComponent({
   }
 })
 </script>
+<style>
+@reference 'tailwindcss';
+
+.files-view-options-list-item > * {
+  @apply flex justify-between;
+}
```
JammingBen, line 307:
```suggestion
@layer utilities {
  .files-view-options-list-item > * {
    @apply flex justify-between;
  }
}
```

packages/web-runtime/src/components/Modals/AppTokenModal.vue
```diff
@@ -41,7 +41,7 @@
     />
     <div class="mt-4 mb-2 flex items-center oc-rounded">
       <div class="created-token-container">
-        <div class="created-token oc-rounded p-2 font-bold">
+        <div class="created-token flex items-center oc-rounded p-2 font-bold">
```
JammingBen: I think this got lost.

```suggestion
        <div class="created-token flex items-center justify-between oc-rounded p-2 font-bold">
```

packages/web-runtime/src/components/SidebarNav/SidebarNavItem.vue
```diff
@@ -4,7 +4,15 @@
       :type="handler ? 'button' : 'router-link'"
       :appearance="active ? 'filled' : 'raw-inverse'"
       color-role="secondaryContainer"
-      :class="['oc-sidebar-nav-item-link', 'oc-oc-width-1-1', 'p-2', { active: active }]"
+      :class="[
+        'oc-sidebar-nav-item-link',
+        '!flex',
+        '!items-center',
+        '!justify-between',
```
JammingBen:
```suggestion
        'flex',
        'items-center',
        'justify-between',
```

packages/web-pkg/src/components/ViewOptions.vue
AlexAndBear: Did you test that? In my try/opinion that won't work

JammingBen: I didn't try, but it should work. If not, we should try and make it work 😅 There are only very few exceptions where styles must not be inside a layer (overwriting vendor styles or button mixins).

AlexAndBear: I checked this and it won't work, chad gpts explenation was that mixing layers and > * are invalid

JammingBen: Hmm okay. Can you then at least add a comment why this is an exception and should not be part of any layer?

packages/web-runtime/src/components/SidebarNav/SidebarNavItem.vue
AlexAndBear: w-o the important statement, it breaks

JammingBen (approved):
👍

PR #1115, opened 2025-08-25
fix(design-system): OcSelect: reintroduce getOptionLabel prop

dschmidt:
## Description

The `getOptionLabel` was removed when OcSelect was ported to `script setup`.
I assume this happened because it's not possible to reference another prop (`optionLabel`) from `withDefaults` - at least I couldn't convince the Vue SFC compile to handle it.

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [x] Bugfix

### DIFF
--- a/packages/design-system/src/components/OcSelect/OcSelect.spec.ts
+++ b/packages/design-system/src/components/OcSelect/OcSelect.spec.ts
@@ -59,14 +59,14 @@ describe('OcSelect', () => {
       )
     })
     it('displays with a custom label function', () => {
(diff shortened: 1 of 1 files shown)


dschmidt:
An alternative could be to make the `optionLabel` prop `string | (unknown) => string` and handle that internally.
Personally I would prefer that API but it breaks with our former api and with vue-select. Depends on how much of a lightweight wrapper or facade we want to be for vue-select.

kulmann (approved):

Issue #1011, opened 2025-07-25
Add config to disable "Login allowed" for users


JammingBen:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

As an admin who manages users I don't want the option to dis- or enable user logins if this has no effect and is managed by the IDP so that I don't get confused.

### User Value

- Reduces confusion for admins and users
- Prevents false assumptions about feature functionality.

### Acceptance Criteria

- OpenCloud has a new config to disable the feature to enable/disable user logins (defaults to false = not disabling the feature)
- Web consumes the config and hides the corresponding UI parts in the admin settings if necessary:
  - Action when selecting user(s)
  - Column in user table
  - Right sidebar details panel
  - Right sidebar edit panel

kulmann:
last step is done via https://github.com/opencloud-eu/web/pull/1121 - closing here. But the web part needs to be released and then we need QA if everything works as required.

PR #1121, opened 2025-08-26
feat: hide Login allowed for users

tammi-23:
Feat: hide "login allowed/forbidden" for users if users managed extern.

Ref Issue: #1011


packages/web-app-admin-settings/src/composables/useAdminUsersFlags.ts
```diff
@@ -0,0 +1,20 @@
...
+import { storeToRefs } from 'pinia'
+import { useCapabilityStore } from '@opencloud-eu/web-pkg'
+
+type Caps = {
+  graph?: { users?: { edit_login_allowed_disabled?: boolean } }
+}
+
+export function useAdminUsersFlags(): {
+  isLoginToggleHidden: ComputedRef<boolean>
+} {
```
JammingBen: No need to type that explicitly, TS infers the correct type automatically.

```suggestion
export function useAdminUsersFlags() {
```

packages/web-app-admin-settings/src/composables/useAdminUsersFlags.ts
```diff
@@ -0,0 +1,20 @@
...
+  const capabilityStore = useCapabilityStore()
+  const { capabilities } = storeToRefs(capabilityStore)
+
+  const isLoginToggleHidden = computed<boolean>(() => {
+    return (capabilities.value as Caps)?.graph?.users?.edit_login_allowed_disabled === true
```
JammingBen: Please add `edit_login_allowed_disabled` to the `Capabilities` interface in `packages/web-client/src/ocs/capabilities.ts`. Then you can remove `Caps` here.

Also I think you should add a default value in `packages/web-pkg/src/composables/piniaStores/capabilities.ts` and a computed in the store in the same file (e.g. `graphUsersEditLoginAllowedDisabled`) - check the others for examples.

packages/web-app-admin-settings/src/components/Users/SideBar/EditPanel.vue
```diff
@@ -201,6 +201,12 @@ export default defineComponent({
     const isInputFieldReadOnly = (key: string) => {
       return capabilityStore.graphUsersReadOnlyAttributes.includes(key)
     }
+    const isLoginToggleHidden = computed(() => {
+      const list = (capabilityStore.graphUsersReadOnlyAttributes ?? []).map((s: string) =>
+        s.trim().toLowerCase()
+      )
+      return list.includes('user.accountenabled') || list.includes('accountenabled')
+    })
```
JammingBen: You can simply use `capabilityStore.graphUsersEditLoginAllowedDisabled` once you implemented the computed in the store, which make this one obsolete.

packages/web-app-admin-settings/src/components/Users/SideBar/DetailsPanel.vue
```diff
@@ -98,12 +102,19 @@ export default defineComponent({
...
-      return language.current
+    const currentLanguage = computed(() => language.current)
+
+    const capabilityStore = useCapabilityStore()
+    const isLoginToggleHidden = computed(() => {
+      const list = (capabilityStore.graphUsersReadOnlyAttributes ?? []).map((s: string) =>
+        s.trim().toLowerCase()
+      )
+      return list.includes('user.accountenabled') || list.includes('accountenabled')
     })
```
JammingBen: Same here.

JammingBen (changes requested):

PR #1130, opened 2025-08-28
chore: adjust position css props to tailwind

AlexAndBear:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix


packages/web-runtime/src/pages/oidcCallback.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="oc-login-card oc-position-center">
+  <div class="oc-login-card absolute top-[50%] left-[50%]">
```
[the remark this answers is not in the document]
JammingBen: No needed here.

packages/design-system/src/components/OcList/OcList.vue
```diff
@@ -55,26 +63,17 @@ defineSlots<Slots>()
...
   li {
-    position: relative;
     box-sizing: border-box;
 
     &::before {
       content: '';
-      position: absolute;
-      left: -4px;
-      top: 50%;
       transform: translateY(-50%);
```
kulmann: arguable, but I'd move the transform into the tailwind style block in the scope of this PR as well (same for some other files)

packages/design-system/src/components/OcNotifications/OcNotifications.spec.ts
```diff
@@ -5,17 +5,6 @@ describe('OcNotifications', () => {
   function getWrapper(options = {}) {
     return shallowMount(OcNotifications, options)
   }
-  describe('position prop', () => {
```
kulmann, line 8: instead of deleting the test, you could just check for `left-2`, `right-2` or `mx-auto` instead. Or with a prefix-match for `left-` and `right-`.

packages/web-pkg/src/components/SideBar/SideBar.vue
```diff
@@ -285,11 +285,23 @@ onBeforeUnmount(() => {
   }
 })
 </script>
+<style>
+@reference '@opencloud-eu/design-system/tailwind';
 
+@layer components {
+  .sidebar-panel.is-root-panel {
+    right: 100px;
```
kulmann, line 296: do we care about consistency here? is `right: 100px` fine or do we want to work with classes wherever possible, i.e. `@apply right-[100px];`?

packages/web-runtime/src/components/Topbar/ApplicationsMenu.vue
```diff
@@ -130,11 +130,6 @@ export default defineComponent({
   a,
   button {
     gap: var(--oc-space-medium) !important;
-
-    .active-check {
```
kulmann, line 134: was the class unused?

packages/web-runtime/src/layouts/Application.vue
```diff
@@ -40,7 +40,9 @@
       <portal-target name="app.runtime.bottom.drawer" :multiple="true" />
       <portal-target name="app.runtime.footer" />
     </div>
-    <div class="snackbars mx-auto sm:m-0">
+    <div
+      class="snackbars absolute inset-x-0 sm:left-auto sm:right-[20px] bottom-[20px] mx-auto sm:m-0"
```
kulmann: I'd prefer this:


```suggestion
      class="snackbars absolute inset-x-[20px] sm:left-auto bottom-[20px]"
```

and I don't know where you had the margins from... seems to work fine without the margins.

packages/web-runtime/src/components/Topbar/ApplicationsMenu.vue
JammingBen: Yes. I guess we had a check icon in the past for the active icon, but this is long gone.

packages/web-pkg/src/components/SideBar/SideBar.vue
JammingBen: As discussed in chat: `@apply` doesn't really give us any advantages over pure CSS when working a) with custom values *and* b) in CSS directly at the same time. I'd prefer pure CSS in such scenarios.

PR #1132, opened 2025-08-29
fix(design-system): docs after tailwind migration

JammingBen:

### DIFF
--- a/packages/design-system/docs/.vitepress/theme/custom.scss
+++ b/packages/design-system/docs/.vitepress/theme/custom.scss
@@ -22,19 +22,6 @@ html * {
   -webkit-font-smoothing: antialiased;
 }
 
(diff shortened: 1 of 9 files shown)


packages/design-system/docs/.vitepress/theme/custom.scss
```diff
@@ -3,22 +3,36 @@
...
+
+html * {
+  text-rendering: optimizeLegibility;
+  -webkit-font-smoothing: antialiased;
 }
-p.oc-recipient-name {
-  line-height: unset !important;
+
+// some vitepress styles need to be overwritten to properly display our components
+.live-code-block {
```
kulmann: If this is only for the live-code-block component, why not move it to the styles of the component?

kulmann (approved):

PR #1136, opened 2025-09-01
refactor: gaps to tailwind

JammingBen:


packages/web-pkg/src/components/Filters/DateFilter.vue
```diff
@@ -335,10 +335,6 @@ export default defineComponent({
 <style lang="scss">
 .date-filter {
   &-list {
-    &-item {
-      gap: 8px;
-    }
```
JammingBen, line 340: Removed, is already set on a button.

packages/design-system/src/helpers/tailwind.ts
kulmann, line 1: we have `getTailwindSizeClass` in `design-system/src/helpers/sizeClasses.ts`. Can't we place the gap helper there as well instead of a new file? Or move the `getTailwindSizeClass` helper into this new `tailwind.ts` file. ;-)

packages/web-pkg/src/components/FilesList/ResourceTiles.vue
```diff
@@ -42,7 +42,7 @@
         </oc-filter-chip>
       </div>
     </div>
-    <oc-list class="oc-tiles grid justify-start">
+    <oc-list class="oc-tiles grid justify-start gap-3">
```
kulmann, line 45: you got me confused: here you have `gap-3` for 1rem, in other places you had `gap-2` for 1rem. What's correct now?

packages/design-system/src/helpers/tailwind.ts
JammingBen: I have a PR coming up that removes the `getTailwindSizeClass` helper. We must not use interpolation with Tailwind classes because Tailwind doesn't recognize these classes and they might be missing in the bundle therefore.

packages/web-pkg/src/components/FilesList/ResourceTiles.vue
JammingBen: `gap-3` is the 1:1 adaption for `1rem`. However, I found `gap-2` fitting for the other place (I think it was the tile grid in the app store).

JammingBen: Okay I checked again. Removed one of them (didn't do anything because elements are not side by side) and set the other to `gap-3` as well.

PR #1138, opened 2025-09-02
refactor: css mixins and custom-classes to tailwind

JammingBen:
This is not a necessary, but IMO nice-to-have adjustment. The mixins don't really comply with Tailwind spacings/sizes and do calculations with random numbers instead. So I figured it makes sense to migrate them to proper Tailwind classes. It's also easier to read and understand. The only exception is the button appearance and color mixin. I haven't decided whether we should keep it or not, but for now I think it's fine to have it.

Also removes some of the custom classes (like e.g. `oc-button-justify-content`) and replaces them with Tailwind classes. IMO this is more streamlined and easy to read, since we can (and should) handle almost all CSS within an element's `class` (or `:class`) tag.

Also removes the Tailwind class interpolation again. We must not interpolate Tailwind CSS classes because we would risk them not getting included in the final bundle.

tl;dr: This PRs aims to unify and streamline the styling of components in the design system by using Tailwind classes via the `class` attribute whenever possible. No more styling all over the place.


packages/design-system/src/components/OcIcon/OcIcon.vue
```diff
@@ -81,15 +87,15 @@ const nameWithFillType = computed(() => {
...
-
-const prefix = (string: string) => {
-  if (string !== null) {
-    return `oc-icon-${string}`
-  }
-}
+const tailwindSize = computed(() => ({
+  'size-3': size === 'xsmall',
+  'size-4': size === 'small',
+  'size-5.5': size === 'medium',
```
JammingBen, line 93: Keeping this the same as before results in an odd number. We might re-visit it later.

packages/web-app-files/src/components/AppBar/CreateAndUpload.vue
```diff
@@ -481,6 +481,16 @@ const folderIconResource = computed(() => {
...
+<style>
+@reference '@opencloud-eu/design-system/tailwind';
+
+@layer utilities {
+  .create-and-upload-actions .oc-drop .oc-icon svg {
+    /* reset the resource icon height because the ResourceIcon component messes with it */
+    @apply h-5.5;
+  }
+}
+</style>
```
JammingBen, line 493: This is super messed up. All because the `ResourceIcon` component reduces the file icons height to `70%`, which leads to the icons in this component to be distorted, because they sit in a container with `h-full`. Which is necessary because we want the icons to have some custom size here... We should probably clean that up at some point.

packages/design-system/src/components/OcButton/OcButton.vue
```diff
@@ -65,7 +79,7 @@ export interface Props {
    * @docs The alignment of the button content.
    * @default center
```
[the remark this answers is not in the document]
JammingBen: There is nothing breaking here.

packages/design-system/src/components/OcTable/OcTable.vue
```diff
@@ -351,6 +351,26 @@ const extractTableProps = () => {
...
+    case 'medium':
+      return side === 'right' ? 'pr-4' : 'pl-4'
+    case 'large':
+      return side === 'right' ? 'pr-6' : 'pl-6'
+    case 'xlarge':
+      return side === 'right' ? 'pr-12' : 'pl-12'
+    case 'xxlarge':
+      return side === 'right' ? 'pr-24' : 'pl-24'
+  }
+}
```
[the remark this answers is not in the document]
JammingBen: Not quite, it about the x-axis padding here.

packages/design-system/src/components/OcCheckbox/OcCheckbox.vue
```diff
@@ -5,8 +5,13 @@
       v-model="model"
       type="checkbox"
       name="checkbox"
-      :class="classes"
-      class="m-0.5 border-2 border-role-outline outline-0 focus-visible:outline outline-role-secondary"
+      class="oc-checkbox m-0.5 border-2 border-role-outline outline-0 focus-visible:outline outline-role-secondary rounded-sm checked:bg-white disabled:bg-role-surface-container-low indeterminate:bg-white bg-transparent inline-block overflow-hidden"
+      :class="{
+        'oc-checkbox-checked bg-white': isChecked,
```
kulmann, line 10: is it doing the same that we have `bg-white` when `isChecked` is true and `checked:bg-white` in the static class list?

kulmann (approved):

packages/design-system/src/components/OcCheckbox/OcCheckbox.vue
JammingBen: The last time I checked/tried, `isChecked` and the native `:checked` behaved a bit different for some reason, so I kept both 🤷 Didn't dive too deep into the implementation though.

Issue #1135, opened 2025-09-01
Add Feedback Form to the Web UI


tbsbdr:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

As a developer I want to know how OpenCloud is perceived by its users so that I know what should be improved.

### User Value

- Simple feedback possibility without login obligation

### Acceptance Criteria

- enable the feedback URL int he Web UI
- link to https://opencloud.eu/feedback-web
- https://opencloud.eu/feedback-web redirects to our limesurvey form cc @db-ot

tbsbdr:
================ Current
# OpenCloud Web Survey
_There are 3 questions in this survey._
This survey is anonymous.
The record of your survey responses does not contain any identifying information about you, unless a specific survey question explicitly asked for it.

If you used an identifying access code to access this survey, please rest assured that this code will not be stored together with your responses. It is managed in a separate database and will only be updated to indicate whether you did (or did not) complete this survey. There is no way of matching identification access codes with survey responses.

================ Proposal

# Share your Feedback

We’d love to hear your thoughts! This survey helps us understand what you enjoy about OpenCloud and what we can do better.

Your answers are anonymous. [Data Protection Notice](https://opencloud.eu/en/data-protection-notice)

tbsbdr:
@kulmann the redirect URL is set, could you enable and enter https://opencloud.eu/feedback-web as `feedbackLink` ?

kulmann:
@db-ot can we do something about the cut-off title (see screenshot)?

<img width="1624" height="1056" alt="Image" src="https://github.com/user-attachments/assets/1c783695-993b-4fec-aaa3-d7f812cb7ea1" />

PR #1145, opened 2025-09-03
feat: enable feedback link and fix button color

kulmann:
## Description

We have a feedback survey now. This PR enables the feedback button in the topbar and sets our survey url as default. Also fixes the color of the button.

## Related Issue

- Fixes https://github.com/opencloud-eu/web/issues/1135

Screenshot 2025-09-03 at 20 11 57

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/web-runtime/src/components/Topbar/FeedbackLink.vue
+++ b/packages/web-runtime/src/components/Topbar/FeedbackLink.vue
@@ -6,7 +6,7 @@
       :href="hrefOrFallback"
       target="_blank"
       appearance="raw-inverse"
-      color-role="surface"
+      color-role="chrome"
       :aria-label="ariaLabelOrFallback"
       aria-describedby="oc-feedback-link-description"
       no-hover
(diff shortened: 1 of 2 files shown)


tbsbdr:
How about keeping it shorter and a little bit more precise: `Share improvement ideas for OpenCloud`?
imo second best: "Feedback"

packages/web-runtime/src/components/Topbar/FeedbackLink.vue
```diff
@@ -41,17 +41,15 @@ export default defineComponent({
...
     ariaLabelOrFallback() {
       return this.ariaLabel || this.$gettext('OpenCloud feedback survey')
     },
     descriptionOrFallback() {
       return (
         this.description ||
-        this.$gettext(
-          "Provide your feedback: We'd like to improve the web design and would be happy to hear your feedback. Thank you! Your OpenCloud team."
-        )
+        this.$gettext("We'd be happy to hear your anonymous feedback about OpenCloud. Thank you!")
```
tbsbdr: my favorite would be "Share improvement ideas"

kulmann: `Share your feedback`? 🤔

kulmann: well, yes, improvement ideas is a little bit more direct... let's go with that.

PR #1148, opened 2025-09-04
refactor: migrate grid to tailwind

JammingBen:
Also simplifies the topbar grid a bit and makes it work nicely on all screen resolutions (this includes the search icon not being placed in the middle of the screen).

### DIFF
--- a/packages/web-app-files/src/components/FilesViewWrapper.vue
+++ b/packages/web-app-files/src/components/FilesViewWrapper.vue
@@ -29,16 +29,7 @@ export default defineComponent({
 </script>
 
 <style lang="scss" scoped>
-.files-view-wrapper {
-  grid-template-rows: max-content max-content 1fr;
-  grid-template-areas:
-    'header'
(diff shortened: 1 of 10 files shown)


packages/web-app-files/src/components/FilesViewWrapper.vue
```diff
@@ -29,16 +29,7 @@ export default defineComponent({
...
-.files-view-wrapper {
-  grid-template-rows: max-content max-content 1fr;
-  grid-template-areas:
-    'header'
-    'upload'
-    'main';
-}
-
 #files-view {
-  grid-area: main;
```
JammingBen, line 41: I don't see how/why we would need this..?

kulmann: Maybe just "not anymore"... I remember that Pascal introduced css grid into our code base, and that must be years ago.

kulmann (approved):

PR #1151, opened 2025-09-05
refactor: remaining css to tailwind #1

JammingBen:

#### Migration

- `oc-invisible-sr` > `sr-only`
- `oc-cursor-pointer` > `cursor-pointer`
- `oc-invisible` > `invisible`
- `oc-hidden` > `hidden`

For the `oc-visible@` classes, you generally want to set `hidden` on the element and then something like `sm:block` (or whatever `display` value fits). For `oc-hidden@` it's the same but vice versa.


packages/web-app-ocm/src/views/OutgoingInvitations.vue
```diff
@@ -40,7 +40,7 @@
                 !descriptionErrorMessage && `${formInput.description?.length || 0}/${50}`
               "
             />
-            <input type="submit" class="oc-hidden" />
+            <input type="submit" class="hidden" />
           </form> </template
```
kulmann: Could you break `</template>` into the next line here?

packages/web-pkg/src/components/AppTemplates/PartialViews/LoadingScreen.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div class="text-center flex justify-center items-center h-full">
     <oc-spinner size="xlarge" />
-    <p v-translate class="oc-invisible">Loading app</p>
+    <p v-translate class="invisible">Loading app</p>
```
kulmann: No `v-translate`, could you please fix this to `v-text="$gettext('Loading app')"` on the way? Also: why is it invisible? Shouldn't it be `sr-only` (I know, was wrong before as well)?

kulmann (approved):
💪

packages/web-pkg/src/components/AppTemplates/PartialViews/LoadingScreen.vue
JammingBen: Hmm yup, I guess you're right 👍

PR #1153, opened 2025-09-05
refactor!: remove scss files

JammingBen:


packages/design-system/docs/.vitepress/config.ts
```diff
@@ -53,26 +53,6 @@ export default defineConfig({
...
-            const split = src.split(stripScssMarker)
-            const newSrc = split[split.length - 1]
-
-            return {
-              code: newSrc,
-              map: null
-            }
-          }
-        }
-      },
```
dschmidt, line 75: I'm not sad to see this go... 😍 🎉

packages/design-system/src/index.ts
```diff
@@ -9,6 +9,7 @@ import * as directives from './directives'
 // fonts must be imported here to ensure they are included in the build
 import './styles/fonts.scss'
 import './styles/layers.css'
+import './../node_modules/tippy.js/dist/tippy.css'
```
dschmidt: Does this really need the node_modules in the path?

Issue #1147, opened 2025-09-04
update design system color roles


tbsbdr:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

collor roles at https://docs.opencloud.eu/design-system/designTokens/colorRoles.html are outdated

### Reproduction Steps

- look at https://docs.opencloud.eu/design-system/designTokens/colorRoles.html and compare to demo.opencloud.eu
- colors are not 100% the same

### Expected Outcome

- colors of demo.opencloud.eu and https://docs.opencloud.eu/design-system/designTokens/colorRoles.html should be aligned

### Actual Outcome

- colors are not aligned

PR #1161, opened 2025-09-09
chore!: remove token generation and unnecessary theming options

JammingBen:
See here for the full Tailwind migration guide: https://docs.opencloud.eu/design-system/gettingStarted/tailwindMigration.html

---

There is just no need for these things anymore, they just add unnecessary complexity.

### DIFF
--- a/packages/design-system/package.json
+++ b/packages/design-system/package.json
@@ -124,7 +124,6 @@
     "clean-publish": "^5.0.0",
     "glob": "^11.0.0",
     "markdown-it-container": "^4.0.0",
-    "onchange": "^7.1.0",
     "postcss": "^8.5.2",
     "process": "^0.11.10",
     "sass": "1.90.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -254,9 +254,6 @@ importers:
       markdown-it-container:
         specifier: ^4.0.0
         version: 4.0.0
-      onchange:
-        specifier: ^7.1.0
-        version: 7.1.0
       postcss:
         specifier: ^8.5.2
         version: 8.5.6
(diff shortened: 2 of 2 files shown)


packages/web-runtime/src/container/bootstrap.ts
```diff
@@ -362,15 +361,12 @@ export const announceTheme = async ({
...
-  const { currentTheme } = storeToRefs(themeStore)
 
   const webTheme = await loadTheme(configStore.theme)
 
-  await initializeThemes(webTheme)
+  initializeThemes(webTheme)
 
-  app.use(designSystem, {
-    tokens: currentTheme.value.designTokens
-  })
```
JammingBen, line 373: The theme store already applies tokens to `html`, no need for the design-system to do it again.

packages/web-runtime/src/container/bootstrap.ts
```diff
@@ -362,15 +361,12 @@ export const announceTheme = async ({
 }): Promise<void> => {
   const themeStore = useThemeStore()
   const { initializeThemes } = themeStore
-  const { currentTheme } = storeToRefs(themeStore)
 
   const webTheme = await loadTheme(configStore.theme)
 
-  await initializeThemes(webTheme)
+  initializeThemes(webTheme)
```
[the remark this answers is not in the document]
JammingBen: It's not.

packages/design-system/src/styles/defaults.css
kulmann, line 1: did you "sync" the color roles with the ones in the opencloud default theme?

JammingBen: Not yet, https://github.com/opencloud-eu/web/issues/1147 is still open.

PR #1168, opened 2025-09-10
fix: adjusted notification area with tailwind

tammi-23:
Refactor tailwind changes:

Related Issue: https://github.com/opencloud-eu/web/issues/937#issuecomment-3244280953
notification-Area

### DIFF
--- a/packages/web-runtime/src/components/Topbar/Notifications.vue
+++ b/packages/web-runtime/src/components/Topbar/Notifications.vue
@@ -41,11 +41,11 @@
           <li
             v-for="(el, index) in notifications"
             :key="index"
-            class="oc-notifications-item [&>a]:text-role-on-surface -z-1000 relative"
+            class="oc-notifications-item [&>a]:text-role-on-surface z-1000 relative"
           >
             <component
               :is="el.computedLink ? 'router-link' : 'div'"
-              class="flex items-center"
+              class="flex items-center gap-2"
               :to="el.computedLink"
             >
               <user-avatar


packages/web-app-files/src/components/Spaces/SpaceHeader.vue
```diff
@@ -75,8 +75,11 @@
       <div
         v-else-if="markdownResource && markdownContent"
         ref="markdownContainerRef"
-        class="markdown-container flex"
-        :class="{ 'mask-linear-[180deg,black,transparent] collapsed': markdownCollapsed }"
+        class="markdown-container flex min-h-0 [mask-image:none]"
+        :class="{
+          'collapsed [mask-image:linear-gradient(180deg,black,transparent)] collapsed':
```
JammingBen:
```suggestion
          'mask-image:linear-gradient(180deg,black,transparent)] collapsed':
```

packages/web-app-files/src/components/Spaces/SpaceHeader.vue
```diff
@@ -75,8 +75,10 @@
       <div
         v-else-if="markdownResource && markdownContent"
         ref="markdownContainerRef"
-        class="markdown-container flex"
-        :class="{ 'mask-linear-[180deg,black,transparent] collapsed': markdownCollapsed }"
+        class="markdown-container flex min-h-0 [mask-image:none]"
+        :class="{
+          'mask-image:linear-gradient(180deg,black,transparent)] collapsed': markdownCollapsed
+        }"
```
JammingBen: I just saw that this removes the gradient completely, which we want to keep. The intention with the `min-height` was to not make it as noticeable when the description is short. But it's probably just a bad workaround. Better adjust the gradient to be less prominent in general, e.g. via `mask-linear-[180deg,black,80%,transparent]`.

JammingBen (approved):
👍  (needs a rebase though)

PR #1172, opened 2025-09-10
Oc card component

kulmann:
## Description

Introduces an `OcCard` component and kills and hand-crafted usages of `oc-card` or `oc-login-card` classes.

## Related Issue

- Works towards https://github.com/opencloud-eu/web/issues/937

## Types of changes

- [ ] Bugfix


packages/design-system/src/components/OcCard/OcCard.vue
```diff
@@ -0,0 +1,128 @@
...
+    @apply bg-role-surface text-role-on-surface rounded-sm relative transition-shadow duration-100 ease-in-out;
+  }
+
+  .oc-card-header {
+    @apply p-4 pb-0 flex flex-col items-center relative;
+  }
+
+  .oc-card-body {
+    @apply p-4;
+    display: flow-root;
```
JammingBen:
```suggestion
    @apply p-4 flow-root;
```

packages/design-system/src/components/OcBottomDrawer/OcBottomDrawer.vue
```diff
@@ -16,33 +16,31 @@
...
-                  class="raw-hover-surface oc-bottom-drawer-close-button"
-                  :aria-label="$gettext('Close the context menu')"
-                  @click="hide"
-                >
-                  <oc-icon name="close" fill-type="line" />
-                </oc-button>
-              </div>
-            </div>
-            <div ref="bottomDrawerCardBodyRef" class="oc-card-body px-4 pb-4 pt-2">
+          <oc-card class="bg-transparent" header-class="flex flex-row justify-between items-center">
```
JammingBen, line 20: We should import `OcCard` since we're in the design-system here. It works without in Web, but only because Web installs the design-system via `vue.use()`. That should be a given in any case, but still...

packages/design-system/src/components/OcCard/OcCard.vue
```diff
@@ -0,0 +1,128 @@
...
+  return Object.hasOwn(slots, 'footer')
+})
+</script>
+
+<style>
+@reference "@opencloud-eu/design-system/tailwind";
+
+@layer components {
+  .oc-card {
+    @apply bg-role-surface text-role-on-surface rounded-sm relative transition-shadow duration-100 ease-in-out;
```
JammingBen: I know it was like that before, but why do we have a shadow transition here? 🤔 Can we maybe remove it? Cards don't have any shadows AFAIK.

kulmann: removed it

PR #1173, opened 2025-09-10
fix mobile flaky test

ScharfViktor:
image

Fix race condition when opening Collabora
Wait for load state, URL change, and response only after clicking Create button. Ensures Collabora is available before proceeding.


AlexAndBear (approved):
🤩

PR #1159, opened 2025-09-08
feat: set tiles view as default view option

AlexAndBear:
## Related Issue

- Fixes #1160 
## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix
- [ ] Enhancement (a change that doesn't break existing code or deployments)
- [ ] Breaking change (a modification that affects current functionality)
- [ ] Technical debt (addressing code that needs refactoring or improvements)
- [ ] Tests (adding or improving tests)
- [ ] Documentation (updates or additions to documentation)
- [ ] Maintenance (like dependency updates or tooling adjustments)

PR #1177, opened 2025-09-11
e2e. view mode switch

v-scharf:
waiting until https://github.com/opencloud-eu/web/pull/1159 will merged

### DIFF
--- a/packages/web-app-files/src/composables/extensions/useFolderViews.ts
+++ b/packages/web-app-files/src/composables/extensions/useFolderViews.ts
@@ -34,7 +34,7 @@ export const useFolderViews = (): FolderViewExtension[] => {
       ],
       folderView: {
         name: 'resource-table',
-        label: $gettext('Table view'),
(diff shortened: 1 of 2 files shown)


packages/web-app-files/src/composables/extensions/useFolderViews.ts
```diff
@@ -34,7 +34,7 @@ export const useFolderViews = (): FolderViewExtension[] => {
       ],
       folderView: {
         name: 'resource-table',
-        label: $gettext('Default table view'),
+        label: $gettext('Table view'),
```
JammingBen: I'd argue that the naming was fine (here and down below for the tiles view). We have a condensed table view, and we have a default table view. However, that does not mean that the default table view is actually the default view mode.

v-scharf: Ops, I'm confused here. Then I'll undo my changes.

PR #1203, opened 2025-09-16
fix: avatar alignment and z-index

JammingBen:
refs https://github.com/opencloud-eu/web/issues/937#issuecomment-3292272285


kulmann (approved):
👌

PR #1209, opened 2025-09-16
fix: line-height of resource name

JammingBen:
refs https://github.com/opencloud-eu/web/issues/937#issuecomment-3292272285


JammingBen:
@kulmann This fixes the vertical alignment of the resource names, especially in the top bar. It's still not 100% like before because the top bar height is now `40px` (= `h-10`) instead of `42px` (which would be `h-10.5`). To me it looks fine, what do you think?

The buttons still look cramped just because there are too many, especially with the auto save icon. But this is no regression from Tailwind. Still, do we want to do anything about that?

kulmann:
> The buttons still look cramped just because there are too many, especially with the auto save icon. But this is no regression from Tailwind. Still, do we want to do anything about that?

I think we discussed recently that we could/should get rid of the autosave-icon by using a different save-icon... but for that we'd need some sort of icon stacking I guess. Do you have other ideas?

JammingBen:
> I think we discussed recently that we could/should get rid of the autosave-icon by using a different save-icon... but for that we'd need some sort of icon stacking I guess. Do you have other ideas?

The only other idea I have is to display an overlay in some corner during the process of auto-saving. But that's not an always visible indicator. Then again: how does the user benefit from this information?

PR #1219, opened 2025-09-18
feat: remove edit space description inline button

kulmann:
## Description

Removing the edit space description inline-button from the space header. It's supposed to be pretty for everyone, not only for those who can't edit the description. ;-) Also, aligning the button vertically is virtually impossible because it depends on the vertical paddings of the rendered content - which is different if some headline or block text is used.

## Types of changes

- [x] Bugfix

### DIFF
--- a/packages/web-app-files/src/components/Spaces/SpaceHeader.vue
+++ b/packages/web-app-files/src/components/Spaces/SpaceHeader.vue
@@ -84,22 +84,10 @@
         }"
       >
         <text-editor
-          class="markdown-container-content w-full [&_#text-editor-preview-component]:!bg-transparent"
+          class="markdown-container-content w-full"
           is-read-only
           :current-content="markdownContent"
         />
-        <div v-if="isEditReadmeVisible" class="markdown-container-edit ml-2">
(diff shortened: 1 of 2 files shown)


packages/web-app-files/src/components/Spaces/SpaceHeader.vue
```diff
@@ -84,22 +84,10 @@
         }"
       >
         <text-editor
-          class="markdown-container-content w-full [&_#text-editor-preview-component]:!bg-transparent"
```
kulmann, line 87: `bg-transparent` is not needed here anymore, because I fixed that in den text-editor component styles in another PR.

tbsbdr:
> Please note: we also don't offer an image upload when clicking on the image or a title/subtitle edit for the space, why would we allow to edit the description from an inline-button then?

I woudl say its less about consistency. more important is, that the most frequent actions are "in your face" and the less frequent get a backseat. editing the space description turned out to be less frequently used at least from what I see out there.

tbsbdr (approved):
lgtm

PR #1233, opened 2025-09-22
fix: authenticated requests in password protected public links

JammingBen:
As authenticated user in a password protected public link context, the `makeRequest` method of the `useRequest` composable should use the password for authentication instead of the Bearer token.

refs https://github.com/opencloud-eu/opencloud/issues/1517

### DIFF
--- a/packages/web-pkg/src/composables/authContext/useRequest.ts
+++ b/packages/web-pkg/src/composables/authContext/useRequest.ts
@@ -1,7 +1,7 @@
 import { useClientService } from '../clientService'
 import type { Router, RouteLocationNormalizedLoaded } from 'vue-router'
 import type { Method, AxiosRequestConfig, AxiosResponse } from 'axios'
-import { ClientService } from '../../services'
+import { Auth, ClientService } from '../../services'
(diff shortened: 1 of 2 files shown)


packages/web-pkg/src/composables/authContext/useRequest.ts
```diff
@@ -24,25 +24,20 @@ export function useRequest(options: RequestOptions = {}): RequestResult {
...
-        config.headers.Authorization =
-          'Basic ' +
-          Buffer.from(['public', authStore.publicLinkPassword].join(':')).toString('base64')
-      }
-      if (authStore.publicLinkToken) {
-        config.headers['public-token'] = authStore.publicLinkToken
-      }
-    }
+    const httpClient =
+      authStore.accessToken && !authStore.publicLinkPassword
```
kulmann: Shouldn't we turn this check around and use `httpUnAuthenticated` if we're in any public link context (which is the case when a publicLinkToken is set, whether or not the link has a password doesn't matter, right?)

JammingBen: Discussed in chat, update pushed

Issue #1050, opened 2025-08-05
Load config.json authenticated


kulmann:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

> * As an admin, I want to protect some config.json content from being visible unauthenticated so that I can protect my environment.
> * As an admin, I want to be able to serve app config user-specific, based on rules, so that certain config is not available to all users.

### User Value

Security

### Acceptance Criteria

- apps don't get registered and initialized until the config.json has been loaded with an authenticated request => requires to postpone app registration and initialization until after the user access token is available and the config.json has been loaded a second time with authentication.

PR #1249, opened 2025-09-23
feat: load config authenticated

JammingBen:
Loads the config a second time authenticated after the user has been loaded.

The acceptance criteria from the story can not be fulfilled because we need routes to load the auth context, and we need apps for the routes. Hence it's not possible to load the apps after the authenticated config. However, this doesn't seem to be needed because the `extensions` prop of an application is defined as a `ref` and is therefore reactive. That means checks on a user-specific config can still be made.

Also removes deprecated app properties.

closes https://github.com/opencloud-eu/web/issues/1050

### DIFF
--- a/packages/web-runtime/src/container/bootstrap.ts
+++ b/packages/web-runtime/src/container/bootstrap.ts
@@ -138,12 +138,16 @@ const getEmbedConfigFromQuery = (
  */
 export const announceConfiguration = async ({
   path,
-  configStore
+  configStore,
+  token
 }: {
   path: string
   configStore: ConfigStore
+  token?: string
 }) => {
-  const request = await fetch(path, { headers: { 'X-Request-ID': uuidV4() } })
+  const request = await fetch(path, {
+    headers: { 'X-Request-ID': uuidV4(), ...(token && { Authorization: `Bearer ${token}` }) }
+  })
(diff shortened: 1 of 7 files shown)


packages/web-runtime/src/container/bootstrap.ts
```diff
@@ -138,12 +138,16 @@ const getEmbedConfigFromQuery = (
...
+  configStore,
+  token
 }: {
   path: string
   configStore: ConfigStore
+  token?: string
 }) => {
-  const request = await fetch(path, { headers: { 'X-Request-ID': uuidV4() } })
+  const request = await fetch(path, {
+    headers: { 'X-Request-ID': uuidV4(), ...(token && { Authorization: `Bearer ${token}` }) }
```
kulmann, line 149: Just thinking, what about public links? Should we authenticate with public link auth as well?

JammingBen: Hmm do you think it's necessary? Isn't public basically always anonymous and therefore non-user-specific?

kulmann: True. I guess we can start without and if there'll be good reasons from someone, then we think about it again.

kulmann (approved):
💪

PR #1264, opened 2025-09-25
feat: add update check

### LINKED ISSUE #1158: Inform Users about new OpenCloud version
### User Story

> As an admin I want to know if there is a new Version so that the server stays secure.

### User Value

- ensure a secure server
- provide new features for users

### Acceptance Criteria

- The web-ui version doesn't get shown anymore in the sidebar (we can keep it in the browser console)
- The web-ui version doesn't get shown anymore in the admin settings `General` view
- If a new server version is available, the version number at the bottom left shows Version `X.X.X available `
- If the server version is up to date, the version number at the bottom left shows `Up to date ✓` 
- for both: on click open link to well formated release notes ( comes from the server)
- during request runtime show `Checking for updates` must not block anything else.
- depending on whether opencloud is a stable or rolling release, the version information gets requested from:
  - <update server url>/server/rolling.json
  - <update server url>/server/stable.json
  - content is:
     ```json
      {
          "current_version": "1.2.3"
      }
     ```
  - web announces the version of the requesting instance via `version=a.b.c` query param in the update server url
  - web announces the server url of the requesting instance via `server=https://demo.opencloud.eu` query param in the update server url
  - web announces the server edition of the requesting instance via `edition=rolling` (or `edition=production`) query param in the update server url

### Implementation Detail

- Information are refreshed from pageload - not polling 

### Visuals

Image

Image

AlexAndBear:
## Description
image

## Related Issue

- Fixes #1158 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/web-runtime/src/components/VersionCheck.vue
+++ b/packages/web-runtime/src/components/VersionCheck.vue
@@ -83,7 +83,8 @@ const loadVersionsTask = useTask(function* (signal) {
           server: configStore.serverUrl,
           edition: serverEdition,
           version: currentServerVersion
-        }
+        },
+        signal
       }
     )
 
--- a/packages/web-runtime/src/components/VersionCheck.vue
+++ b/packages/web-runtime/src/components/VersionCheck.vue
@@ -73,8 +73,6 @@ const loadVersionsTask = useTask(function* (signal) {
       }
     )
 
-    // TODO: implement lts comparison
-
     const newestVersion = data.channels[serverEdition].current_version
     if (semver.gt(newestVersion, currentServerVersionSanitized)) {
       updateAvailable.value = true
--- a/packages/web-runtime/src/components/VersionCheck.vue
+++ b/packages/web-runtime/src/components/VersionCheck.vue
@@ -64,7 +64,14 @@ const isLoading = computed(() => loadVersionsTask.isRunning || !loadVersionsTask
 const loadVersionsTask = useTask(function* (signal) {
   try {
     const { data }: { data: UpdateResponseData } = yield httpUnAuthenticated.get(
-      `https://update.opencloud. eu/server.json?version=${currentServerVersion}&server=${configStore.serverUrl}`
+      `https://update.opencloud.eu/server.json`,
+      {
+        params: {
+          server: configStore.serverUrl,
+          edition: serverEdition,
+          version: currentServerVersion
+        }
+      }
     )
 
     // TODO: implement lts comparison
--- a/packages/web-runtime/src/components/VersionCheck.vue
+++ b/packages/web-runtime/src/components/VersionCheck.vue
@@ -1,0 +1,85 @@
+<template>
+  <div v-if="isLoading" class="flex items-center">
+    <span v-text="$gettext('Checking for updates')" />
+    <oc-spinner class="ml-1" size="xsmall" />
+  </div>
+  <div v-else>
+    <div v-if="!updateAvailable" class="flex items-center">
+      <span v-text="$gettext('Up to date')" />
+      <oc-icon class="ml-0.5" name="checkbox-circle" size="xsmall" fill-type="line" />
+    </div>
+    <div v-else>
+      <oc-button
+        class="text-role-on-surface-variant"
+        size="small"
+        type="router-link"
+        :href="updateData.url"
+        target="_blank"
+        gap-size="small"
+        appearance="raw"
+      >
+        <span
+          class="text-xs"
+          v-text="$gettext('Version %{version} available', { version: updateData.current_version })"
+        />
+        <oc-icon name="refresh" size="xsmall" fill-type="line" />
+      </oc-button>
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+import { onMounted, computed, ref } from 'vue'
+import { useTask } from 'vue-concurrency'
+import { useGettext } from 'vue3-gettext'
+import semver from 'semver'
+import { useCapabilityStore, useClientService, useConfigStore } from '@opencloud-eu/web-pkg/src'
+
+export interface UpdateChannelData {
+  current_version: string
+  url: string
+}
+
+export type UpdateChannelName = 'rolling' | 'stable' | `lts${string}`
+
+export type UpdateChannels = Record<UpdateChannelName, UpdateChannelData>
+
+export interface UpdateResponseData {
+  channels: UpdateChannels
+}
+
+const { $gettext } = useGettext()
+const { httpUnAuthenticated } = useClientService()
+const capabilityStore = useCapabilityStore()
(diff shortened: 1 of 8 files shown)


packages/web-runtime/src/components/VersionCheck.vue
```diff
@@ -0,0 +1,92 @@
...
+      {
+        params: {
+          server: configStore.serverUrl,
+          edition: serverEdition,
+          version: currentServerVersion
+        }
+      }
+    )
+
+    // TODO: implement lts comparison
```
kulmann: I think you can remove the TODO. That's not in scope for this PR and when we need it we'll find the code location.

packages/web-runtime/src/components/VersionCheck.vue
```diff
@@ -0,0 +1,89 @@
...
+const configStore = useConfigStore()
+
+const updateAvailable = ref(false)
+const updateData = ref<UpdateChannelData>()
+//TODO: retrieve serverEdition
+const serverEdition = 'rolling'
+const currentServerVersion = capabilityStore.status.productversion
+const currentServerVersionSanitized = currentServerVersion.split('+')[0]
+
+const isLoading = computed(() => loadVersionsTask.isRunning || !loadVersionsTask.last)
```
kulmann: can we have a debounce of 1 second or so, so that the `Checking for updates` string is visible in the beginning? Otherwise it just flickers, and then I'm annoyed because it flickers too fast an I can't read the text.

packages/web-runtime/src/components/VersionCheck.vue
```diff
@@ -0,0 +1,89 @@
+<template>
+  <div v-if="isLoading" class="version-check-loading flex items-center">
+    <span v-text="$gettext('Checking for updates')" />
+    <oc-spinner class="ml-1" size="xsmall" />
+  </div>
+  <div v-else>
+    <div v-if="!updateAvailable" class="version-check-no-updates flex items-center">
```
kulmann: We also need an error case here. If the request fails (e.g. because of a missing `connect-src` entry in the CSP rules), we can't show `Up to date`. We should hint towards an error requesting the most recent version. `Version check failed` or something similar (short!) would be fine.

packages/web-runtime/src/components/VersionCheck.vue
```diff
@@ -0,0 +1,89 @@
...
+import { useGettext } from 'vue3-gettext'
+import semver from 'semver'
+import { useCapabilityStore, useClientService, useConfigStore } from '@opencloud-eu/web-pkg/src'
+
+export interface UpdateChannelData {
+  current_version: string
+  url: string
+}
+
+export type UpdateChannelName = 'rolling' | 'stable' | `lts${string}`
```
kulmann: Could you please remove the `lts` type here and only keep `rolling` and `stable` for now? We can introduce lts types when we have an LTS version.

kulmann (changes requested):

packages/web-runtime/src/components/VersionCheck.vue
```diff
@@ -4,7 +4,11 @@
     <oc-spinner class="ml-1" size="xsmall" />
   </div>
   <div v-else>
-    <div v-if="!updateAvailable" class="version-check-no-updates flex items-center">
+    <div v-if="hasError" class="version-check-error flex items-center">
```
kulmann, line 7: Why is this wrapped in another div? Can't this be a `<div v-else-if="hasError" ...` one level higher?

packages/web-runtime/src/components/VersionCheck.vue
```diff
@@ -52,14 +57,19 @@ const { httpUnAuthenticated } = useClientService()
...
 const updateAvailable = ref(false)
+const hasError = ref(false)
 const updateData = ref<UpdateChannelData>()
 //TODO: retrieve serverEdition
 const serverEdition = 'rolling'
 const currentServerVersion = capabilityStore.status.productversion
 const currentServerVersionSanitized = currentServerVersion.split('+')[0]
 
-const isLoading = computed(() => loadVersionsTask.isRunning || !loadVersionsTask.last)
+const stopLoading = debounce(() => {
```
kulmann: huh? `isLoading` should be the same as before, just additionally also check that it ran at least 1 second. Now, if the request runs longer than 1 second, we'll be in a broken state.

kulmann (changes requested):

packages/web-runtime/src/components/VersionCheck.vue
AlexAndBear: Because it's much easier to read with all the other if conditions following. I will cange the div to template

packages/web-runtime/src/components/VersionCheck.vue
AlexAndBear: why should it be in a broken state ? it would add 1 more second.

packages/web-runtime/src/components/VersionCheck.vue
```diff
@@ -0,0 +1,104 @@
...
+
+  try {
+    const { data }: { data: UpdateResponseData } = yield httpUnAuthenticated.get(
+      `https://update.opencloud.eu/server.json`,
+      {
+        params: {
+          server: configStore.serverUrl,
+          edition: serverEdition,
+          version: currentServerVersion
+        }
```
JammingBen:
```suggestion
        },
        signal
```

packages/web-runtime/src/components/VersionCheck.vue
```diff
@@ -0,0 +1,104 @@
...
+    </oc-button>
+  </template>
+</template>
+
+<script setup lang="ts">
+import { computed, onMounted, ref } from 'vue'
+import { useTask } from 'vue-concurrency'
+import { useGettext } from 'vue3-gettext'
+import semver from 'semver'
+import { useCapabilityStore, useClientService, useConfigStore } from '@opencloud-eu/web-pkg/src'
```
JammingBen:
```suggestion
import { useCapabilityStore, useClientService, useConfigStore } from '@opencloud-eu/web-pkg'
```

packages/web-runtime/src/components/VersionCheck.vue
```diff
@@ -0,0 +1,104 @@
+<template>
+  <div v-if="isLoading" class="version-check-loading flex items-center">
+    <span v-text="$gettext('Checking for updates')" />
+    <oc-spinner class="ml-1" size="xsmall" />
+  </div>
+  <template v-else>
```
kulmann, line 6: Even with the `template` tag I don't think that it improves readability. :-P But fine by me to keep it like it is.

kulmann (approved):
Love it! 🥇

PR #1268, opened 2025-09-26
refactor: add own password generator

JammingBen:
Removes the `js-generate-password` lib in favor of our own implementation because it uses `Math.random`, which is not cryptographically secure.

See https://blog.hboeck.de/archives/907-How-to-create-a-Secure,-Random-Password-with-JavaScript.html for more detail.


packages/web-pkg/src/services/passwordPolicy/passwordPolicy.ts
```diff
@@ -130,13 +129,59 @@ export class PasswordPolicyService {
   }
 
   public generatePassword(): string {
```
kulmann, line 131: Could you add unit tests that this really generates some password, please? 😅  Doesn't need to be fancy.. I'd be happy with "this generates something that's not an empty string" already. Anything more useful is appreciated as well of course.

Issue #1280, opened 2025-09-29
Editing shares re-loads versions and tags


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

Editing shares via the right sidebar re-loads the versions and tags, which is unnecessary.

### Reproduction Steps

- open the right sidebar for a file
- open the network tab of your browser
- add a share
- check the latest requests

### Expected Outcome

No loading of versions and tags.

### Actual Outcome

Versions and tags get loaded.

This might be a bit tricky, I guess each sidebar panel gets re-rendered, causing versions and tags to get loaded on mount.

PR #1285, opened 2025-09-30
fix: sidebar versions loading (performance)

JammingBen:
Previously, when the current resource had changed, the sidebar would re-fetch all versions, causing the entire sidebar to get re-rendered.

I improved this in 2 ways:

* only load versions if the mdate of the current resource has changed.
* (re-)loading versions now only (re-)renders the file details list, no the entire sidebar.

fixes https://github.com/opencloud-eu/web/issues/1280

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
+++ b/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
@@ -27,7 +27,11 @@
         <oc-status-indicators :resource="resource" :indicators="shareIndicators" />
         <p class="my-0 mx-2" v-text="detailSharingInformation" />
       </div>
+      <div v-if="detailsLoading" class="flex justify-center">
+        <oc-spinner :aria-label="$gettext('Loading details')" />
+      </div>
       <dl
+        v-else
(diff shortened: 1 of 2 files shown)


packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
```diff
@@ -27,7 +27,11 @@
         <oc-status-indicators :resource="resource" :indicators="shareIndicators" />
         <p class="my-0 mx-2" v-text="detailSharingInformation" />
       </div>
+      <div v-if="versionsLoading" class="flex justify-center">
+        <oc-spinner :aria-label="$gettext('Loading details')" />
```
kulmann, line 31: Why not `Loading versions`? 🤔

JammingBen: I thought just "Loading versions" might raise the question why none of the details are showing. The loading spinner hides all details, although (currently) only depending on the version loading.

kulmann: ok, makes sense :-) maybe rename the computed to `areDetailsLoading` then... but fine either way.

Issue #889, opened 2025-07-02
A11y list of issues


ScharfViktor:
## Description:

This ticket originates from automated accessibility (A11Y) testing. The report highlights multiple issues that need to be addressed, such as missing or incorrect labels, invalid attributes, and other markup problems. Each identified issue must be reviewed and corrected to ensure compliance with accessibility standards.

### Acceptance Criteria:

- Review the full list of reported accessibility issues
- Verify each item individually and confirm whether it is valid
- Fix all confirmed accessibility violations (e.g., missing labels, incorrect attributes)
- Rerun the automated accessibility tests to ensure no violations remain with viktor

### Reporting

```
♿ Accessibility violations detected in before clicking login submit:

[autocomplete-valid] autocomplete attribute must be used correctly
  Impact: serious
  Description: Ensure the autocomplete attribute is correct and suitable for the form field
  Help: https://dequeuniversity.com/rules/axe/4.10/autocomplete-valid?application=playwright
  Node 1: <input class="oc-input " autocapitalize="off" spellcheck="false" autocomplete="kopano-account username" placeholder="Username" id="oc-login-username" aria-invalid="false" value="Alice-a96">
  Node 2: <input class="oc-input " type="password" margin="normal" autocomplete="kopano-account current-password" placeholder="Password" id="oc-login-password" aria-invalid="false">
Accessibility check failed with 1 violation(s) in before clicking login submit.
    ✔ When "Alice" logs in                                                                                                                          # file:/Users/v.scharf/Work/opencloud/web/tests/e2e/cucumber/steps/ui/session.ts:32
♿ Accessibility violations detected in before clicking login submit:

[autocomplete-valid] autocomplete attribute must be used correctly
  Impact: serious
  Description: Ensure the autocomplete attribute is correct and suitable for the form field
  Help: https://dequeuniversity.com/rules/axe/4.10/autocomplete-valid?application=playwright
  Node 1: <input class="oc-input " autocapitalize="off" spellcheck="false" autocomplete="kopano-account username" placeholder="Username" id="oc-login-username" aria-invalid="false" value="Brian-a96">
  Node 2: <input class="oc-input " type="password" margin="normal" autocomplete="kopano-account current-password" placeholder="Password" id="oc-login-password" aria-invalid="false">
Accessibility check failed with 1 violation(s) in before clicking login submit.
[...]

PR #1296, opened 2025-10-01
fix: resolve axe violations (button-name, empty-table-header, spinner)

tammi-23:
Fixed A11y issues mentioned in https://github.com/opencloud-eu/web/issues/889


packages/design-system/src/components/OcAvatars/OcAvatars.vue
```diff
@@ -150,6 +150,14 @@ const otherItems = computed(() => {
   return a.slice(0, maxDisplayed! - avatars.value.length)
 })
 
+const avatarsAriaLabel = computed(() => {
+  if (isTooltipDisplayed) return tooltip.value as string
+  const names = (avatars.value || [])
+    .map((a: any) => a?.displayName || a?.name || a?.username)
```
JammingBen: Is `any` necessary? Could you remove it, or find the correct type?

```suggestion
    .map((a) => a?.displayName || a?.name || a?.username)
```

packages/design-system/src/components/OcButton/OcButton.vue
```diff
@@ -2,7 +2,7 @@
   <component
     :is="type"
     v-bind="additionalAttributes"
-    :aria-label="ariaLabel"
+    :aria-label="ariaLabel || undefined"
```
JammingBen: Why is this necessary? `ariaLabel` defaults to `undefined` when not provided, no? 🤔

packages/web-pkg/src/components/ContextActions/ActionMenuItem.vue
```diff
@@ -5,7 +5,13 @@
       :type="componentType"
       v-bind="componentProps"
       :class="[action.class, 'action-menu-item', 'align-middle', 'w-full', ...buttonClasses]"
-      :aria-label="componentProps.disabled ? action.disabledTooltip?.(actionOptions) : ''"
+      :aria-label="
+        componentProps.disabled
+          ? action.disabledTooltip?.(actionOptions)
+          : typeof action.label === 'function'
+            ? action.label(actionOptions)
+            : action.label
```
JammingBen: `label` is typed as a function, it should work without the condition?

packages/web-pkg/src/components/FilesList/ResourceLink.vue
```diff
@@ -90,6 +91,15 @@ export default {
...
       }
+    },
+    mergedProps(): Record<string, unknown> {
+      const {
+        target: _t,
+        draggable: _d,
+        'no-hover': _nh,
+        ...restAttrs
+      } = this.$attrs as Record<string, unknown>
+      return { ...restAttrs, ...this.componentProps }
```
JammingBen: Hmm this feels a bit weird 😅 I'd suggest removing `componentProps`, `mergedProps`, `inheritAttrs` and `v-bind` all together. The props from `componentProps` can be set on the element directly. Then you can simply set things like `aria-label` from the outside (like you already do 👍 ).

packages/web-pkg/src/components/FilesList/ResourceListItem.vue
```diff
@@ -10,6 +10,7 @@
       :is-resource-clickable="isResourceClickable"
       class="contents relative"
       :class="{ 'hover:underline': isResourceClickable }"
+      :aria-label="$gettext('Open: %{name}', { name: resource?.name ?? '' })"
```
JammingBen:
```suggestion
      :aria-label="$gettext('Open »%{name}«', { name: resource?.name ?? '' })"
```

packages/design-system/src/components/OcButton/OcButton.vue
tammi-23: I’ve updated the definition of ariaLabel for spinners, so this part can be changed back.

packages/web-pkg/src/components/FilesList/ResourceLink.vue
JammingBen, line 1: I think you can now remove the computed properties `componentType` and `componentProps`.

JammingBen (approved):
Nice 👍 

@ScharfViktor Can we adjusts the a11y CI now so it fails when there are violations?

Issue #1181, opened 2025-09-11
Removing a user from a group lists group the user is not member of


schweigisito:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

Once a user gets removed from a group the dropdown list displays all groups which exist on the system.

E.g. the test user "Dennis" is member of

```
    "mail": "dennis@example.org",
    "memberOf": [
        {
            "displayName": "users",
            "groupTypes": [],
            "id": "509a9dcd-bb37-4f4f-a01a-19dca27d9cfa"
        },
        {
            "displayName": "unix-lovers",
            "groupTypes": [],
            "id": "75bc3882-dd94-11ef-ad60-335f3df6cef3"
        },
        {
            "displayName": "basic-haters",
            "groupTypes": [],
            "id": "a4eb2c12-dd94-11ef-9ebe-eb96f938d517"
        },
        {
            "displayName": "programmers",
            "groupTypes": [],
            "id": "ce4aa240-dd94-11ef-82b8-4f4828849072"
        }
    ],
```

but the dropdown list shows also e.g. "visi-lovers".

### Reproduction Steps

1. List users in Admin module
2. Select a user and try to remove him from a group
3. Check the drop down list

=> It contains all groups

### Expected Outcome

Only lists groups the user is member of.

### Actual Outcome

All groups are listed while it could/should only show those the user is member of.

Removing a user from a group where it is not member of throws a notification saying the membership had already been removed.

PR #1298, opened 2025-10-01
fix: user group assignment options, details panel alignment

JammingBen:
Only show relevant groups in user-group assignment. For removing groups, only show groups the user(s) are assigned to. For adding groups for one user, only show groups the uses is not already assigned to. For multiple users, we just show all groups because computing the intersection feels a bit overkill.

Also fixes the alignment in the user details panel and refactors some components to script setup.

fixes https://github.com/opencloud-eu/web/issues/1181

### DIFF
--- a/packages/web-app-admin-settings/tests/unit/components/Users/RemoveFromGroupsModal.spec.ts
+++ b/packages/web-app-admin-settings/tests/unit/components/Users/RemoveFromGroupsModal.spec.ts
@@ -4,11 +4,23 @@ import { mock } from 'vitest-mock-extended'
 import { Group, User } from '@opencloud-eu/web-client/graph/generated'
 import { Modal, useMessages } from '@opencloud-eu/web-pkg'
 import { useUserSettingsStore } from '../../../../src/composables/stores/userSettings'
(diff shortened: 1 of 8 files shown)


packages/web-app-admin-settings/tests/unit/components/Users/RemoveFromGroupsModal.spec.ts
```diff
@@ -4,13 +4,25 @@ import { mock } from 'vitest-mock-extended'
...
+import GroupSelect from '../../../../src/components/Users/GroupSelect.vue'
 
 describe('RemoveFromGroupsModal', () => {
   it('renders the input', () => {
     const { wrapper } = getWrapper()
     expect(wrapper.find('group-select-stub').exists()).toBeTruthy()
   })
 
+  describe('available groups', () => {
+    it('lists only groups the user is assigned to', () => {
```
kulmann, line 16: Can we also have a second unit test with multiple users where users are assigned to different groups (partially overlapping, partially different)? The result should be, that any group is shown which is assigned to at least one of the users, right?

Issue #615, opened 2025-04-30
Enable strict type checks


JammingBen:
We want to enable strict type checks in Web for a better (and safer) development experience. That not only affects Web itself, but also all other apps using our TypeScript config.

I just tried it by simply setting `strict: true`, that gives about 3000 errors. Almost all of them come from `strictNullChecks`, which is part of the strict mode. So I'd suggest enabling the strict mode while keeping `strictNullChecks: false` for now. That brings it down so a manageable number of 303 failed type checks.

In addition to that, I suggest enabling the eslint rule `vue/require-typed-ref` to enforce properly typed refs.

PR #1307, opened 2025-10-02
refactor: require typed refs

JammingBen:
Make typing `ref`s required via the `vue/require-typed-ref` eslint plugin.

Also removes some unnecessary `ref`s (mainly Mark instances) and adds template refs via `useTemplateRef`. Note that HTML native elements don't need to be typed explicitly with `useTemplateRef` when using script setup. Because of that, I also refactored 2 components to script setup.

refs https://github.com/opencloud-eu/web/issues/615


AlexAndBear (approved):
👍

PR #1319, opened 2025-10-06
fail accessibility tests on any violation

ScharfViktor:
Now that we fixed all a11y issues (see #889), I switched the test to fail if new a11y issues appear.


JammingBen (approved):
Thank you!

PR #1323, opened 2025-10-06
feat: add update check for mobile view

### LINKED ISSUE #1317: Update check on mobile viewport
### Initial Checklist

### Bug Description

Mobile viewport sizes don't perform the update check.

### Reproduction Steps

1. Open the web ui with a mobile viewport size
2. Check the network tab
3. Look for the update server request

### Expected Outcome

Update server request is being made

### Actual Outcome

Update server request ist **not** being made

AlexAndBear:
## Description

image

## Related Issue

- Fixes #1317 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix


AlexAndBear:
@kulmann in order to have a clean loading and error state, instead of using the config store we might want to implement an updateStore (updates, isLoading, error...). 

We can also make the request non blocking in the bootstrap process, I think.

packages/web-pkg/src/composables/piniaStores/config/config.ts
```diff
@@ -96,6 +98,10 @@ export const useConfigStore = defineStore('config', () => {
     }
   }
 
+  const loadConfigUpdates = (data: RawConfig['updates']) => {
```
kulmann: Since this is not loading anything I'd choose a setter style name (`setAvailableUpdates`, `setUpdates`, ... ?)

kulmann (commented):
I like this approach 👍

packages/web-pkg/src/composables/piniaStores/config/config.ts
AlexAndBear: I just used the pre existing name for setting the config "loadConfig" for consistency (this method doesn't load anything as well). 

I left another comment, asking if it would make sense to have a new updateStore, so we can be more free with naming and implement an error and loading state, what's your take on that ?

packages/web-runtime/src/container/bootstrap.ts
```diff
@@ -676,6 +681,51 @@ export const announceVersions = ({
...
+  capabilityStore: CapabilityStore
+  configStore: ConfigStore
+  clientService: ClientService
+}): Promise<void> => {
+  if (!capabilityStore.capabilities.core['check-for-updates']) {
+    return
+  }
+
+  try {
+    updatesStore.setIsLoading(true)
```
kulmann, line 708: Why do you set loading to `true` here? It's initialized with `true`.

packages/web-runtime/src/container/bootstrap.ts
```diff
@@ -676,6 +681,51 @@ export const announceVersions = ({
...
+    const { data }: { data: Updates } = await clientService.httpUnAuthenticated.get(
+      'https://update.opencloud.eu/server.json',
+      {
+        params: {
+          server: configStore.serverUrl,
+          edition: 'rolling', //TODO: retrieve serverEdition
```
kulmann: Could you already set the edition from the config here and do a fallback to `rolling`? Then it would automatically be used when the backend starts of include the edition.

Might be necessary to validate for a limited set of values (rolling or stable).

packages/web-runtime/src/container/bootstrap.ts
AlexAndBear: just in case, if we re-use this function, e.G polling or something else, this would come in handy

packages/web-runtime/src/container/bootstrap.ts
AlexAndBear: if you refer to capabilities -> yes

kulmann (approved):
nice 🚀  thank you!

PR #1334, opened 2025-10-08
refactor: remove old archiver handling and semver

JammingBen:
We only support archivers with file id support by now, hence removes the old archiver handling.

Also removes the `semver` dependency since we only use it for comparing versions, which can be done via a simple helper method. No need for a full-blown library that lacks proper type support.

### DIFF
--- a/packages/web-pkg/src/utils/semver.ts
+++ b/packages/web-pkg/src/utils/semver.ts
@@ -1,0 +1,41 @@
+export const parseVersion = (tag: string) => {
+  const version = tag.startsWith('v') ? tag.slice(1) : tag
+  const [main, pre] = version.split('-')
+  const [major, minor, patch] = main.split('.').map(Number)
+  return { major, minor, patch, pre }
+}
+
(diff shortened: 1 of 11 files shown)


packages/web-pkg/src/utils/semver.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+/**
+ * Compares two version strings.
+ * Returns:
+ * - negative number if a is less than b,
+ * - positive number if a is greater than b,
+ * - 0 if a and b are equal.
+ */
+export const compareVersions = (a: string, b: string) => {
```
kulmann, line 15: Do you think it would also make sense to create `lt` (less than) and `gt` (greater than) helpers, like they existed in the semver package?

JammingBen: At the moment there is only one place where we would need that, so I'd say we can add it later if we need it somewhere else. Or what do you think?

Issue #1358, opened 2025-10-10
Don't always reload entire folder content after upload


JammingBen:
After uploading a file/folder, we currently reload the entire folder (given that the user hasn't left the folder or space meanwhile). We can't always prevent this, but if only a few files/folders have been uploaded, we could just fetch those. This should alleviate some load on the server and improve UX slightly since there wouldn't be any flickering due to the loading indicator.

PR #1361, opened 2025-10-10
feat: add single resources to store after uploading

JammingBen:
Instead of reloading the entire file list. This results in an improved UX and performance. Also, fixes the highlighting of new resources in the files list.

closes https://github.com/opencloud-eu/web/issues/1358
fixes https://github.com/opencloud-eu/web/issues/1364

### DIFF
--- a/packages/web-pkg/src/services/uppy/uppyService.ts
+++ b/packages/web-pkg/src/services/uppy/uppyService.ts
@@ -263,9 +263,9 @@ export class UppyService {
         return
       }
       this.publish('uploadCompleted', result)
-      result.successful.forEach((file) => {
-        this.uppy.removeFile(file.id)
-      })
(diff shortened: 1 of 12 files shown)


packages/web-pkg/src/services/uppy/uppyService.ts
```diff
@@ -263,9 +263,9 @@ export class UppyService {
         return
       }
       this.publish('uploadCompleted', result)
-      result.successful.forEach((file) => {
-        this.uppy.removeFile(file.id)
-      })
+      for (let i = 0; i < result.successful.length; i++) {
+        this.uppy.removeFile(result.successful[i].id)
+      }
```
JammingBen, line 268: For performance reasons, there may be several thousand files.

kulmann (approved):
nice 💪

PR #1403, opened 2025-10-21
perf: increase performance for folder and multiple file downloads

JammingBen:
Instead of manually creating a blob which can be downloaded by the browser, we now pass the signed URL directly to the browser which gives it a huge performance boost, especially with large downloads. Also saves up one unnecessary `HEAD` request when downloading single files.

In addition to that:

* deprecates `isUrlSigningEnabled` since it's always supported by the server
* deprecates `signUrlTimeout` since it has now effect anyways


kulmann (approved):
Ooof size is large 😅 Thanks for fixing this!

Issue #1392, opened 2025-10-17
Put view options into mobile menu earlier if the right sidebar is open


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

The view options mobile menu should display one breakpoint earlier than 640px (= 960px) if the right sidebar is open to avoid them being cramped.

### Reproduction Steps

- go to your personal space
- open the right sidebar
- reduce the screen size to about 800px

### Expected Outcome

The view options should go into the mobile menu to save up space.

### Actual Outcome

The view options split into multiple lines, everything looks cramped and bad.

Maybe even put the breadcrumbs in the mobile menu... The screen space is just so limited here.

<img width="819" height="226" alt="Image" src="https://github.com/user-attachments/assets/cf4dabf8-9eff-4a87-bef1-e91a0deae14b" />

PR #1419, opened 2025-10-23
fix: view options and breadcrumbs on mobile with right sidebar 

JammingBen:
Switch to the mobile versions of the view options and the breadcrumbs one breakpoint earlier if the right sidebar is open.

fixes https://github.com/opencloud-eu/web/issues/1392


AlexAndBear (approved):
👍

PR #1427, opened 2025-10-24
fix: only shorten links in markdown editor if character count is long…

### LINKED ISSUE #1410: Pasting URLs always causes them to be shortened.
## Describe the bug
Trying to insert a URL into a Markdown or Plain Text file. The URL is automatically shortened so that it cannot be read. This happens also when I use "paste as plain text" option.
If I type the URL by hand then after a short pause it gets shortened but the link can be removed by pressing Ctrl+Z
In the case of pasting then the link cannot be removed.

## Steps to reproduce
1. Create a new Plain Text File or Markdown file.
2. Paste in a URL using either Ctrl+V or right-click Paste or right-click Paste as plain text
3.

## Expected behavior
The URL is made into links only in Documents not in plain text files. In Markdown files the URL should only be shortened on the "Preview" window and not in the edit window.

## Actual behavior
When pasting a URL such as https://outlook.com the URL gets automatically shortened to https://ou... you can neither read nor edit the URL.

## Setup
Proxmox 13 with Debian 14 LXC container with Docker Compose installed. Separate container running Caddy reverse proxy.

Details available on request, although I assume this is normal behaviour out of the box.

## Additional context
none

AlexAndBear:
…er than 120

## Related Issue

- Fixes #1410 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/web-pkg/src/components/TextEditor/TextEditor.vue
+++ b/packages/web-pkg/src/components/TextEditor/TextEditor.vue
@@ -44,7 +44,6 @@
         'htmlPreview',
         'pageFullscreen'
       ]"
-      :read-only="isReadOnly"
       auto-focus
       @on-change="(value: string) => $emit('update:currentContent', value)"
     />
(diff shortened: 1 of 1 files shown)


packages/web-pkg/src/components/TextEditor/TextEditor.vue
```diff
@@ -44,7 +44,6 @@
         'htmlPreview',
         'pageFullscreen'
       ]"
-      :read-only="isReadOnly"
```
AlexAndBear, line 47: not neccessary

Issue #1436, opened 2025-10-27
Personal breadcrumb item is missing


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

The breadcrumb item for "Personal" is missing for screen sizes >=640 <= 960px when the right sidebar is being opened.

### Reproduction Steps

- open your personal space
- open right sidebar
- Set window size to e.g. 800px

### Expected Outcome

The breadcrumb shows "Personal".

### Actual Outcome

The breadcrumb is empty.

This is a regression caused by https://github.com/opencloud-eu/web/pull/1419.

PR #1437, opened 2025-10-27
fix: add missing first breadcrumb on table size screens

JammingBen:
fixes https://github.com/opencloud-eu/web/issues/1436 (and a flaky unit test).

### DIFF
--- a/packages/web-app-admin-settings/src/views/Groups.vue
+++ b/packages/web-app-admin-settings/src/views/Groups.vue
@@ -194,7 +194,6 @@ export default defineComponent({
   computed: {
     breadcrumbs() {
       return [
-        { text: this.$gettext('Administration Settings'), to: { path: '/admin-settings' } },
         {
           text: this.$gettext('Groups'),
           onClick: () => {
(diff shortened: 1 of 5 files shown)


packages/web-app-admin-settings/src/views/General.vue
```diff
@@ -17,49 +21,26 @@
...
-    return {
-      template,
-      sideBarAvailablePanels,
-      ...useSideBar()
-    }
-  },
-  computed: {
-    breadcrumbs() {
-      return [
-        { text: this.$gettext('Administration Settings'), to: { path: '/admin-settings' } },
```
JammingBen, line 57: Doesn't make sense to have the 'admin-settings' breadcrumb item for the General page, it basically the same as General itself.

kulmann: Wouldn't it make sense to remove the breadcrumb item "Administration Settings" for all of the admin settings views? E.g. "Users" is not a sub-page of "General".

kulmann: In the "files" app we also don't show "Files" as first breadcrumb item on all the pages.

JammingBen: Yep I agree!

Issue #1420, opened 2025-10-23
"Set as space image" action is showing in personal space


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

The "Set as space image" action is showing in the personal space after leaving the trash overview.

### Reproduction Steps

- make sure you have at least one project space to enable the trash overview
- navigate to the trash overview
- navigate to the personal space
- right click any file

### Expected Outcome

There shouldn't be a "Set as space image" option.

### Actual Outcome

There is a "Set as space image" option.

PR #1439, opened 2025-10-27
fix: set space image action showing outside of project spaces

JammingBen:
fixes https://github.com/opencloud-eu/web/issues/1420


kulmann (approved):
💪

Issue #1454, opened 2025-10-28
Hide version check when request failed


kulmann:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

> * As a user, I don't want to be confused by a failing version check so that my FUD-level stays low

### User Value

Avoid FUD

### Acceptance Criteria

- if the version check failed
  - don't show the version check component at all
  - server version jumps to the bottom of the sidebar (= should be natural behaviour when the version check component is not mounted)
- if backend decides to fix CSP merging, then a failing version check wouldn't annoy people in the meantime

Issue #1453, opened 2025-10-28
Update check on Admin-Settings General page


kulmann:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

> * As an admin, I want to see on the "General" page of the admin-settings app if my version is up to date that I know when to update

### User Value

Consistency when displaying the version

### Acceptance Criteria

- Update info (Update available or version is up to date) is shown on the `General` page in the admin-settings app below the server version

PR #1455, opened 2025-10-28
feat: hide update check if server is not available, show update check in admin settings

AlexAndBear:
## Related Issue

- Fixes https://github.com/opencloud-eu/web/issues/1453
- Fixes https://github.com/opencloud-eu/web/issues/1454

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/web-app-admin-settings/src/components/General/InfoSection.vue
+++ b/packages/web-app-admin-settings/src/components/General/InfoSection.vue
@@ -2,15 +2,13 @@
   <div>
     <h2 class="py-2" v-text="$gettext('Info')" />
     <dl class="details-list grid grid-cols-[auto_minmax(0,1fr)]">
-      <dt v-text="$gettext('Edition')" />
-      <dd v-text="backendProductName" />
       <template v-if="backendEdition">
         <dt v-text="$gettext('Edition')" />
         <dt v-text="backendEdition" />
       </template>
-      <dt v-text="$gettext('Version')" />
+      <dt class="flex items-start" v-text="$gettext('Version')" />
       <dd>
-        <div class="flex items-start flex-col">
+        <div class="flex flex-col">
           <span v-text="backendVersion" />
           <version-check />
         </div>


packages/web-app-admin-settings/src/components/General/InfoSection.vue
```diff
@@ -2,16 +2,18 @@
   <div>
     <h2 class="py-2" v-text="$gettext('Info')" />
     <oc-definition-list :items="infoItems" />
+    <version-check class="mt-2" />
```
kulmann: I suppose there is no easy way to have the version string (e.g. `3.6.0+dev`) and the update info (e.g. `Up to date`) with the same horizontal alignment? 😢

kulmann: Image

like this

packages/web-app-admin-settings/src/components/General/InfoSection.vue
```diff
@@ -1,17 +1,32 @@
 <template>
   <div>
     <h2 class="py-2" v-text="$gettext('Info')" />
-    <oc-definition-list :items="infoItems" />
+    <dl class="details-list grid grid-cols-[auto_minmax(0,1fr)]">
+      <dt v-text="$gettext('Edition')" />
```
kulmann: Can we kick out the `OpenCloud` row entirely? `Edition` is not a good label and I can't think of a good one. `Edition Rolling` or `Edition Stable` would be fine to have. And also if there is only the row with the version would be fine as well.

packages/web-app-admin-settings/src/components/General/InfoSection.vue
```diff
@@ -1,17 +1,32 @@
...
+      <dd v-text="backendProductName" />
+      <template v-if="backendEdition">
+        <dt v-text="$gettext('Edition')" />
+        <dt v-text="backendEdition" />
+      </template>
+      <dt v-text="$gettext('Version')" />
```
kulmann: Both the label and the div with the values should be aligned to the top.

packages/web-pkg/tests/unit/components/VersionCheck.spec.ts
```diff
@@ -45,7 +45,9 @@ function getWrapper({
         plugins: [
           ...defaultPlugins({
             piniaOptions: {
-              capabilityState: { capabilities: { core: { status: { productversion } } } },
+              capabilityState: {
+                capabilities: { core: { status: { productversion }, 'check-for-updates': true } }
```
kulmann, line 49:
```suggestion
                capabilities: { core: { status: { productversion } } }
```

AlexAndBear: huh ? this would crash the test!

kulmann: ah lol, it's the spec file. sorry 😅

Issue #1443, opened 2025-10-27
Remove focus mixin


JammingBen:
The [focus mixin](https://github.com/opencloud-eu/web/blob/main/packages/web-runtime/src/mixins/focusMixin.ts) seems to be a left over from ancient times that is not being used anymore, hence we can safely remove it.

Issue #1409, opened 2025-10-22
Remove deprecated `requestRouter` method


JammingBen:
The `requestRouter` has been deprecated for quite a while now, I think we can savely remove it by now: https://github.com/opencloud-eu/web/blob/main/packages/web-runtime/src/container/api.ts#L131

Issue #1462, opened 2025-10-28
Refactor components to script setup


JammingBen:
This is a placeholder issue since it's a long-running task. We even still have the options API in place for some really old components.

As script setup also gives us improved runtime performance, it would make sense to prioritize the components that get rendered the most.

### Acceptance criteria

- No more options API, no more setup + return
- Props are defined via `defineProps`, using TS types and destructuring for default values
- Every slot and emit is defined via `defineEmits` and `defineSlots` respectively

PR #1464, opened 2025-10-28
refactor: remove deprecated requestRouter method and focus mixin

JammingBen:
And fixes an issue with the version check where the `oc-button` type was wrong.

fixes https://github.com/opencloud-eu/web/issues/1409
fixes https://github.com/opencloud-eu/web/issues/1443


kulmann (approved):
thanks for the cleanup! 🧹

Issue #1477, opened 2025-10-29
Space description edit moves table header


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

Editing the description of a space from the spaces list moves the table header somewhere it shouldn't be.

### Reproduction Steps

- go to the spaces list and create a space
- switch to the table view
- edit the space description via the context menu
- close the editor via the topbar "x"

### Expected Outcome

The table looks good.

### Actual Outcome

The table header sits somewhere it shouldn't.

<img width="611" height="184" alt="Image" src="https://github.com/user-attachments/assets/462c5461-1109-4c28-aef5-2a63dd801dff" />

PR #1498, opened 2025-11-03
fix: only render drop content when opened

JammingBen:
This fixes a small UI glitch when closing the file context menu and an unwanted displacement of the table header row in certain edge cases.

fixes https://github.com/opencloud-eu/web/issues/1477

### DIFF
--- a/packages/web-pkg/src/components/ContextActions/ContextMenuQuickAction.vue
+++ b/packages/web-pkg/src/components/ContextActions/ContextMenuQuickAction.vue
@@ -29,37 +29,25 @@
   </oc-button>
 </template>
 
-<script lang="ts">
-import { computed, defineComponent } from 'vue'
+<script setup lang="ts">
+import { computed } from 'vue'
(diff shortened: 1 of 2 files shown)


packages/design-system/src/components/OcDrop/OcDrop.spec.ts
```diff
@@ -53,25 +53,6 @@ describe('OcDrop', () => {
...
-    const { wrapper } = dom({ paddingSize: size })
-
-    const drop = wrapper.findComponent({ name: 'oc-drop' })
-    await nextTick()
-
-    const tailwindClassObj = getTailwindPaddingClass(size)
-    const tailwindClass = Object.entries(tailwindClassObj).find(([, v]) => v)?.[0]
-    expect(drop.html().includes(tailwindClass)).toBeTruthy()
-  })
-
```
JammingBen, line 74: I added a dedicated unit test for `getTailwindPaddingClass` instead of testing it implicitly here.

kulmann (approved):
thanks for the fix, seems to work 💪

PR #1552, opened 2025-11-06
fix: collabora not working in dev setup anymore

AlexAndBear:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -299,8 +299,6 @@ services:
     container_name: web_tika
     ports:
       - 9998:9998
-    networks:
-      - traefik
     restart: unless-stopped
 
   radicale:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -287,12 +291,16 @@ services:
...
+    networks:
+      - traefik
 
   tika:
     image: apache/tika:3.2.3.0
     container_name: web_tika
     ports:
       - 9998:9998
+    networks:
+      - traefik
```
kulmann: Tika is not public-facing, shouldn't need the traefik network.

docker-compose.yml
```diff
@@ -304,6 +312,8 @@ services:
       RADICALE_STORAGE_FILESYSTEM_FOLDER: /data
     ports:
       - '5232:5232'
+    networks:
+      - traefik
```
kulmann: Radicale is also not directly reachable, but only via opencloud proxy, so it should not be in the traefik network.

kulmann (approved):

PR #1612, opened 2025-11-13
adjust tests after fixing #1594

v-scharf:
related #1594

read from clipboard for webkit browser


ScharfViktor:
> Apparently it still fails in CI...? 

but why? it should work

maybe due difference operation systems? I run on mac oc, CI uses linux

PR #1613, opened 2025-11-13
feat: share links and passwords in embed mode

### LINKED ISSUE #1592: "Copy link and password" action makes no sense when creating multiple links
### Initial Checklist

### Bug Description

The "Copy link and password" action makes no sense when creating multiple links because we don't copy anything in this case.

### Reproduction Steps

- switch to embed mode via `&embed=true`
- select 2 or more files from the file list
- click "Share link(s)" on the bottom right side of the screen
- in the modal, click on the arrow right to the "Share link(s)" button

### Expected Outcome

The arrow and the action should not be present.

### Actual Outcome

They are present.

Image

dschmidt:
## Description

Change the "copy link and password" action in CreateLinkModal in embed mode to actually pass link(s) and password(s) to the parent application.

## Related Issue

- Fixes #1592
This is not the expected outcome, but I think it's the better solution. The ticket suggest to remove the button, I think it's better to pass the password with the url to the parent application. That way the parent application can insert the password with the link into a chat message or an email.
Yes, second channel would be better - but let's be honest, being forced to use a password in the default configuration is annoying enough for most users (me included).

## How Has This Been Tested?

- using it in a roundcube plugin I'm working on :) 
- easiest way to test this is to replace `postMessage<Array<{ url: string; password?: string }>>` with `console.log` and open OpenCloud with `?embed=true`

## Types of changes

- [x] Bugfix

### DIFF
--- a/packages/web-pkg/src/components/CreateLinkModal.vue
+++ b/packages/web-pkg/src/components/CreateLinkModal.vue
@@ -270,7 +270,7 @@ const onConfirm = async (options: { copyPassword?: boolean } = {}) => {
     const result = await createLinkHandler()
     const succeeded = result.filter(({ status }) => status === 'fulfilled')
     if (succeeded.length) {
-      // **DEPRECATED**: Always emit the share url for backwards compatibility
+      /** @deprecated Always emit the share url for backwards compatibility */
       postMessage<string[]>(
         'opencloud-embed:share',
         (succeeded as PromiseFulfilledResult<LinkShare>[]).map(({ value }) => value.webUrl)


packages/web-pkg/src/components/CreateLinkModal.vue
```diff
@@ -266,6 +266,29 @@ const createLinkHandler = async () => {
 }
 
 const onConfirm = async (options: { copyPassword?: boolean } = {}) => {
+  if (unref(isEmbedEnabled)) {
+    const result = await createLinkHandler()
+    const succeeded = result.filter(({ status }) => status === 'fulfilled')
+    if (succeeded.length) {
+      // **DEPRECATED**: Always emit the share url for backwards compatibility
```
JammingBen: Because we will grep for `@deprecated` in the future:

```suggestion
      /** @deprecated Always emit the share url for backwards compatibility */
```

packages/web-pkg/src/components/CreateLinkModal.vue
```diff
@@ -266,6 +266,29 @@ const createLinkHandler = async () => {
 }
 
 const onConfirm = async (options: { copyPassword?: boolean } = {}) => {
+  if (unref(isEmbedEnabled)) {
+    const result = await createLinkHandler()
+    const succeeded = result.filter(({ status }) => status === 'fulfilled')
```
dschmidt, line 271: Hm the filter could be a type guard... I'll fix

dschmidt: Actually I've seen there are a bunch of these  checks (one even with a local type guard) - I'd like to refactor this in a follow up PR

JammingBen (approved):
Thank you!

PR #1627, opened 2025-11-17
fix: move folder replace

JammingBen:
Fixes the "Replace" option of the conflict dialog when moving one folder into another.

The issue was the edge case check if a user moves a subfolder with the same name as the current folder into the parent of the current folder. I just added another check to ensure this only happens when moving a folder up in the hierarchy.

fixes https://github.com/opencloud-eu/web/issues/1617


kulmann:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|stable-4.2|[](https://github.com/opencloud-eu/web/pull/1662)|

### Questions ?

Issue #1628, opened 2025-11-17
Embed mode click area for selection


kulmann:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

When being in the web embed mode, the various options of creating a selection of multiple files/folders don't work as expected because the click area is not as expected.

### Reproduction Steps

1. Have at least 3 files in a folder
2. Open the web embed mode by appending `&embed=true` to the url
3. Try to select the first file
4. If you succeeded with (3), try to add the third file to the selection
5. If you succeeded with (3) and/or (4), try to select all files from first to last

### Expected Outcome

Various selection options work:

1. single select: works by clicking on checkbox, file name, preview/icon, whitespace, any button
2. multi select: click area as described in (1) can be used for multi selects via `ctrl/cmd` or `shift` keys

### Actual Outcome

Click area is very limited: Only clicking the checkbox or the three-dots-button works.

Issue #1518, opened 2025-11-05
Extract duplicated logic in files table and tiles view


JammingBen:
We have a lot of duplicated logic in the `ResourceTable` and the `ResourceTiles` components. Some of it should be extracted to some composables to remove unnecessary code duplication.

PR #1641, opened 2025-11-18
fix: embed mode click behavior

JammingBen:
Fixes several issues with the click behavior in the embed mode.

- hide select all checkbox in tiles view with file picker enabled
- fix click behavior in tiles view with embed enabled
- make tiles not selectable via keyboard with file picker enabled
- prevent files from getting downloaded when selecting them in table view with file picker enabled

Moves quite a bit of shared logic between the table and the tiles view into one helper composable. Mainly the ones for clicking files (or their containers/checkboxes).

fixes https://github.com/opencloud-eu/web/issues/1628
part of https://github.com/opencloud-eu/web/issues/1518


packages/web-pkg/src/components/FilesList/ResourceTile.vue
```diff
@@ -59,7 +59,7 @@
               }"
               :src="resource.thumbnail"
               :data-test-thumbnail-resource-name="resource.name"
-              @click="toggleTile([resource, $event])"
+              @click.stop="$emit('tileClicked', [resource, $event])"
```
kulmann, line 62: I don't understand why `tileClicked` exists, it is never consumed anywhere 🙈

JammingBen: It's consumed in the `ResourceTiles` component.

kulmann: Ah, didn't check for `@tile-clicked`... -.- sorry for the noise

kulmann (approved):
😍

Issue #1609, opened 2025-11-13
Mail: Mail compose as Widget


AlexAndBear:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

> * As a user I want to compose emails inside a widget, so that I'm able to navigate through my inbox at the same time in the desktop view.

### User Value

Easy accessible 

### Acceptance Criteria

- New mail form must be displayed in a widget view instead of a new route (current state)
- The widget must have "New message" title at the topbar
- The widget must have a expanded view toggle at the topbar 
- The widget must have a close button at the topbar
- The user must be able to navigate trough their inbox (widget stays always on top until we have a "draft bar")
- In mobile view the widget must be a full size window (no widget-like apperance) and current state muste be aligned to v2 agency proposal 

<img width="689" height="499" alt="Image" src="https://github.com/user-attachments/assets/4777e808-70f7-4068-a559-b9514c9aa726" />

<img width="337" height="627" alt="Image" src="https://github.com/user-attachments/assets/7c8090a9-a79d-477a-b967-b0f31583242a" />

PR #1687, opened 2025-12-01
Mail: Mail compose as Widget #1609

tammi-23:
PR for #1609

### DIFF
--- a/packages/design-system/src/components/OcModal/OcModal.vue
+++ b/packages/design-system/src/components/OcModal/OcModal.vue
@@ -18,8 +18,8 @@
           class="oc-modal-title bg-role-surface-container flex items-center flex-row flex-wrap justify-between py-3 px-4 rounded-t-sm"
         >
           <h2 id="oc-modal-title" class="truncate m-0 text-base" v-text="title" />
-          <div v-if="$slots['modal-slot-header-actions']" class="flex items-center gap-1">
-            <slot name="modal-slot-header-actions" />
+          <div v-if="$slots['headerActions']" class="flex items-center gap-1">
+            <slot name="headerActions" />
           </div>
         </div>
         <div class="oc-modal-body px-4 pt-4">
@@ -214,6 +214,11 @@ export interface Slots {
    * @docs Custom content of the modal.
    */
   content?: () => unknown
+
+  /**
+   * @docs Custom actions in the modal header.
+   */
+  headerActions?: () => unknown
 }
 
 const {
--- a/packages/web-app-mail/src/components/MailWidget.vue
+++ b/packages/web-app-mail/src/components/MailWidget.vue
@@ -4,7 +4,7 @@
     class="z-50 transition absolute inset-0 md:fixed md:inset-0 pointer-events-auto md:pointer-events-none bg-transparent"
   >
     <div
-      class="oc-mail-compose-widget pointer-events-auto absolute bg-role-surface border-0 md:border md:border-role-outline-variant flex flex-col md:rounded-xl top-0 left-0 right-0 bottom-0 md:top-auto md:bottom-2 md:left-auto md:right-8 md:w-[720px] md:h-[800px]"
+      class="oc-mail-compose-widget pointer-events-auto absolute bg-role-surface border-0 md:border md:border-role-outline-variant flex flex-col md:rounded-xl top-0 left-0 right-0 bottom-0 md:top-auto md:bottom-2 md:left-auto md:right-8 md:w-[720px] md:h-[800px] md:max-h-[80vh]"
     >
       <div class="flex items-center justify-between px-4 py-2">
         <h2
@@ -30,7 +30,7 @@
           </oc-button>
         </div>
       </div>
-      <div class="flex-1 overflow-auto">
+      <div class="flex-1 overflow-y-auto overflow-x-hidden min-h-0">
         <MailComposeForm v-model="composeState" />
       </div>
       <div class="px-4 py-3 border-t border-role-outline-variant flex items-center justify-between">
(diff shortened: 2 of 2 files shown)


packages/design-system/src/components/OcModal/OcModal.vue
```diff
@@ -18,6 +18,9 @@
           class="oc-modal-title bg-role-surface-container flex items-center flex-row flex-wrap py-3 px-4 rounded-t-sm"
         >
           <h2 id="oc-modal-title" class="truncate m-0 text-base" v-text="title" />
+          <div v-if="$slots['modal-slot-header-actions']" class="ml-auto flex items-center gap-1">
```
AlexAndBear: Instead of ml-auto please use justify-content: space-between on the outer div if possible.

kulmann (commented):
You need to kill the import for `MailCompose.vue` and the corresponding route in `index.ts`.

packages/web-app-mail/src/components/MailComposeForm.vue
```diff
@@ -0,0 +1,167 @@
...
+export type ComposeFormState = {
+  from: FromOption | undefined
+  to: string
+  cc: string
+  bcc: string
+  subject: string
+  body: string
+}
+
+const loadAccountsTask = useTask(function* (signal) {
```
AlexAndBear: Please use the accounts store. We don't need to load the accounts here anymore

packages/web-app-mail/src/components/MailList.vue
```diff
@@ -55,26 +58,27 @@
           </oc-button>
         </li>
       </oc-list>
+      <MailWidget v-model="showCompose" />
```
AlexAndBear, line 61: Why not a simple v-if instead of v-model ?

packages/web-app-mail/src/components/MailboxTree.vue
```diff
@@ -70,6 +69,10 @@ import { useLoadMails } from '../composables/useLoadMails'
 import { unref } from 'vue'
 import MailAccountSwitch from './MailAccountSwitch.vue'
 
+const emit = defineEmits<{
+  (e: 'compose-mail'): void
```
JammingBen: Please use camelCase for event names because of consistency.

```suggestion
  (e: 'composeMail'): void
```

packages/web-app-mail/src/views/Inbox.vue
```diff
@@ -64,11 +64,16 @@ const { loadMails } = useLoadMails()
 const { loadMail } = useLoadMail()
 
 const isLoading = ref<boolean>(true)
+const mailListRef = ref<InstanceType<typeof MailList> | null>(null)
```
JammingBen:
```suggestion
const mailListRef = useTemplateRef<ComponentPublicInstance<typeof MailList>>('mailListRef')
```

packages/web-app-mail/src/components/MailComposeForm.vue
```diff
@@ -0,0 +1,167 @@
...
+)
+
+onMounted(() => {
+  loadAccountsTask.perform()
+})
+
+defineExpose({
+  fromOptions,
+  isLoading
+})
```
JammingBen: What's the reason your exposing these? I don't see them being used outside the component 🤔 In general, working with exposed props and methods should be a kind of 'last resort' if it can be solved otherwise.

packages/web-app-mail/src/components/MailWidget.vue
```diff
@@ -0,0 +1,150 @@
...
+  'z-50 transition',
+  'absolute inset-0 md:fixed md:inset-0',
+  'pointer-events-auto md:pointer-events-none bg-transparent'
+])
+
+const widgetPositionClasses = computed(
+  () =>
+    'top-0 left-0 right-0 bottom-0 ' +
+    'md:top-auto md:bottom-2 md:left-auto md:right-8 md:w-[720px] md:h-[800px]'
+)
```
JammingBen: I don't think they need to be `computed` properties since they are not expected to change at any time. Why not just add them in the template above? 😅 You can also pass then as a list if you decide it's more readable.

packages/web-app-mail/src/components/MailList.vue
tammi-23: The v-model give two-way binding between the parent showCompose and the child isOpen.

packages/web-app-mail/src/components/MailComposeForm.vue
tammi-23: Good catch, fromOptions is only used inside the component template, so theres no need to expose it. isLoading was already removed.

packages/design-system/src/components/OcModal/OcModal.vue
```diff
@@ -15,9 +15,12 @@
         @keydown.esc.stop="cancelModalAction"
       >
         <div
-          class="oc-modal-title bg-role-surface-container flex items-center flex-row flex-wrap py-3 px-4 rounded-t-sm"
+          class="oc-modal-title bg-role-surface-container flex items-center flex-row flex-wrap justify-between py-3 px-4 rounded-t-sm"
         >
           <h2 id="oc-modal-title" class="truncate m-0 text-base" v-text="title" />
+          <div v-if="$slots['modal-slot-header-actions']" class="flex items-center gap-1">
+            <slot name="modal-slot-header-actions" />
+          </div>
```
JammingBen: Please also define this slot in the `Slots` interface down below for it to appear in our design-system docs.

Also,  `modal-slot-` seems a bit redundant in the naming, don't you think? I'd just name it `headerActions`.

AlexAndBear (approved):
Top notch ⭐

PR #1699, opened 2025-12-03
feat: add photo roll to preview app 

AlexAndBear:
## Description

image

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/web-app-preview/src/App.vue
+++ b/packages/web-app-preview/src/App.vue
@@ -11,6 +11,8 @@
     tabindex="-1"
     @keydown.left="goToPrev"
     @keydown.right="goToNext"
+    @keydown.down="goToNext"
+    @keydown.up="goToPrev"
   >
     <photo-roll
       v-if="photoRollEnabled"
@@ -440,8 +442,6 @@ export default defineComponent({
           emit('delete:resource', unref(activeFilteredFile))
         )
       )
-      keyBindings.push(bindKeyAction({ primary: Key.ArrowDown }, () => goToNext()))
-      keyBindings.push(bindKeyAction({ primary: Key.ArrowUp }, () => goToPrev()))
     })
 
     onBeforeUnmount(() => {
--- a/packages/web-app-preview/src/components/MediaControls.vue
+++ b/packages/web-app-preview/src/components/MediaControls.vue
@@ -26,6 +26,16 @@
       >
         <oc-icon size="large" name="arrow-drop-right" />
       </oc-button>
+      <oc-button
+        v-oc-tooltip="togglePhotoRollDescription"
+        class="raw-hover-surface p-1 hidden md:flex"
+        data-testid="toggle-photo-roll"
+        appearance="raw"
+        :aria-label="togglePhotoRollDescription"
+        @click="$emit('togglePhotoRoll')"
+      >
+        <oc-icon name="side-bar" :fill-type="photoRollEnabled ? 'fill' : 'line'" />
+      </oc-button>
       <div class="flex">
         <oc-button
           v-oc-tooltip="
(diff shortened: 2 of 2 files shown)


packages/web-app-preview/src/App.vue
```diff
@@ -318,6 +336,12 @@ export default defineComponent({
       } as RouteLocationRaw)
     }
 
+    const preloadImages = () => {
+      for (const file of unref(filteredFiles)) {
+        loadFileIntoCache(file)
```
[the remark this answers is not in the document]
JammingBen: This is intentional.

packages/web-app-preview/src/components/MediaControls.vue
```diff
@@ -107,6 +107,16 @@
       >
         <oc-icon fill-type="line" name="delete-bin" />
       </oc-button>
+      <oc-button
```
kulmann: not entirely sure, just a gut feeling... should the button be the first button on the left, because the photo roll is also left? 🤔

packages/web-app-preview/src/App.vue
```diff
@@ -377,6 +421,40 @@ export default defineComponent({
...
+        bindKeyAction({ modifier: Modifier.Ctrl, primary: Key.Backspace }, () =>
+          emit('delete:resource', unref(activeFilteredFile))
+        )
+      )
+      keyBindings.push(
+        bindKeyAction({ primary: Key.Delete }, () =>
+          emit('delete:resource', unref(activeFilteredFile))
+        )
+      )
+      keyBindings.push(bindKeyAction({ primary: Key.ArrowDown }, () => goToNext()))
```
kulmann: The right and left arrow keys have a binding via template (`@keydown.left="goToPrev"` and `@keydown.right="goToNext"`). Any reason why we have two different ways of registering key bindings here?

packages/web-app-preview/src/App.vue
```diff
@@ -184,6 +196,18 @@ export default defineComponent({
       })
     })
 
+    const photoRollFiles = computed(() => {
```
kulmann, line 201: Why do we need a new array instead of directly using `filteredFiles`? 🤔

packages/web-app-preview/src/components/MediaControls.vue
JammingBen: Hmm I feel like the navigation should stay on the left side. Maybe after that, before the fullscreen button?

packages/web-app-preview/src/App.vue
JammingBen: If I remember correctly, `filteredFiles` are just regular resources without the preview URL that is needed to display the images. Only `cachedFiles` has that.

packages/web-app-preview/src/App.vue
JammingBen: Because I couldn't find it 😂 Thanks for the hint!

packages/web-app-preview/src/App.vue
kulmann: I don't want to say `sounds like that should be refactored`.... there, I said it. 😆 but fine for now... refactoring that mess is for another day I guess.

packages/web-app-preview/src/components/MediaControls.vue
kulmann: Yes, that's good for now I think. Let's wait for user feedback. I definitely like, that photo roll toggle button and fullscreen toggle button are next to each other. Because when you enter fullscreen it's a bit more likely that you want to hide the photo roll I think... again, just guessing. Let's wait for user feedback.

PR #1779, opened 2025-12-19
enforce password for writable public links

v-scharf:
related https://github.com/opencloud-eu/opencloud/issues/2048

- Check the need to entered password when changing a public link to a role with write permission


JammingBen (approved):
Code LGTM, formatter complains though.

Issue #1788, opened 2025-12-20
Access and Refresh tokens are visible in browser console


krambo-ss:
## Describe the bug
The Access Token and Refresh Token are both shown in browser's console.

## Steps to reproduce
1. Login to OpenCloud
2. Open browser's dev tools (firefox in my case)
3. Wait until "AccessToken Expiring" appears

## Expected behavior
Those two should not be visible

## Actual behavior
The Access Token and Refresh Token are both shown in browser's console.

## Setup
OpenCloud production image (version 4.0.1 stable) was spun up with Keycloak, Collabora, and Radicale. Using my own nginx server for reverse proxy. Main functionalities appear to work fine.

I'm pasting below my sanitized .env file contents and a screenshot of said browser console.

<img width="1141" height="550" alt="Image" src="https://github.com/user-attachments/assets/44e8bd72-3c9a-41e9-ac12-cf6a197099f2" />

<details>
<p>

```console
LOG_DRIVER=
INSECURE=true
COMPOSE_FILE=docker-compose.yml:weboffice/collabora.yml:external-proxy/opencloud-exposed.yml:external-proxy/collabora-exposed.yml:idm/ldap-keycloak.yml:external-proxy/keycloak-exposed.yml:radicale/radicale.yml
TRAEFIK_DASHBOARD=
TRAEFIK_DOMAIN=
TRAEFIK_BASIC_AUTH_USERS=
TRAEFIK_ACME_MAIL=
TRAEFIK_ACME_CASERVER=
TRAEFIK_SERVICES_TLS_CONFIG="tls.certresolver=letsencrypt"
TRAEFIK_ACCESS_LOG=
TRAEFIK_LOG_LEVEL=
OC_DOCKER_IMAGE=opencloudeu/opencloud
OC_DOCKER_TAG=
OC_DOMAIN=<removed>
DEMO_USERS=
INITIAL_ADMIN_PASSWORD=<removed>
LOG_LEVEL= warn
LOG_PRETTY=true
OC_CONFIG_DIR=/opt/opencloud-data/config
OC_DATA_DIR=/opt/opencloud-data/data
OC_APPS_DIR=/opt/opencloud-data/apps
DECOMPOSEDS3_ENDPOINT=
DECOMPOSEDS3_REGION=
DECOMPOSEDS3_ACCESS_KEY=
DECOMPOSEDS3_SECRET_KEY=
DECOMPOSEDS3_BUCKET=
SMTP_HOST=
[...]

PR #1789, opened 2025-12-23
fix: hide access_token and refresh_token details in the web console

fschade:
## Description

Removes logging of access and refresh tokens to prevent them from potentially leaking through browser extensions, addons, or .... Replaces token logs with hashed fingerprints. 

Also introduces a logger that only logs messages when the appropriate log level is set.

## Related Issue

- Fixes https://github.com/opencloud-eu/web/issues/1788

## How Has This Been Tested?

- locally in the browser

## Types of changes

- [X] Bugfix
- [X] Enhancement (a change that doesn't break existing code or deployments)


packages/web-runtime/src/services/auth/authService.ts
```diff
@@ -160,19 +161,25 @@ export class AuthService implements AuthServiceInterface {
...
-            `New User Loaded. access_token： ${user.access_token}, refresh_token: ${user.refresh_token}`
-          )
+          logger.debug(`User Loaded`, {
+            ...(user.access_token && {
+              'access_token (sha256)': sha256(Buffer.from(user.access_token)).slice(0, 8)
+            }),
+            ...(user.refresh_token && {
+              'refresh_token (sha256)': sha256(Buffer.from(user.refresh_token)).slice(0, 8)
+            })
+          })
```
JammingBen, line 171: Hmm I'm not sure if hashing the tokens has any value because you can't decode them. Why not print them as they are since we're only doing this in dev mode?

fschade: console.log can be overwritten which could leak the token to the outside, also services like sentry or similar can access the data:

https://skillstuff.com/if-you-store-tokens-like-this-youre-already-vulnerable/

JammingBen: Ahh I didn't know, thanks for the info!

JammingBen (approved):
Code LGTM, needs a rebase though to make CI happy.

packages/web-runtime/src/services/auth/authService.ts
kulmann: but why are we logging anything at all then? what benefit does a hashed token bring?

fschade: > but why are we logging anything at all then? what benefit does a hashed token bring?

I wasn't sure why we log the token either; the only case I can think of is to see that the token refresh is working, that the token is new, and possibly to compare the hash.

kulmann: IMO you can remove it entirely 🙈

fschade: > IMO you can remove it entirely 🙈

+1, what do you think, should we keep the logger, I prefer it over the native console.*

kulmann: I have the impression that it's a bit over-engineered to be honest

fschade: > I have the impression that it's a bit over-engineered to be honest

Ok, maybe … 😂, let’s keep the pr open till after my vacation, or just remove the token logs and call it a day 🤗

AlexAndBear:
This is too complex, we decided on not showing this information at all

PR #1817, opened 2026-01-13
feat: folder editor apps

kulmann:
## Description

This PR brings a set of changes to be able to register editor apps on folders.

Required for https://github.com/opencloud-eu/web-extensions/pull/327

I decided to clean up the signature of `useFileActions`, because it was rather confusing. This is breaking.

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/web-pkg/src/components/FilesList/ContextActions.vue
+++ b/packages/web-pkg/src/components/FilesList/ContextActions.vue
@@ -115,10 +115,7 @@ export default defineComponent({
     })
 
     const menuItemsContextDrop = computed(() => {
-      return [
-        ...getAllOpenWithActions({ ...unref(actionOptions), omitSystemActions: true }),
-        ...unref(extensionsContextActions).filter((a) => a.category === 'context')
-      ]
+      return getAllOpenWithActions({ ...unref(actionOptions), omitSystemActions: true })
         .filter((item) => item.isVisible(unref(actionOptions)))
         .sort((x, y) => Number(y.hasPriority) - Number(x.hasPriority))
     })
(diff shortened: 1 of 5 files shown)


packages/web-client/src/helpers/resource/functions.ts
```diff
@@ -142,7 +139,7 @@ export function buildResource(
     parentFolderId: resource.props[DavProperty.FileParent],
     mimeType: resource.props[DavProperty.MimeType],
     name,
-    extension: isFolder ? '' : extension,
+    extension,
```
kulmann, line 142: we also need the extension for folders so that we can register editor apps on folder extensions (since all folders have the same mimetype).

packages/web-pkg/src/components/FilesList/ContextActions.vue
```diff
@@ -109,14 +109,14 @@ export default defineComponent({
     )
 
     const menuItemsContext = computed(() => {
-      return [...unref(navigateActions), ...unref(defaultEditorActions)]
+      return unref(openWithDefaultActions)
```
kulmann, line 112: `navigate` is no longer treated as a system action, but is part of the possible default actions on a file (not so much) or folder (yep).

packages/web-pkg/src/composables/actions/files/index.ts
```diff
@@ -1,29 +1,30 @@
 export * from './useFileActions'
-export * from './useFileActionsEnableSync'
-export * from './useFileActionsToggleHideShare'
 export * from './useFileActionsCopy'
 export * from './useFileActionsCopyPermanentLink'
-export * from './useFileActionsDisableSync'
+export * from './useFileActionsCreateLink'
```
kulmann, line 4: re-ordered the exports lexicographically so that it's harder for you to spot the new one 😇

JammingBen (approved):
Awesome! 🚀

Issue #1478, opened 2025-10-29
Mail: Integrate TipTap for writing emails


kulmann:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

> * As a user, I want to write comprehensive email content with an easy to use editor so that my email content is awesome.

### User Value

UX for writing email content

### Acceptance Criteria

- integrate [TipTap](https://tiptap.dev) for writing email content 
- tiptop replaces the textarea introduced via https://github.com/opencloud-eu/web/issues/1476

PR #1831, opened 2026-01-15
Issues/1478 integrate tiptap for writing emails

tammi-23:
Integrate TipTap for writing emails #1478

### DIFF
--- a/packages/design-system/src/components/OcModal/OcModal.vue
+++ b/packages/design-system/src/components/OcModal/OcModal.vue
@@ -18,8 +18,8 @@
           class="oc-modal-title bg-role-surface-container flex items-center flex-row flex-wrap justify-between py-3 px-4 rounded-t-sm"
         >
           <h2 id="oc-modal-title" class="truncate m-0 text-base" v-text="title" />
-          <div v-if="$slots['modal-slot-header-actions']" class="flex items-center gap-1">
-            <slot name="modal-slot-header-actions" />
+          <div v-if="$slots['headerActions']" class="flex items-center gap-1">
+            <slot name="headerActions" />
           </div>
         </div>
         <div class="oc-modal-body px-4 pt-4">
--- a/packages/web-app-mail/src/components/MailBodyEditor.vue
+++ b/packages/web-app-mail/src/components/MailBodyEditor.vue
@@ -1,6 +1,6 @@
 <template>
-  <div class="mail-body-editor flex flex-col gap-2 h-full">
-    <div class="mail-body-editor-editor-wrapper flex-1" @click="onWrapperClick">
+  <div class="mail-body-editor flex flex-col gap-2 h-full min-h-0">
+    <div class="mail-body-editor-editor-wrapper flex-1 min-h-0" @click="onWrapperClick">
       <EditorContent v-if="editor" :editor="editor" class="mail-body-editor-editor" />
     </div>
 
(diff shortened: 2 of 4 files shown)


packages/design-system/src/components/OcModal/OcModal.vue
```diff
@@ -18,8 +18,8 @@
           class="oc-modal-title bg-role-surface-container flex items-center flex-row flex-wrap justify-between py-3 px-4 rounded-t-sm"
         >
           <h2 id="oc-modal-title" class="truncate m-0 text-base" v-text="title" />
-          <div v-if="$slots['headerActions']" class="flex items-center gap-1">
-            <slot name="headerActions" />
+          <div v-if="$slots['modal-slot-header-actions']" class="flex items-center gap-1">
+            <slot name="modal-slot-header-actions" />
```
JammingBen: This looks like a rebase gone wrong, no?

packages/web-app-mail/src/components/MailBodyEditor.vue
```diff
@@ -0,0 +1,354 @@
...
+            <oc-button appearance="raw" @click="cancelLink">
+              <span v-text="$gettext('Cancel')" />
+            </oc-button>
+            <oc-button appearance="filled" @click="applyLink">
+              <span v-text="$gettext('Apply')" />
+            </oc-button>
+          </div>
+        </div>
+      </template>
+    </oc-modal>
```
JammingBen: Ideally, modals should be created programmatically via the `dispatchModal` method of the modal store. Any custom component can be used via the `customComponent` prop if needed. You can check out e.g. how the create link modal is being created: https://github.com/opencloud-eu/web/blob/main/packages/web-pkg/src/composables/actions/files/useFileActionsCreateLink.ts#L31

While it also works the way you did it, we prefer the other approach because you don't need to care about things like handling the visibility state of the modal or making sure the confirm method is included in the loading service.

packages/web-app-mail/src/components/MailComposeFormattingToolbar.vue
```diff
@@ -0,0 +1,169 @@
...
+
+<script setup lang="ts">
+import type { Editor } from '@tiptap/vue-3'
+
+const toolbarProps = defineProps<{
+  editor: Editor | null
+}>()
+
+const emit = defineEmits<{
+  (e: 'open-link-modal'): void
```
JammingBen, line 133:
```suggestion
  (e: 'openLinkModal'): void
```

packages/web-app-mail/src/components/MailComposeFormattingToolbar.vue
```diff
@@ -0,0 +1,169 @@
...
+  toolbarProps.editor.chain().focus().toggleOrderedList().run()
+}
+
+const toggleBlockquote = () => {
+  if (!toolbarProps.editor) return
+  toolbarProps.editor.chain().focus().toggleBlockquote().run()
+}
+
+const emitOpenLinkModal = () => {
+  emit('open-link-modal')
```
JammingBen, line 167:
```suggestion
  emit('openLinkModal')
```

packages/web-app-mail/src/components/MailWidget.vue
```diff
@@ -30,24 +30,42 @@
...
     </div>
   </div>
+
   <oc-modal
     v-if="isOpen && isExpanded"
     :title="$gettext('New message')"
     :hide-actions="true"
     element-class="mail-compose-modal"
   >
-    <template #headerActions>
```
JammingBen, line 50: Another left over from the rebase gone wrong? `headerActions` should be the correct one.

packages/web-app-mail/src/components/MailComposeFormattingToolbar.vue
```diff
@@ -0,0 +1,169 @@
+<template>
+  <div class="mail-body-editor-toolbar-shell">
```
JammingBen, line 2: Do you need all the `mail-body-editor-toolbar-` classes? With Tailwind CSS, we usually don't need to bloat the DOM with custom classes, and I don't see any CSS instructions for them  👀

packages/web-app-mail/src/views/Inbox.vue
```diff
@@ -9,7 +9,7 @@
         }"
       >
         <div class="overflow-y-auto md:border-r-2 bg-role-surface-container w-full">
-          <MailboxTree @compose-mail="onComposeMail" />
+          <MailboxTree @composeMail="onComposeMail" />
```
JammingBen: Running `pnpm lint --fix` will undo this change. Is it something you did or your IDE? 🤔  Maybe a wrong config?

It's a bit confusing: events are declared via camelCase, but when using them in the template, you use snake-case (as with all HTML attributes).

packages/design-system/src/components/OcModal/OcModal.vue
tammi-23: Yes, true.

packages/web-app-mail/src/components/MailComposeFormattingToolbar.vue
tammi-23: Good catch 👀 — some of those mail-body-editor-toolbar-* classes are there for special handling. That said, a few might be legacy/no-ops now. I can audit them and remove anything that isn’t used.

Issue #1806, opened 2026-01-09
Composable `useFileActionsDelete` can only be used by the files app


kulmann:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

The `isVisible` callback of the `useFileActionsDelete` composable is very restrictive. It requires one of three routes of the files app to be the current route. This makes the composable unusable in other apps.

### Reproduction Steps

1. Embed `useFileActionsDelete` composable in another app
2. Call `isVisible` of the actions correctly with `space` and `resources` in `actionOptions`

### Expected Outcome

A `Delete` action becomes available on files or folders the current user has write access to

### Actual Outcome

No `Delete` action becomes available, ever.

PR #1832, opened 2026-01-15
fix: resource delete action visibility

JammingBen:
Fixes the visibility of the delete actions of the `useFileActionsDelete` composable outside of the files app. To achieve this, I removed all the route-specific logic in favor of resource-specific checks.

Note that this leads to a slightly different behavior on the search result page: When selecting e.g. 5 resources and one of them is not deletable for some reason, the delete action will not be displayed as a batch action.

fixes https://github.com/opencloud-eu/web/issues/1806


kulmann (approved):
Awesome, thanks for cleaning this up and making it usable in apps 🤓

Issue #1696, opened 2025-12-02
[Bug] Sidbar button/text are cut off


suse-coder:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

I am on Version [4.0.0-rc.1](https://github.com/opencloud-eu/opencloud/releases/tag/4.0.0-rc.1)
When i click on a space -> right sidebar and add a public link then i see it and can copy it. 
But inside the space when I click on a file -> the public link section is to wide and one cant see it in the browser window anymore.

<img width="1131" height="602" alt="Image" src="https://github.com/user-attachments/assets/6272d9cc-7994-4137-82b5-6681aac3bff7" />
<img width="1880" height="560" alt="Image" src="https://github.com/user-attachments/assets/8941323e-a95d-4df6-b1c2-bbdaf8e3c153" />
<img width="608" height="930" alt="Image" src="https://github.com/user-attachments/assets/e1e97ce9-a9ca-4826-8b41-4808ebd958a0" />

### Reproduction Steps

1. Click on Space -> see right sidebar the public link section all fine
2. click on a file (share) in that space -> the public link section is to wide and one cant see it in the browser window anymore

### Expected Outcome

Correct screen size

### Actual Outcome

Sidebar width off screen

PR #1836, opened 2026-01-16
fix: break long resource names in right sidebar

JammingBen:
fixes https://github.com/opencloud-eu/web/issues/1696

Also refactors `SpaceInfo` and `ResourceName` to script setup (see individual commits).


kulmann (approved):
Nitpick: the second commit has a typo in the commit message 🙈 (`refactpr:` instead of `refactor:`).
Thanks for the fix and the script setup goodness!

Issue #1608, opened 2025-11-12
Text editor should suggest 'save as' to resolve conflicts.


jnweiger:
### Initial Checklist

- [x] I understand this is a feature request and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Idea

Seen with opencloud 3.7.0

* Admin creates a folder 'sha', and shares it with Mary read/write. 
* Admin adds a text file Mary-here.txt with a few lines of text.
* Mary (in a different browser) opens the web UI text editor and continues to add text to the file.
* Between two autosaves in Mary's browser, admin renames the parent folder from `sha` to `share`.
* Mary sees error messages: 

<img width="1742" height="847" alt="Image" src="https://github.com/user-attachments/assets/9976c793-dca2-4687-8ea5-751932d5cd3c" />

Doing what is suggested "Please refresh this page" results in what is predicted: All work is lost. BAD.

Expected behaviour:
* A "Save As" button should allow to save with a different name.
* The message should at least say "Please copy your contents elsewhere before refreshing this page"

Workaround:
* Mary needs to manually copy paste all new text elsewhere to rescue her work.

db-ot:
Suggestion change wording:

"Beim Speichern ist ein Fehler aufgetreten:
Das Dokument zwischenzeitlich von einer anderen Person geändert wurde. Bitte kopieren Sie Ihre Änderungen oder speichern Sie die Datei unter einem neuen Namen („Speichern unter…“)."

"An error occurred while saving:
The document was modified by another user in the meantime. Please copy your changes or save the file under a new name (“Save As…”)."

PR #1850, opened 2026-01-20
feat: clarify error text on 412 error in editor

JammingBen:
closes https://github.com/opencloud-eu/web/issues/1608

### DIFF
--- a/packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
+++ b/packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
@@ -420,7 +420,7 @@ const saveFileTask = useTask(function* () {
         errorPopup(
           new HttpError(
             $gettext(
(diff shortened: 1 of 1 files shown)


packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
```diff
@@ -420,7 +420,7 @@ const saveFileTask = useTask(function* () {
         errorPopup(
           new HttpError(
             $gettext(
-              'This file was updated outside this window. Please refresh the page (all changes will be lost).'
+              'This file was updated outside this window. Please copy your changes or save the file under a new name (“Save As…”).'
```
kulmann:
```suggestion
              'This file was updated outside this window. Please copy your changes or save the file under a new name (»Save As...«).'
```

We usually use »« for quoting file names etc. I'd say we do that in this case as well. What do you think?

Issue #1642, opened 2025-11-18
Tiles view is missing lowered opacity for cut resources


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

The tiles view is missing the lowered opacity for cut resources.

### Reproduction Steps

- switch to tiles view
- create a file
- cut the resource via ctrl/cmd + x

### Expected Outcome

The resource should have a slightly lower opacity compared to other files, just like in the table view.

### Actual Outcome

The resource looks like every other resource.

Related: https://github.com/opencloud-eu/web/issues/1518

PR #1852, opened 2026-01-20
fix: add missing low opacity for cut resources in tiles view

JammingBen:
fixes https://github.com/opencloud-eu/web/issues/1642

### DIFF
--- a/packages/web-pkg/src/components/FilesList/ResourceTiles.vue
+++ b/packages/web-pkg/src/components/FilesList/ResourceTiles.vue
@@ -63,7 +63,7 @@
           :draggable="dragDrop"
           :lazy="areTilesLazy"
           :is-loading="isResourceInDeleteQueue(resource.id)"
-          :class="{ 'opacity-70': isResourceCut(resource) }"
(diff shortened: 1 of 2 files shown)


packages/web-pkg/src/components/FilesList/ResourceTiles.vue
```diff
@@ -63,6 +63,7 @@
           :draggable="dragDrop"
           :lazy="areTilesLazy"
           :is-loading="isResourceInDeleteQueue(resource.id)"
+          :class="{ 'opacity-70': isResourceCut(resource) }"
```
kulmann: Do you know why we went for `opacity-70` in the first place? At least in the tiles view the difference is barely visible... maybe because of the greyish background of the tiles. In the list view it's a tiny bit better, but also not very visible.

JammingBen: No idea. Should I increase it a bit?

kulmann: as discussed: we'll roll with `opacity-60`

Issue #1479, opened 2025-10-29
Mail: Add attachments when writing emails


kulmann:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

> * As a user, I want to add file attachments from my device to the email so that I can send files.

### User Value

Writing comprehensive email

### Acceptance Criteria

- Email composer footer has a button for adding attachments (paperclip icon)
- clicking the paperclip button opens an upload context menu / bottom drawer
- Attachment gets added as blob to the draft (? clarify with Pascal)
- Attachments can be removed from the draft again by clicking their respective delete button
- re-use the attachments card for visualising already added attachments
- not supported in this first iteration:
  - drag & drop
  - copy & paste
  - inline attachments

<img width="375" height="812" alt="Image" src="https://github.com/user-attachments/assets/4268251d-85ba-4703-a14e-6d015dd5bd2d" />

<img width="375" height="1075" alt="Image" src="https://github.com/user-attachments/assets/9aa66b74-2803-45e4-9d39-ae25eed39e59" />

PR #1873, opened 2026-01-26
Feat/1479 Add attachments when writing emails

tammi-23:
Add attachments when writing emails https://github.com/opencloud-eu/web/issues/1479


packages/web-app-mail/src/components/MailAttachmentItem.vue
```diff
@@ -36,34 +55,72 @@ import {
...
+  return true
+})
+
 const download = async () => {
+  if (!canDownload.value) {
+    return
+  }
+  if (!hasBlobId(attachment)) {
+    return
+  }
```
JammingBen: You're already checking this via `canDownload`, so I think we can remove it here?

packages/web-app-mail/src/components/MailAttachmentList.vue
```diff
@@ -27,27 +27,76 @@
...
+}>()
+
+defineEmits<{
+  (e: 'remove', id: string): void
 }>()
-const collapsed = ref(attachments.length > 3)
+
+const { $gettext, $ngettext } = useGettext()
+
+const mode = props.mode ?? 'download'
```
JammingBen: I assume you want `download` to be the default if no `mode` is given? Please do this via destructuring `defineProps` like we usually do:

```
const { ..., mode = 'download' } = defineProps()
```

packages/web-app-mail/src/components/MailComposeAttachmentButton.vue
```diff
@@ -0,0 +1,104 @@
...
+  <div class="relative inline-flex">
+    <input ref="fileInput" class="hidden" type="file" multiple @change="onFileChange" />
+    <oc-button
+      :id="mobileToggleId"
+      type="button"
+      appearance="raw"
+      class="md:hidden flex h-9 w-9 items-center justify-center rounded-full border border-role-outline-variant bg-role-surface hover:bg-role-surface-variant transition"
+      :aria-label="$gettext('Add attachment')"
+      :title="$gettext('Add attachment')"
+    >
```
JammingBen, line 10: Things like `border`, `background-color`, `hover` and `justify-content` shouldn't be determined via classes but via the [oc-button props](https://docs.opencloud.eu/design-system/components/OcButton.html#properties). Maybe you can just remove the classes since you're using the defaults anyway?

The same goes for the other buttons in this component.

packages/web-app-mail/src/components/MailComposeAttachmentButton.vue
```diff
@@ -0,0 +1,104 @@
...
+const props = defineProps<{
+  modelValue?: MailComposeAttachment[]
+  accountId?: string
+}>()
+
+const emit = defineEmits<{
+  (e: 'update:modelValue', value: MailComposeAttachment[]): void
+}>()
+
+const fileInput = ref<HTMLInputElement | null>(null)
```
JammingBen: Generally we want to use `useTemplateRef` when using refs in templates. The type annotation then becomes obsolete in this case because it refers to a native HTML element (it would still be necessary on a custom Vue component though).

```suggestion
const fileInput = useTemplateRef('fileInput')
```

packages/web-app-mail/src/components/MailComposeAttachmentButton.vue
```diff
@@ -0,0 +1,104 @@
+<template>
+  <div class="relative inline-flex">
+    <input ref="fileInput" class="hidden" type="file" multiple @change="onFileChange" />
```
JammingBen, line 3: We have a FileInput component: https://docs.opencloud.eu/design-system/components/OcFileInput.html. Don't you want to use it here? Or does it make things more complicated?

packages/web-app-mail/src/components/MailComposeForm.vue
```diff
@@ -105,6 +116,8 @@ const { accounts } = storeToRefs(accountsStore)
 
 const selectedAccountIdQuery = useRouteQuery('accountId')
 
+const showFormattingToolbarValue = computed(() => props.showFormattingToolbar ?? false)
```
JammingBen: I'd also prefer a default value via prop destructuring here instead of adding a new `computed` property.

packages/web-app-mail/src/components/MailAttachmentList.vue
```diff
@@ -27,27 +27,76 @@
...
+    if (mode === 'download') collapsed.value = len > 3
+  }
+)
+
+type AttachmentWithBlobId = { blobId: string }
+type AttachmentWithId = { id: string }
+
+const hasBlobId = (a: Attachment): a is Attachment & AttachmentWithBlobId => {
+  return typeof (a as AttachmentWithBlobId).blobId === 'string'
+}
```
JammingBen: Does it make sense to move these types and the helper somewhere so they can be re-used? Because you seem to need them in the `MailAttachmentItem` component as well.

Issue #1397, opened 2025-10-20
Transform `useSideBar` composable into a store


JammingBen:
The `useSideBar` composable should be a pinia store to avoid registering multiple event bus listeners on the same events. Currently, toggling the sidebar calls the same events 4 times. We can also stop passing `isSideBarOpen` to components then.

Also, watchers on `isSideBarOpen` can be replaced with `sidebarStore.$onAction` then. This should be a bit more performant.

PR #1897, opened 2026-01-29
refactor: transform useSidebar composable to a pinia store

JammingBen:
Replaces the `useSideBar` composable with a pinia store. This results in some changes across the code base:

* we don't need to pass the sidebar states around since the global store is accessible from everywhere
* we don't need to rely on events to toggle the sidebar, hence those have been deprecated in favor of store actions

closes https://github.com/opencloud-eu/web/issues/1397


packages/web-pkg/src/composables/piniaStores/sideBar.ts
```diff
@@ -46,43 +50,42 @@ export const useSideBar = (options?: SideBarOptions) => {
...
+  const openSideBarPanel = (panelName: string) => {
+    isSideBarOpen.value = true
+    sideBarActivePanel.value = panelName
+    focusSidebar()
+  }
+  const setActiveSideBarPanel = (panelName: string) => {
+    sideBarActivePanel.value = panelName
+  }
+
+  eventBus.subscribe(SideBarEventTopics.toggle, toggleSideBar)
```
kulmann, line 77: should we be really annoying and wrap the functions into another function which prints a deprecation message to debug output? 🙈

JammingBen: I don't think we really need to, no? When using the event with `SideBarEventTopics` like you usually do, you immediately see in your IDE that this is deprecated.

PR #1904, opened 2026-01-30
refactor: move getIndicators to a composable

JammingBen:
This way we don't need to pass all the data to `getIndicators()`. Also, it allows us to use the sidebar store actions instead of using the now deprecated events.

The old `getIndicators` method has been deprecated for compatibility reasons. It now calls the new composable internally.

Follow-up for https://github.com/opencloud-eu/web/pull/1897


packages/web-pkg/src/composables/resources/useResourceIndicators.ts
```diff
@@ -0,0 +1,246 @@
...
+  IncomingShareResource,
+  isIncomingShareResource,
+  isPersonalSpaceResource,
+  isProjectSpaceResource,
+  Resource,
+  ShareTypes,
+  SpaceResource
+} from '@opencloud-eu/web-client'
+import { useInterceptModifierClick } from '../keyboardActions'
+import { useResourcesStore, useSideBar, useUserStore } from '../piniaStores'
```
kulmann, line 12: didn't notice it before, but why is `useSideBar` not named `useSideBarStore`? 🤔

kulmann (approved):
works 🥳

packages/web-pkg/src/composables/resources/useResourceIndicators.ts
JammingBen: Mainly because of compatibility reasons (would be breaking otherwise). But I also find that `...Store` in this context is not really needed. We also don't have it for e.g. `useModals` and `useMessages`.

Issue #1772, opened 2025-12-18
Preview app: Reasonable preloading


kulmann:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

> * As a user with many photos in one folder, I want the preview app to cause as little load on the server as possible, so that I don't DOS my server.

### User Value

Fast UI, less load on the server.

### Acceptance Criteria

- ~~after loading the current image in high resolution, only pre-load the 2 previous and the 2 next image (photo roll makes precise navigation possible, preloading many images causes too much load on the server)~~
- if the user cycles through images in a very fast way, requests for other images should be cancelled again
- if the user leaves the preview app, the active request for the current image (if it was not finished, yet) should be cancelled. especially important for large images on mobile bandwidth.
- respect sorting (see https://github.com/opencloud-eu/web/issues/1771)

PR #1921, opened 2026-02-02
feat: adjust preview app loading

AlexAndBear:
## Description

- Refactored component to  `<script setup>`
- Merged `filteredFiles` and `cachedFiles` props into a single prop `mediaFiles`
 - Cancel ongoing image loading when quickly skipping to another image before loading completes 
 - Decoupled photo roll previews

## Related Issue

- Fixes https://github.com/opencloud-eu/web/issues/1772

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/tests/e2e/support/objects/app-files/resource/actions.ts
+++ b/tests/e2e/support/objects/app-files/resource/actions.ts
@@ -1825,15 +1825,7 @@ export const openFileInViewer = async (args: openFileInViewerArgs): Promise<void
       ])
       break
     case 'mediaviewer': {
-      await Promise.all([
-        page.waitForResponse(
-          (resp) =>
-            resp.url().includes('preview') &&
-            resp.status() === 200 &&
-            resp.request().method() === 'GET'
-        ),
-        page.locator(util.format(resourceNameSelector, name)).click()
-      ])
+      await page.locator(util.format(resourceNameSelector, name)).click()
       const extension = name.split('.').pop()
       switch (extension) {
         case 'mp3':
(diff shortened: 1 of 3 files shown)


tests/e2e/support/objects/app-files/resource/actions.ts
AlexAndBear, line 1828: We can’t rely on requesting a preview at this point. When files are opened from the tiles view, all images in the photo roll are already loaded. If, for example, an MP4 file is opened, it doesn’t provide a preview flag via PROPFIND, so we don’t make any preview-endpoint requests here.

packages/web-app-preview/src/App.vue
```diff
@@ -82,47 +81,39 @@
...
-  ref,
-  unref,
-  PropType,
-  nextTick,
-  watch,
-  Ref,
-  onBeforeUnmount,
-  onMounted
-} from 'vue'
+export const appId = 'preview'
```
kulmann, line 84: In the notes app I've placed the appId in a completely separate file because it felt cleaner that way... see https://github.com/opencloud-eu/web-extensions/blob/0fa6d00df1c9c382b890077099d2235859561019/packages/web-app-notes/src/util.ts#L1

I don't see a reason why the App.vue would be a good place for the const. 🙈 I know that you didn't move it here, but could you add the extra refactoring and move it to a dedicated ts file? 🫶

packages/web-app-preview/src/App.vue
```diff
@@ -133,362 +124,337 @@ import { mimeTypes } from './mimeTypes'
...
+  resetImage
+} = useImageControls()
+const { isFullScreenModeActivated, toggleFullScreenMode } = useFullScreenMode()
+
+const activeIndex = ref<number>()
+const mediaFiles = ref<MediaFile[]>([])
+const folderLoaded = ref(false)
+const isAutoPlayEnabled = ref(true)
+const photoRollEnabled = ref(true)
+const preview = ref<HTMLElement>()
```
kulmann, line 177: You could/should make use of `useTemplateRef` here 😇

PR #1927, opened 2026-02-03
feat: only request photo roll thumbnails, when item is in viewport

### LINKED ISSUE #1770: Preview app: improve client side performance of photo roll
### Initial Checklist

### User Story

> * As a user I want the photo roll of the preview app to have a good performance so that it doesn't slow down my browser / device.

### User Value

Fast UI

### Acceptance Criteria

- for the images in the photo roll of the preview app, introduce a "loader"-div (skeleton loader) like we do it in the files app, which gets replaced with the actual image when it scrolls into the viewport
- request a smaller resolution for the images in the photo roll, so that we don't cause heavy memory load in the browser (use the same resolution as for the tiles view... if the user comes from the tiles view, best case is that the images are already cached server side and fetching them is fast then)

AlexAndBear:
## Related Issue

- Fixes #1770 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/web-app-preview/src/components/PhotoRollItem.vue
+++ b/packages/web-app-preview/src/components/PhotoRollItem.vue
@@ -1,0 +1,84 @@
+<template>
+  <div
+    ref="photoRollItem"
+    class="flex flex-col items-center p-4 mb-1 photo-roll-item"
+    :class="{ 'bg-role-surface rounded-md': isActive }"
+  >
+    <oc-button
+      appearance="raw"
+      class="flex flex-col w-full"
+      :aria-label="item.name"
+      no-hover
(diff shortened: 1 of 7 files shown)


packages/web-app-preview/src/components/PhotoRollItem.vue
```diff
@@ -0,0 +1,84 @@
...
+    mimeType: props.item.mimeType,
+    isFolder: false,
+    isFile: true,
+    type: 'file'
+  } as Resource
+})
+
+onMounted(() => {
+  if (!itemRef.value) return
+
```
AlexAndBear, line 67:
```suggestion
```

Issue #1733, opened 2025-12-10
[QA] Flaky web e2e tests


saw-jan:
## 1. app-provider-onlyOffice/officeSuites.feature:17
<details>
<summary>content check failure - file lock issue [FIXED]</summary>

Builds:
- https://ci.opencloud.rocks/repos/6/pipeline/330/197

Tracings:
```bash
npx playwright show-trace https://s3.ci.opencloud.eu/public/web/tracing/web/330/create-a-microsoft-word-file-with-only-office-alice-2025-12-10-12-17-18.zip

npx playwright show-trace https://s3.ci.opencloud.eu/public/web/tracing/web/330/create-a-microsoft-word-file-with-only-office-anonymous-2025-12-10-12-17-44.zip
```
```feature
Scenario: create a Microsoft Word file with OnlyOffice
  ...
  Then "Anonymous" should see the content "Microsoft Word Content" in editor "OnlyOffice"
    ✖ failed
      Error: Failed to read non-empty clipboard content after retries
          at tryCopyClipboard (file:///woodpecker/src/github.com/opencloud-eu/web/web/tests/e2e/support/objects/app-files/resource/actions.ts:376:11)
          at async Module.openAndGetContentOfDocument (file:///woodpecker/src/github.com/opencloud-eu/web/web/tests/e2e/support/objects/app-files/resource/actions.ts:362:12)
          at async Public.getContentOfOpenDocumentOrMicrosoftWordDocument (file:///woodpecker/src/github.com/opencloud-eu/web/web/tests/e2e/support/objects/app-files/page/public.ts:61:16)
          at async World.<anonymous> (file:///woodpecker/src/github.com/opencloud-eu/web/web/tests/e2e/cucumber/steps/ui/public.ts:60:31)
```
</details>

<details>
<summary>update share failure</summary>

Builds:
- https://ci.opencloud.rocks/repos/6/pipeline/330/197

```feature
Scenario: create a Microsoft Word file with OnlyOffice
  ...
  When "Alice" updates following sharee role
      │ resource           │ recipient │ type │ role     │ resourceType │
      │ MicrosoftWord.docx │ Brian     │ user │ Can edit │ file         │
    ✖ failed
      page.waitForResponse: Timeout 30000ms exceeded while waiting for event "response"
          at Module.changeShareeRole (/woodpecker/src/github.com/opencloud-eu/web/web/tests/e2e/support/objects/app-files/share/actions.ts:117:18)
[...]

PR #1937, opened 2026-02-05
[full-ci]test: harden flaky web e2e tests

saw-jan:
## Description
- fix search result assertion step
- separate test and action timeouts
- wait files to be unlocked and out of processing before any performing actions

## Related Issue
- Part of #1733

## How Has This Been Tested?
- test environment:

## Types of changes
- [ ] Bugfix

### DIFF
--- a/tests/e2e/support/objects/app-files/fileEvents.ts
+++ b/tests/e2e/support/objects/app-files/fileEvents.ts
@@ -1,17 +1,11 @@
-import { Page, Locator } from '@playwright/test'
+import { Page, Locator, expect } from '@playwright/test'
 import util from 'util'
 import { config } from '../../../config'
 
-const resourceNameSelector =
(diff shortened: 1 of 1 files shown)


tests/e2e/support/objects/app-files/fileEvents.ts
```diff
@@ -0,0 +1,68 @@
...
+    console.info(`[INFO] ${message}...`)
+    await page.waitForTimeout(config.minTimeout * 1000)
+    await page.reload()
+
+    if (fn) {
+      await fn()
+    }
+
+    elapsedTime = Date.now() - startTime
+  }
```
v-scharf: can we try to use `await expect(locator, message).toBeHidden({ timeout })` instead?
I would avoid to use `locator.isVisible()`

saw-jan: I have used `await expect(locator, message).toBeHidden({ timeout })` now and removed the reload logic. we will report the failures if lock and processing icons are not removed even after the timeout.

PR #1997, opened 2026-02-18
feat: create or upload floating action button

### LINKED ISSUE #1960: Make "Create or Upload" the FAB of the files app
### Initial Checklist

### User Story

> * As a user, I want to have a fixed location in the design where the main interaction for creating things happens so that I know where to look without thinking.

### User Value

UX

### Acceptance Criteria

- "Create or Upload" button is the floating action button (FAB) of the files app, meaning: not part of the header area above the file list anymore but sits in the "floating action button for desktop" location, which is at the top of the left sidebar (see e.g. mail app)
- Desktop resolution:
  - button uses full width of the sidebar
  - dropdown has the same width as the button and aligns exactly below the triggering button (with a little bit of margin)
  - if write operations are not possible, disable the FAB
- using a mobile resolution positions the FAB where it needs to be on mobile (bottom right corner)
  - clicking the mobile FAB opens the menu in a bottom drawer
  - if write operations are not possible, hide the FAB entirely on mobile

### Technical detail

Additional detail we discussed in the last refinement

- the runtime should provide a new extension for that can register a component in the left sidebar (that gets placed on the bottom right on mobile, preferably without duplicating the extension point or having 2 of them - maybe teleport needed if CSS only doesn't work?)
- the files app should mount a component onto that extension point
  - the component gets displayed on every view to avoid jumpyness
  - the component decides on the space and the route if it's disabled (the `CreateAndUpload` component already does that, maybe just use this one and mount it onto the extension point?)

Image

Image

AlexAndBear:
## Related Issue

- Fixes #1960 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/tests/woodpecker/app-registry.yaml
+++ b/tests/woodpecker/app-registry.yaml
@@ -40,8 +40,8 @@ app_registry:
       name: Microsoft Excel
       description: Microsoft Excel document
       icon: ''
-      default_app: OnlyOffice
-      allow_creation: Collabora
+      default_app: Collabora
+      allow_creation: true
     - mime_type: application/vnd.openxmlformats-officedocument.presentationml.presentation
       extension: pptx
       name: Microsoft PowerPoint


packages/web-app-files/src/extensions.ts
```diff
@@ -38,6 +47,20 @@ export const extensions = (appInfo: ApplicationInformation) => {
       type: 'search',
       searchProvider: new SDKSearch(capabilityStore, router, searchFunction, configStore)
     },
+    {
+      id: 'com.github.opencloud-eu.web.files.floating-action-button',
+      extensionPointIds: ['global.floating-action-button'],
+      type: 'floatingActionButton',
+      icon: 'add',
+      label: () => $gettext('Create or Upload'),
```
JammingBen: I can't remember, what did we decide on with the label? Because with a German label like "Erstellen und hochladen" this would be 2 lines, which looks really bad. I'd go with a shorter label. @kulmann Opinions?

packages/web-app-files/tests/unit/components/CreateOrUploadMenu.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    mock<FileAction>({ label: () => 'Mark-down file', ext: 'md', isExternal: false })
+  ],
+  extensionActions = [],
+  createFolderHandler = vi.fn(),
+  createShortcutHandler = vi.fn()
+}: {
+  currentFolder?: Resource
+  areFileExtensionsShown?: boolean
+  createActions?: FileAction[]
+  extensionActions?: any[]
```
JammingBen:
```suggestion
  extensionActions?: Extension[]
```

packages/web-app-files/tests/unit/components/CreateOrUploadMenu.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+  const plugins = defaultPlugins({
+    piniaOptions: {
+      resourcesStore: { areFileExtensionsShown, currentFolder }
+    }
+  })
+
+  const { requestExtensions } = useExtensionRegistry()
+  vi.mocked(requestExtensions).mockReturnValue(
+    extensionActions.map((action) => ({ action })) as any
+  )
```
JammingBen:
```suggestion
  vi.mocked(requestExtensions).mockReturnValue(extensionActions)
```

packages/web-runtime/src/App.vue
```diff
@@ -10,6 +10,26 @@
...
+          :handler="floatingActionButton.handler"
+        />
+        <component
+          :is="floatingActionButton.dropComponent"
+          v-if="floatingActionButton.dropComponent"
+          :toggle="`#mobile-floating-action-button-${floatingActionButton.id.replace(/\./g, '-')}`"
+        />
+      </template>
+    </div>
+  </template>
```
JammingBen: This looks like quite a bit of duplication. Maybe introduce a dedicated component that requests the FAB extensions and then renders the template (button or FAB depending on `isMobile`)?

JammingBen (commented):
Nice, already feels really good IMO! A few quirks I noticed, some may be for a follow-up.

- The batch actions when selecting a file have a bit of margin to the left, that looks weird.
- Didn't we say we want the action for creating spaces to be in this menu on the spaces overview page? I can't remember though 🤔 
- In views without a FAB we should make sure to align the first nav item with the headline/breadcrumb of the content area (or at least align the top and the left margin within the sidebar): 

Image

- When navigating into folders, the button blinks for a short time because it's disabled initially. While not dramatically bad, it's a bit annoying. But I doubt we can fix that easily. Maybe with some hacky debounce.

At last, but that's something for PM to decide, we could think about not disabling the button and just defaulting to creating in the personal space IMO. That's how other clouds do it. The disabling feels a bit weird to me... on the other hand though it's really transparent to the user we they are creating files, which wouldn't be the case if we implicitly default to the personal space.

packages/web-app-files/src/extensions.ts
AlexAndBear: I suggested New but I don't know if it was accepted, the ticket doesn't say anything about it.

Issue #1929, opened 2026-02-04
Replace tippy.js with Floating UI


JammingBen:
[tippy.js](https://github.com/atomiks/tippyjs) has been stale for about 5 years now, the repository got archived 2 years ago. The original maintainer pushes to [floating-ui](https://github.com/floating-ui/floating-ui) which seems to be the "new" standard and includes a dedicated Vue implementation.

I think it's worth the jump to ensure we stay up to date with a properly maintained lib.

PR #1998, opened 2026-02-18
refactor!: floating UI

JammingBen:
Replaces tippy.js with the more modern Floating UI. This results in some **breaking changes when using the `OcDrop` component**.

### `OcDrop` breaking changes

- The property `popperOptions` has been removed
- The property `target` has been removed
- The property `offset` is now supposed to be given as a single digit number and controls the vertical offset
- The exposed method `tippy` has been removed
- The component now expects a `toggle` even in `manual` mode.

Floating UI in general is more lightweight and much more similar to a JS native implementation. It basically offers you a set of helpers to position your floating element correctly, but you have to take care of things like event handling, showing/hiding the floating element etc.

closes https://github.com/opencloud-eu/web/issues/1929

### DIFF
--- a/tests/e2e/support/objects/app-files/resource/actions.ts
+++ b/tests/e2e/support/objects/app-files/resource/actions.ts
@@ -377,7 +377,7 @@ export const createNewFolder = async ({
 
 export const createNewFileOrFolder = async (args: createResourceArgs): Promise<void> => {
   const { page, name, type, content } = args
-  await page.locator(addNewResourceButton).waitFor({ state: 'visible' })
+  await expect(page.locator(addNewResourceButton)).toBeVisible()
   await expect(page.locator(addNewResourceButton)).toBeEnabled()
   await page.locator(addNewResourceButton).click()
   switch (type) {


packages/web-app-admin-settings/src/components/Groups/GroupsList.vue
```diff
@@ -94,7 +94,7 @@
           <oc-icon name="pencil" fill-type="line" />
         </oc-button>
         <context-menu-quick-action
-          ref="contextMenuButtonRef"
+          :ref="(el: any) => (contextMenuDrops[item.id] = el?.drop)"
```
JammingBen: `any` cast can be removed when we switch to script setup. But I don't want to bloat this already big PR even more.

packages/web-app-admin-settings/tests/unit/components/Groups/GroupsList.spec.ts
```diff
@@ -63,20 +62,6 @@ describe('GroupsList', () => {
...
-    await wrapper.find(`[data-item-id="${groups[0].id}"]`).trigger('contextmenu')
-    expect(spyDisplayPositionedDropdown).toHaveBeenCalledTimes(1)
-  })
-  it('should show the context menu on context menu button click', async () => {
-    const groups = getGroupMocks()
-    const spyDisplayPositionedDropdown = vi.mocked(displayPositionedDropdown)
-    const { wrapper } = getWrapper({ mountType: mount, groups })
-    await wrapper.find('.groups-table-btn-action-dropdown').trigger('click')
-    expect(spyDisplayPositionedDropdown).toHaveBeenCalledTimes(1)
-  })
```
JammingBen, line 79: We can't unit test that the drop is being called because the `ref` to the drop element doesn't exist.

packages/web-app-files/src/components/SideBar/Shares/Collaborators/RoleDropdown.vue
```diff
@@ -223,75 +210,6 @@ export default defineComponent({
...
-  beforeUnmount() {
-    window.removeEventListener('keydown', this.cycleRoles)
-  },
-
-  mounted() {
-    window.addEventListener('keydown', this.cycleRoles)
-  },
-
-  methods: {
-    cycleRoles(event: KeyboardEvent) {
```
JammingBen, line 237: This was broken for the link role dropdown and other drop menus don't have this functionality, hence I decided to remove it.

packages/web-pkg/tests/unit/components/ContextActions/ContextActionMenu.spec.ts
```diff
@@ -56,7 +56,6 @@ describe('ContextActionMenu component', () => {
     expect(wrapper.html()).toMatchSnapshot()
 
     expect(wrapper.findAll('#oc-files-context-menu > ul').length).toEqual(menuSections.length)
-    expect(wrapper.findAll('.oc-files-context-action-drop').length).toEqual(3)
```
JammingBen, line 59: Drop menus are not rendered initially anymore.

packages/design-system/src/components/OcDrop/OcDrop.vue
```diff
@@ -44,48 +62,25 @@ export interface Props {
...
    * @default 'medium'
    */
   paddingSize?: SizeType | 'remove'
   /**
-   * @docs The popper options of the drop. Please refer to the component source for more information.
-   */
-  popperOptions?: TippyProps['popperOptions']
-  /**
-   * @docs The position of the drop.
+   * @docs The position of the drop. Check the floating-ui document for more details on the type.
```
kulmann:
```suggestion
   * @docs The position of the drop. Check the floating-ui documentation for more details on the type.
```

packages/design-system/src/components/OcDrop/OcDrop.vue
```diff
@@ -11,23 +11,41 @@
...
+        v-if="isOpen"
+        :id="dropId"
+        ref="drop"
+        class="oc-drop shadow-md/20 rounded-sm bg-role-surface"
+        @click="onClick"
+      >
+        <oc-card v-if="$slots.default" :body-class="[getTailwindPaddingClass(paddingSize)]">
+          <slot />
+        </oc-card>
+        <slot name="special" />
```
kulmann: Shouldn't this have a `v-else` when it was a `v-else-if` before? You've moved the `v-if` condition to the wrapping div, but if the logic for when to show the `special` slot didn't change, it should have a `v-else`.

packages/design-system/src/components/OcDrop/OcDrop.vue
```diff
@@ -138,177 +130,264 @@ const {
...
+
+  isOpen.value = true
+  await nextTick()
+  if (!anchorEl) {
+    console.warn('OcDrop cannot be opened: anchor element not found')
+    return
+  }
+
+  // fixes a timing issue with the rendering of the drop
+  await new Promise((resolve) => setTimeout(resolve, 0))
```
kulmann: wtf, how? 🙈

packages/design-system/src/components/OcDrop/OcDrop.vue
JammingBen: Yap, good catch!

packages/design-system/src/components/OcDrop/OcDrop.vue
JammingBen: I really don't now 🙈 I tried debugging and understanding, but I wasn't able to get behind it.

kulmann: what about something like awaiting the next tick then? 🤔

JammingBen: I tried a dozen of them, didn't work :(

JammingBen: As discussed: improved slightly at least via `requestAnimationFrame`.

Issue #1936, opened 2026-02-05
Replace `portal-vue` plugin with built-in teleport


JammingBen:
Vue3 has introduced a built-in [teleport](https://vuejs.org/guide/built-ins/teleport) feature that makes the `portal-vue` dependency obsolete. We should utilize that and drop the dependency.

Some of the existing usages of `portal-target` can/should be replaced with proper extension points.

PR #2007, opened 2026-02-19
refactor!: mobile nav to web-pkg

JammingBen:
### Breaking change

Apps must now directly use the `MobileNav` component from `web-pkg` instead of using a portal.

I needed to move a few some helpers from the runtime to web-pkg so they can be used globally.

refs https://github.com/opencloud-eu/web/issues/1936


packages/web-runtime/tests/unit/components/SidebarNav/SidebarNav.spec.ts
```diff
@@ -21,16 +15,6 @@ describe('OcSidebarNav', () => {
...
-  it('initially sets the highlighter to the active nav item', async () => {
-    const { wrapper } = getWrapper()
-    await wrapper.vm.$nextTick()
-    expect(wrapper.vm.highlighterAttrs).toEqual({
-      style: {
-        transform: 'translateY(0px)',
-        'transition-duration': '0.2s'
-      }
-    })
-  })
```
JammingBen, line 33: The highlighter will be removed anyways with the redesign, no need to spend time on fixing this test now.

PR #2015, opened 2026-02-20
refactor!: port vue-portal to teleport or extension system

JammingBen:
### Breaking changes for `portal` und `portal-target` usages

- Global `$wormhole` on vue instance has been removed
- `PortalTarget` component has been removed
- `app.files.sidebar.file.details.table` portal target has been removed. Register an extension on the `app.files.sidebar.file-details.table` extension point instead.
- `app.files.sidebar.space.details.table` portal target has been removed. Register an extension to the `app.files.sidebar.space-details.table` extension point instead.
- `app.files.sidebar.sharing.shared-with.top` portal target has been removed. Register an extension on the `app.files.sidebar.shares-panel.shared-with.top` extension point instead.
- `app.files.sidebar.sharing.shared-with.bottom` portal target has been removed. Register an extension on the `app.files.sidebar.shares-panel.shared-with.bottom` extension point instead.
- `app.runtime.header.left` portal target has been removed. Register an extension on the `app.runtime.header.left` extension point instead.
- `app.runtime.header.right` portal target has been removed. Register an extension on the `app.runtime.header.right` extension point instead.

Portals intended for internal use (e.g. the bottom drawer), I didn't replace with new extension points but with teleports. Also, extensions of type `CustomComponentExtension` can now have properties via the new `componentProps` property.

closes https://github.com/opencloud-eu/web/issues/1936

### DIFF
--- a/tests/e2e/support/objects/app-files/resource/actions.ts
+++ b/tests/e2e/support/objects/app-files/resource/actions.ts
@@ -377,6 +377,8 @@ export const createNewFolder = async ({
 
 export const createNewFileOrFolder = async (args: createResourceArgs): Promise<void> => {
   const { page, name, type, content } = args
+  await page.locator(addNewResourceButton).waitFor({ state: 'visible' })
+  await expect(page.locator(addNewResourceButton)).toBeEnabled()
   await page.locator(addNewResourceButton).click()
   switch (type) {
     case 'folder': {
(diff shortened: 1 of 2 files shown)


tests/e2e/support/objects/app-files/resource/actions.ts
```diff
@@ -377,6 +377,8 @@ export const createNewFolder = async ({
 
 export const createNewFileOrFolder = async (args: createResourceArgs): Promise<void> => {
   const { page, name, type, content } = args
+  await page.locator(addNewResourceButton).waitFor({ state: 'visible' })
```
v-scharf, line 380: It would be better to use `expect(page.locator(addNewResourceButton)).toBeVisible()`

JammingBen: Thanks for the hint, I'll do this in https://github.com/opencloud-eu/web/pull/1998 because I need to rebase it anyways and then we can merge this now to unblock other PRs.

Issue #2019, opened 2026-02-20
"New" button flickers on every navigation


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

The new "New" button flickers on every navigation in spaces where the user has write access.

### Reproduction Steps

- open your personal space
- navigate into a folder

### Expected Outcome

The "New" button in the left nav doesn't flicker on navigation.

### Actual Outcome

It flickers.

PR #2026, opened 2026-02-23
fix: prevent 'New'-button flicker when navigating

JammingBen:
Also makes sure to render only one FAB (as we discussed recently in chat), since it doesn't make sense to have multiple FABs displaying.

fixes https://github.com/opencloud-eu/web/issues/2019

### DIFF
--- a/packages/web-runtime/src/components/AppFloatingActionButton.vue
+++ b/packages/web-runtime/src/components/AppFloatingActionButton.vue
@@ -1,60 +1,77 @@
 <template>
-  <div
-    v-for="floatingActionButton in floatingActionButtonExtensions"
-    :key="floatingActionButton.id"
-  >
-    <template v-if="floatingActionButton.isActive()">
-      <div v-if="!isMobile" class="py-2 px-2">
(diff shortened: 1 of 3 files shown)


packages/web-runtime/src/components/AppFloatingActionButton.vue
```diff
@@ -1,60 +1,77 @@
...
 
-const floatingActionButtonExtensions = computed(() => {
+const isDisabled = ref(false)
+
+const floatingActionButton = computed(() => {
   return requestExtensions<FloatingActionButtonExtension>({
     id: 'global.floating-action-button',
     extensionType: 'floatingActionButton'
-  })
+  })[0]
```
AlexAndBear: This doesn't work, with https://github.com/opencloud-eu/web/pull/2025 we now have a FAB that registers via the admin settings app. You can filter for isActive and use the first element

JammingBen: Thanks for the hint, I adjusted it!

Issue #2041, opened 2026-02-27
User avatars not showing in "Shared By" filter


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

User avatars are not showing in the "Shared By" filter on the "Shared with me" page.

### Reproduction Steps

- as user dennis, upload a user avatar
- as user dennis, create a file and share it with admin
- as user admin, go to the "Shared with me" page and open the "Shared by" filter

### Expected Outcome

The avatar is showing in the filter.

### Actual Outcome

The avatar is now showing, it just displays the user's initials.

<img width="364" height="188" alt="Image" src="https://github.com/user-attachments/assets/cdf1335b-18c5-4537-9b1e-f6a1730f68a8" />

PR #2042, opened 2026-02-27
fix: broken AvatarImage usages

JammingBen:
The `AvatarImage` component is outdated and doesn't work with the recent user avatar implementation. Replaces its usages with `oc-avatar` and `user-avatar` and remove the component completely.

Also refactors the components `SharedWithMe` and `GroupSelect` to script setup (see separate commits).

fixes https://github.com/opencloud-eu/web/issues/2041


AlexAndBear (approved):
Thank you 👍

PR #2058, opened 2026-03-03
feat: allow navigation back to last folder, when breadcrumb is truncated

### LINKED ISSUE #2038: Change in breadcrumb navigation
### Initial Checklist

- [x] I understand this is a feature request and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)

### Idea

I would like to be able to navigate just one step back in the folder-tree.
At some point that is not possible anymore via the breadcrumb navigation.
Image

I would suggest, that the breadcrumb of the direct previous folder is always shown.
Or at least the three dots should be a link to the previous folder.

AlexAndBear:
## Related Issue

- Fixes #2038 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/design-system/src/components/OcBreadcrumb/__snapshots__/OcBreadcrumb.spec.ts.snap
+++ b/packages/design-system/src/components/OcBreadcrumb/__snapshots__/OcBreadcrumb.spec.ts.snap
@@ -8,7 +8,7 @@ exports[`OcBreadcrumb > displays all items 1`] = `
       <oc-icon-stub data-v-5914b352="" accessiblelabel="" color="var(--oc-role-on-surface)" filltype="line" name="arrow-right-s" size="medium" type="span" class="mx-1 align-sub"></oc-icon-stub>
       <!--v-if-->
     </li>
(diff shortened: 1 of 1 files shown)


packages/design-system/src/components/OcBreadcrumb/OcBreadcrumb.vue
```diff
@@ -31,7 +31,7 @@
         @drop="dropItemEvent(item, index)"
       >
         <router-link
-          v-if="item.to && !item.isTruncationPlaceholder"
+          v-if="item.to"
```
kulmann, line 33: do you think it makes sense to set the actual folder name as aria-label, to overrule the `...` visible label of the item? screen reader doesn't have the size limitation, on the other hand the screen reader experience with the truncated items is not good anyway.

AlexAndBear: I added the aria label

kulmann:
> not feasible, UX gets horrible

could you elaborate on "horrible"?

PR #2090, opened 2026-03-05
feat: ease use of floating action button extension

AlexAndBear:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/web-app-admin-settings/src/index.ts
+++ b/packages/web-app-admin-settings/src/index.ts
@@ -24,6 +24,7 @@ import {
   useUserActionsCreateUser
 } from './composables'
 import { APPID } from './appid'
+import { extensionPoints } from './extensionPoints'
 
 export const routes: ClassicApplicationScript['routes'] = ({ $ability, $gettext }) => [
   {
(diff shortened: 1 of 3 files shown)


packages/web-app-files/src/extensionPoints.ts
```diff
@@ -39,11 +38,6 @@ export const fileSideBarExtensionPoint: ExtensionPoint<SidebarPanelExtension<any
   multiple: true
 }
 
-export const floatingActionButtonExtension: ExtensionPoint<FloatingActionButtonExtension> = {
-  id: 'global.floating-action-button',
-  extensionType: 'floatingActionButton'
-}
-
```
JammingBen, line 46: I think we should keep this but with id `app.files.floating-action-button`.

packages/web-app-admin-settings/src/index.ts
JammingBen, line 1: Similar to the files app, the app should define an extension point with id `app.admin-settings.floating-action-button` which can then be registered via `defineWebApplication`'s setup return.

Issue #2097, opened 2026-03-06
Drop sub menus can't be reached via keyboard navigation


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

Drop sub menus (e.g. "Open with") can't be reached via keyboard navigation.

### Reproduction Steps

- create a folder
- right click the folder
- try opening the "Open with" sub menu via keyboard navigation

### Expected Outcome

The sub menu can be opened via keyboard navigation. Ideally, the drop menu can be navigated via arrow keys.

### Actual Outcome

There is no way to reach the sub drop menu via keyboard navigation.

Issue #2081, opened 2026-03-04
Items per page don't overflow the drop menu


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

The Items per page don't overflow the drop menu.

### Reproduction Steps

- open the view options for a file list
- open the select for the "Items per page"

### Expected Outcome

The select options expand below, exceeding the drop menu.

### Actual Outcome

The select options get cropped, you need to scroll within the drop menu.

<img width="271" height="160" alt="Image" src="https://github.com/user-attachments/assets/8c050e27-5d60-49b5-8bf5-3c613c2c9782" />

PR #2100, opened 2026-03-06
fix: oc-drop a11y and overflow issues

JammingBen:
Fixes 2 **overflow issues** with the `OcDrop` component:

- sub menus (e.g. "Open with...") now open to the left or right instead of to the bottom so other elements of the drop don't get covered
- the expanded select input in the view options drop now correctly overflows the drop so all items are accessible

Fixes **a11y** issues with the `OcDrop` component:

- drop menus are now properly navigable via arrow keys instead of tabs (with some exceptions where the drops are not classic menus, e.g. contextual helpers)
- drop menus and their items now have proper `role` attributes

fixes https://github.com/opencloud-eu/web/issues/2097
fixes https://github.com/opencloud-eu/web/issues/2081

### DIFF
--- a/packages/web-app-files/src/components/CreateOrUploadMenu.vue
+++ b/packages/web-app-files/src/components/CreateOrUploadMenu.vue
@@ -7,7 +7,6 @@
     mode="click"
     close-on-click
     padding-size="small"
-    @show-drop="showDrop"
   >
     <oc-list
       id="create-list"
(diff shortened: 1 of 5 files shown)


packages/web-app-files/src/components/CreateOrUploadMenu.vue
```diff
@@ -187,12 +185,6 @@ const canUpload = computed(() => {
   return unref(currentFolder)?.canUpload({ user: userStore.user })
 })
 
-const actionKeySuffix = ref(uuidV4())
-const showDrop = () => {
-  // force actions to be re-rendered when the drop is being opened
-  actionKeySuffix.value = uuidV4()
-}
-
```
JammingBen, line 195: Removed because by now, drops automatically get re-rendered when being opened.

Issue #1959, opened 2026-02-10
Mail: Add an action bar for email actions


tammi-23:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

As a User, I want to have shortcut actions in one place

### User Value

Always have the same place for (email) actions

### Acceptance Criteria

- If an email is displayed in desktop view a extra bar is displayed above the email, if any action is available
- first iteration: if the mail is a draft, edit draft is displayed with an icon

PR #2125, opened 2026-03-11
Feat/1959

tammi-23:
Add https://github.com/opencloud-eu/web/issues/1959
Add an action bar for email actions

### DIFF
--- a/packages/web-app-mail/src/components/MailList.vue
+++ b/packages/web-app-mail/src/components/MailList.vue
@@ -69,14 +69,12 @@ import { useMailboxesStore } from '../composables/piniaStores/mailboxes'
 import { storeToRefs } from 'pinia'
 import { useLoadMail } from '../composables/useLoadMail'
 import { unref } from 'vue'
-import { useMailCompose } from '../composables/useMailCompose'
 
 const mailsStore = useMailsStore()
 const mailboxesStore = useMailboxesStore()
 const accountsStore = useGroupwareAccountsStore()
 const { loadMail } = useLoadMail()
 const { isLoading } = useLoadMails()
-const { openNewCompose } = useMailCompose()
 
 const { currentAccount } = storeToRefs(accountsStore)
 const { currentMail, mails } = storeToRefs(mailsStore)


packages/web-app-mail/src/components/MailList.vue
```diff
@@ -82,6 +96,12 @@ const { updateMailField, setCurrentMail } = mailsStore
 const { currentMailbox } = storeToRefs(mailboxesStore)
 const { setCurrentMailbox } = mailboxesStore
 
+const { isOpen: showCompose, draftMail, openNewCompose, closeCompose } = useMailCompose()
+
+defineExpose({
+  openCompose: openNewCompose
+})
```
JammingBen: What's the reason you're exposing this? 🤔

packages/web-app-mail/src/components/MailWidget.vue
```diff
@@ -142,16 +142,22 @@ import { useAutoSaveDraft } from '../composables/useAutoSaveDraft'
...
 
 const { $gettext } = useGettext()
 const { dispatchModal } = useModals()
+const appliedDraftId = ref<string | null>(null)
 
 const SAVED_HINT_DURATION_MS = 2000
 const AUTO_SAVE_INTERVAL_MS = 120000 // 2(min) * 60 * 1000
 
+const props = defineProps<{
+  modelValue?: boolean
```
JammingBen: I removed the `modelValue` the other day because it just adds unnecessary complexity. The visibility of the component is controlled from the outside. So inside this component, you can always assume it's visible, because otherwise it wouldn't exist.

Hence you can remove `modelValue` and `isOpen`.

packages/web-app-mail/src/composables/useMailCompose.ts
JammingBen, line 1: There is another use of the `MailWidget` component in `Inbox.vue`. I guess it makes sense to also use of this composable there.

Furthermore, you could remove the `app.mail.show-compose-mail` event handling completely and directly use this composable's `openNewCompose` method in `packages/web-app-mail/src/extensions.ts` (instead of `eventBus.publish('app.mail.show-compose-mail')`).

PR #2147, opened 2026-03-13
fix: sidebar top margin too low when no FAB is present

AlexAndBear:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix


JammingBen:
Hm I think it was just fine before, because it needs to align the the breadcrumbs.  Now the breadcrumbs are sitting kinda in between the New button and the first nav item.

PR #2166, opened 2026-03-17
refactor: mail nav

JammingBen:
Refactors the mail nav to use the runtime one instead of its own implementation.

Also simplifies the visibility handling of the mail widget a bit. It was controlled within the component as well as on the outside. This doesn't make sense, so this change simplifies it to control the visibility from the outside only.

#### Known issue

The mobile 100% view of the nav currently doesn't work.

#### Next steps

Make the nav look nice like in other apps.

### DIFF
--- a/packages/web-app-mail/src/components/MailboxTree.vue
+++ b/packages/web-app-mail/src/components/MailboxTree.vue
@@ -2,7 +2,7 @@
   <div class="mailbox-tree px-1 flex flex-col">
     <app-loading-spinner v-if="isLoading" />
     <template v-else>
-      <no-content-message v-if="!mailboxes?.length" icon="folder-reduce" icon-fill-type="line">
+      <no-content-message v-if="!mailboxes?.length" img-src="/images/empty-states/empty-mails.svg">
         <template #message>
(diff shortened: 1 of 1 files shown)


packages/web-app-mail/src/views/Inbox.vue
```diff
@@ -81,7 +72,17 @@ const onCloseCompose = () => {
...
+  if (name === 'setCurrentAccount' && unref(currentAccount)) {
+    setCurrentMail(null)
+    await loadMailboxes(unref(currentAccount).accountId)
+    setCurrentMailbox(unref(mailboxes)[0])
+    await loadMails(unref(currentAccount).accountId, unref(currentMailbox).id)
+  }
+})
+
 onMounted(async () => {
+  mailComposeEventToken = eventBus.subscribe('app.mail.show-compose-mail', onComposeMail)
```
JammingBen, line 85: Using an event bus is the easiest solution I can think of right now. Another possibility would be to work with a store that keeps track of the visibility of the mail widget. But I feel like this wouldn't be an improvement.

PR #2171, opened 2026-03-18
Use favorites from graph api

pascalwengerter:
## Related Issue

- Fixes #1341
- Fixes #1342 
- Fixes #2170

## Types of changes

- [x] Enhancement (a change that doesn't break existing code or deployments)
- [ ] Breaking change (a modification that affects current functionality), if you consider removing the legacy favorite option "supported" (@kulmann)
- [x] Technical debt (addressing code that needs refactoring or improvements)

PR #2178, opened 2026-03-19
fix:list favorites endpoint

v-scharf:
fixed https://github.com/opencloud-eu/opencloud/issues/1633
here is new search favorites: https://github.com/opencloud-eu/opencloud/pull/2454

old favorites used old webdav `/files/{user}` endpoint and ` filterRules`

changed to using new DAV `/spaces` and correct search pattern `is:favorite`

result: 

https://github.com/user-attachments/assets/55318f6b-8897-43d5-9a42-7e190d54d0f2

@db-ot

### DIFF
--- a/packages/web-client/src/webdav/index.ts
+++ b/packages/web-client/src/webdav/index.ts
@@ -78,7 +78,7 @@ export const webdav = (baseURI: string, headers?: () => Headers): WebDAV => {
 
   const { search } = SearchFactory(dav, options)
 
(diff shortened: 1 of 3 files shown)


JammingBen:
https://github.com/opencloud-eu/web/pull/2171 also tackles this, but does it a bit differently. It ditches the `ListFavoriteFilesFactory` and uses the search factory instead (because it's the same functionality under the hood).

I don't have a strong opinion on this, @kulmann what do you think? I tend to favor using the search factory.

Issue #2159, opened 2026-03-17
Use `vue-tsc` to generate types for libraries


JammingBen:
We currently use `vite-plugin-dts` to generate types for packages we publish as library (design-system, client, pkg, test-helpers). This has a few drawbacks:

- It uses `@vue/language-core` v2, whereas `vue-tsc` uses v3. This results in different type reports when running `pnpm check:types` vs `pnpm vite build`.
- It uses rollup. With vite 8 around the corner, we want to get rid this dependency.
- One dependency more we need to care about.

We should switch to `vue-tsc` for generating types. It would solve all these issues and we're already using it for checking types.

PR #2194, opened 2026-03-20
build: use vue-tsc to generate `d.ts` files for libraries

JammingBen:
Removes the `vite-plugin-dts` dependency and uses `vue-tsc` to generate `d.ts` files for our library builds (pkg, client, test-helpers, design-system). This means one dependency less and ensures that we use the same type checking mechanism like Web.

Also adds type checks for libraries we build so we can see if emitting `d.ts` files on build would succeed.

closes https://github.com/opencloud-eu/web/issues/2159

~waiting for https://github.com/opencloud-eu/web/pull/2184 to get merged first.~

### DIFF
--- a/packages/web-client/package.json
+++ b/packages/web-client/package.json
@@ -90,7 +90,6 @@
   "devDependencies": {
     "@types/luxon": "3.7.1",
     "@types/node": "^25.5.0",
-    "@vitejs/plugin-vue": "6.0.5",
     "clean-publish": "6.0.4",
     "vite-plugin-node-polyfills": "0.25.0",
     "vite": "^8.0.0"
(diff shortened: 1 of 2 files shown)


packages/web-pkg/src/components/Filters/DateFilter.vue
```diff
@@ -96,15 +96,19 @@
               :is-clearable="true"
               :current-date="fromDate"
               :is-dark="currentTheme.isDark"
-              @date-changed="(value) => setDateRangeDate(value, 'from')"
+              @date-changed="
```
kulmann, line 99: this reverts the changes from https://github.com/opencloud-eu/web/pull/2184/changes to what it was before... what happened? 😅

JammingBen: As discussed in chat 😄

Issue #2128, opened 2026-03-11
Expired access tokens are not renewed when network is offline


Cassolette:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

User gets logged out due to a failed attempt at token renewal. This can fail often due to prolonged network interruption (e.g. no internet / device woke up from standby mode)

### Reproduction Steps

Environment -> OpenCloud compose with keycloak enabled
1. To easily simulate for testing a prolonged downtime exceeding the access token lifespan (default of 5 minutes), we reduce `accessTokenLifespan` to an extremely low value of 3 seconds (Keycloak Admin -> OC realm -> Realm settings -> Tokens -> Access Token Lifespan). 
2. Stay logged in the web UI
3. Start observing UI behaviour
4. Turn WiFi off / Put device to standby mode (simulate >3s downtime)
5. Turn WiFi back on / wake deviece (do not take too long to turn it back on though, it needs to be before `ssoSessionIdleTimeout` which defaults to 30 minutes)
6.  Observe UI behaviour 

### Expected Outcome

Ideally resume UI state as observed from step 3 (silent signin), or at least redirected to IDP/keycloak sign in. User could already be authenticated in the IDP

### Actual Outcome

User is forced to be logged out because of the silent signin failure, affecting UX and possibly discarding their progress

PR #2230, opened 2026-03-26
fix: idp logout on network errors

JammingBen:
Improves the general IDP flow on network errors in 2 ways:

- When the silent signin fails due to a network error, web now tries to re-authenticate every 5 seconds until the network is back up (or the browser tab gets closed). This should be safe to do since we're not causing any server load.
- When throwing an auth error due to an expired token, web now tries one more silent signin before actually throwing the error. This can happen asynchronously to the retry mechanism described above, e.g. when the network comes back, and the token has yet to be renewed, but the user performs an action in the meantime.

fixes https://github.com/opencloud-eu/web/issues/2128

supersedes https://github.com/opencloud-eu/web/pull/2129

### DIFF
--- a/packages/web-app-webfinger/src/views/Resolve.vue
+++ b/packages/web-app-webfinger/src/views/Resolve.vue
@@ -56,7 +56,7 @@ export default defineComponent({
       } catch (e) {
         console.error(e)
         if (e.response?.status === 401) {
-          return authService.handleAuthError(unref(route), { forceLogout: true })
+          return authService.handleAuthError(unref(route))
         }
         hasError.value = true
       }
--- a/packages/web-pkg/src/composables/authContext/useAuthService.ts
+++ b/packages/web-pkg/src/composables/authContext/useAuthService.ts
@@ -2,7 +2,11 @@ import { useService } from '../service'
 import { NavigationFailure } from 'vue-router'
 
 export interface AuthServiceInterface {
-  handleAuthError(route: any, options?: { forceLogout?: boolean }): any
+  handleAuthError(
+    route: any,
+    /** @deprecated This option is no longer used. */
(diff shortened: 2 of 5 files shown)


Cassolette:
Works quite reliably with this fix so far. 🙂Tested on MS Edge and Firefox

```js
[Dbg] AccessToken Expiring
[Dbg] New User Loaded
[Err] GET https://opencloud/ocs/v2.php/apps/notifications/api/v1/notifications/sse net::ERR_NETWORK_IO_SUSPENDED 200 (OK)
[Err] TypeError: network error
[Dbg] AccessToken Expiring
[Dbg] AccessToken Expired
[Err] POST https://keycloak/realms/openCloud/protocol/openid-connect/token net::ERR_NETWORK_CHANGED
[Err] [JsonService] postForm: Network error
[Warn] token renewal failed due to network error, retrying in 5 seconds...
[Err] POST https://keycloak/realms/openCloud/protocol/openid-connect/token net::ERR_NETWORK_CHANGED
[Err] [JsonService] postForm: Network error
[Warn] token renewal failed due to network error, retrying in 5 seconds...
[Err] POST https://keycloak/realms/openCloud/protocol/openid-connect/token net::ERR_NETWORK_CHANGED
[Err] [JsonService] postForm: Network error
[Warn] token renewal failed due to network error, retrying in 5 seconds...
[Dbg] New User Loaded
```

I guess this should be sufficient in catching most of the cases involving network errors to allow UI to recover more gracefully than before. 

Also tested that if the token worker attempts to renew an expired access token and received a non-network related error (i.e. due to invalidated refresh token), the fallback route via `access-denied` works neatly - delegating auth to the IDP where the sso session may still be valid.

kulmann (approved):
nice 💪

PR #2253, opened 2026-03-28
fix: prevent search results from being cleared on task cancellation

tiran133:
> **Disclaimer:** This bug was discovered by myself and fixed with the help of AI. This pull request description was also written by AI and reviewd by me. The fix was confirmed by myself using test methods below.

## Description

Fix a race condition where search results display "No results found" even though the search request returns valid results. This happens when the user is already on the search results page and performs a new search.

Two bugs are fixed:

1. **Stale route name in `Preview.available`** (`web-app-files/src/search/sdk/preview.ts`): The `Preview.available` getter checks against the route name `'search-provider-list'`, which does not exist in the codebase. The actual search results route is `'files-common-search'`. Because of this, `Preview.available` always returns `true` — even when the user is on the search results page — so the SearchBar's preview search is never disabled on that page.

2. **No cancellation handling in list search** (`web-app-search/src/views/List.vue`): Both the list search and the preview search share the same `searchTask` (created via `useSearch()` in `extensions.ts`), which uses vue-concurrency's `.restartable()` modifier. When the preview search fires (due to Bug 1 not disabling it on the search page), it cancels the in-flight list search. The `catch` block in `List.vue` does not distinguish cancellation errors from real errors, so it resets `searchResult` to empty and sets `loading` to `false` — overwriting valid results or preventing the UI from showing them.

### Race condition timeline (when already on the search page):

1. Route query changes → `List.vue` emits `'search'` → `searchTask.perform(term, 200)` — **Request A** starts
2. `SearchBar.vue`'s `$route` watcher fires `parseRouteQuery(route, false)` → sets `restoreSearchFromRoute = false` → `term` watcher calls `debouncedSearch()` (500ms debounce)
3. ~500ms later: `debouncedSearch` fires → `previewSearch.search()` → `searchTask.perform(term, 8)` — **Request B** starts, **cancels Request A** via `.restartable()`
4. Request A rejects with the string `'cancel'` → `catch` block resets `searchResult = { values: [], totalResults: null }`, `loading = false` → **"No results found"**
5. Request B resolves → updates SearchBar's preview dropdown, but **not** the list page's `searchResult` ref

## Related Issue

If you want me to create an Github Issue first, let me know.

## Motivation and Context

When using OpenCloud and performing a search while already on the search results page causes the results to be cleared. The user sees "No results found" even though the network request eventually returns valid results. This is confusing and breaks the search experience, particularly for extensions that navigate to the built-in search page.

## How Has This Been Tested?

- test environment: Chrome DevTools Network tab with network throttling enabled (e.g. "Slow 3G" or "Fast 3G") to reliably reproduce the race condition. The bug requires slow network conditions so that Request A is still in-flight when the SearchBar's debounced preview search fires ~500ms later and cancels it.
- test case 1: Enable network throttling in DevTools, navigate to search results page, then change the search term — verify results are displayed correctly instead of "No results found"
- test case 2: Enable network throttling, perform a search from a non-search page (first navigation) — verify behavior is unchanged and results display correctly
- test case 3: Enable network throttling, rapidly change search terms while on the search results page — verify no stale/empty results are shown
- test case 4: Disable network throttling and repeat the above — verify everything still works correctly under normal network conditions

## Screenshots (if appropriate):

N/A

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/tests/e2e/support/objects/app-files/resource/actions.ts
+++ b/tests/e2e/support/objects/app-files/resource/actions.ts
@@ -1603,6 +1603,9 @@ export const searchResourceGlobalSearch = async (
     return
   }
 
+  // wait for tika indexing
+  await new Promise((resolve) => setTimeout(resolve, 500))
+
   await Promise.all([
     page.waitForResponse((resp) => resp.status() === 207 && resp.request().method() === 'REPORT'),
     page.locator(globalSearchInput).fill(keyword)


tiran133:
Just realized that it might be even better to cancel the debounce in `onKeyUpEnter`

But neither is handling the case properly. When you do the following

- You search and hit enter. 
- While the search is running, type something in the search box, but do not press enter.
- Preview returns and gets displayed.
- Search page stuck in loading state/ No results, depending on if you include the if statement in my pull request.

That's when you mention that both searches share the same task.

packages/web-app-files/src/search/sdk/preview.ts
```diff
@@ -23,9 +23,6 @@ export default class Preview implements SearchPreview {
   }
 
   public get available(): boolean {
-    return (
-      unref(this.router.currentRoute).name !== 'search-provider-list' &&
-      !this.configStore.options?.embed?.enabled
-    )
```
JammingBen, line 29: Please leave the `!this.configStore.options?.embed?.enabled` check in for now, or does it solve any issue you're still having?

tiran133: I have revered it to the previous state. Or do you want me to remove just the route state check?

There is a test for that as well, but just for the route check, not for the `options?.embed?.enabled` check.

JammingBen: Removing the route check was fine (as it still has the invalid `search-provider-list` route), I just wanted to keep the embed check 😄 Sorry for the confusion.

tiran133: No problem. I have removed the check and updated the test to test the embedded check.
But there is now the unused router reference in that component. I can remove that too but not sure if 

https://github.com/opencloud-eu/web/blob/fb08409f4742698e94d872bd98531c77ef4acb65/packages/web-app-files/src/search/sdk/index.ts#L32

Is it the only reference?

tiran133: Just checked again; we would need to remove it here too. Unless you want to keep it for future use, maybe?

https://github.com/opencloud-eu/web/blob/fb08409f4742698e94d872bd98531c77ef4acb65/packages/web-app-files/src/extensions.ts#L54

JammingBen: Thanks! You should be able to remove the `router `, yes. Running `pnpm check:types` will then reveal the remaining references. If they're all within `packages/web-app-files` (which they should be), it's safe to remove them.

tiran133: Done! 
Search still works. 😄

PR #2274, opened 2026-03-31
feat: focus global search bar via keyboard shortcut

### LINKED ISSUE #2161: Focus global search bar via keyboard shortcut
### Initial Checklist

### User Story

As a power user, I want to be able to quickly focus the global search bar with my keyboard so that I don't need to reach for the mouse necessarily.

### User Value

UX

### Acceptance Criteria

- the global search bar can be focussed when typing `s` or `/`
- the search bar indicates more clearly that it's focussed (visually)

AlexAndBear:
## Related Issue

- Fixes #2161 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix


packages/web-app-search/src/portals/SearchBar.vue
```diff
@@ -362,6 +369,25 @@ export default defineComponent({
       term.value = ''
     })
 
+    const { bindKeyAction } = useKeyboardActions({ skipDisabledKeyBindingsCheck: true })
```
JammingBen: Why set `skipDisabledKeyBindingsCheck: true`? 🤔 Leaving it out and you could remove the `isEditableElement` helper, no?

packages/web-app-search/src/portals/SearchBar.vue
```diff
@@ -362,6 +369,25 @@ export default defineComponent({
...
+      }
+
+      const inputElement = unref(searchBarRef)?.querySelector('input') as HTMLElement
+      inputElement?.focus()
+
+      event.preventDefault()
+    }
+
+    bindKeyAction({ primary: Key.S }, onSearchShortcut, { preventDefault: false })
+    bindKeyAction({ primary: Key.Slash, modifier: Modifier.Shift }, onSearchShortcut, {
```
JammingBen: The modifier tailors it heavily to the German keyboard layout. I'd just leave out the modifier, or add another shortcut for slash only.

packages/web-app-search/src/portals/SearchBar.vue
AlexAndBear: ~~unfortantley it had no effect / didn't work as expected~~

Nope, you are right, works just fine

packages/web-app-search/src/portals/SearchBar.vue
AlexAndBear: ~~true~~

We need both unfortantley

Issue #2009, opened 2026-02-19
Contacts: View Details


kulmann:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

> * As a user, I want to view details of a contact so that I can find more information than just name and email from the list view.

### User Value

Contacts MVP

### Acceptance Criteria

- contact list has a three dots button per entry
- three dots button shows "Details" as context menu entry
- clicking the "Details" context menu entry or the contact name triggers showing the Details of the contact
- details view shows all available data for the contact in collapsible panels, like in the attached mockup
- layout similar to mail:
  - on desktop: left sidebar for mail folder selection, left area of the main screen for contacts listing of the selected contacts folder, right half of the screen for viewing the details of the selected account
  - on mobile: details are a full viewport overlay (similar to reading an email)

NOT part of this story:
- don't show the "write email" button
- don't show "Edit" or "Delete" buttons, yet
- no tags
- no attached images/files

<img width="375" height="2515" alt="Image" src="https://github.com/user-attachments/assets/96d9f00e-3196-4d4b-82b1-b5cccdef30b0" />

kulmann:
@pbleser-oc how does stalwart handle labels for fields, e.g. "Work email" and "Personal email" or even a custom label for an email address?

pbleser-oc:
The example in the screenshot above can be represented as JSContact (JMAP) as follows:
```json
{
  "id": "loTh8ahmubei",
  "addressBookIds": {
    "79047052-ae0e-4299-8860-5bff1a139f3d": true
  },
  "@type": "Card",
  "version": "1.0",
  "created": "2025-07-09T07:12:28+02:00",
  "kind": "individual",
  "language": "en-GB",
  "prodId": "OpenCloud Groupware 1.0",
  "relatedTo": {
    "096fc269-ffc4-4f2f-978f-5bd1bb0962d5": {
      "relation": {
        "manager": true
      }
    }
  },
  "uid": "ffcb3f80-8334-46ff-882b-fcb7dff49204",
  "updated": "2025-07-10T09:58:01+02:00",
  "name": {
    "@type": "Name",
    "components": [
      {
        "value": "Bessie",
        "kind": "given"
      },
      {
        "value": "Cooper",
[...]

PR #2322, opened 2026-04-08
Add contact details view

tammi-23:
Implements https://github.com/opencloud-eu/web/issues/2009

### DIFF
--- a/packages/web-app-contacts/src/views/Contacts.vue
+++ b/packages/web-app-contacts/src/views/Contacts.vue
@@ -34,28 +34,30 @@ import { storeToRefs } from 'pinia'
 import { onMounted, ref, unref } from 'vue'
 import { useAddressBooksStore } from '../composables/piniaStores/addressbooks'
 import { useLoadAddressBooks } from '../composables/useLoadAddressbooks'
-import { AddressBook } from '../types'
+import { AddressBook, Contact } from '../types'
 import { useLoadContacts } from '../composables/useLoadContacts'
 import ContactsList from './ContactsList.vue'
 import ContactDetails from './ContactDetails.vue'
 import { useContactsStore } from '../composables/piniaStores/contacts'
 
-const { httpAuthenticated } = useClientService()
 const accountsStore = useGroupwareAccountsStore()
-const { loadCurrentAccount } = accountsStore
-const { currentAccount } = storeToRefs(accountsStore)
 const addressBooksStore = useAddressBooksStore()
-const { setCurrentAddressBook } = addressBooksStore
-const { addressBooks, currentAddressBook } = storeToRefs(addressBooksStore)
 const contactsStore = useContactsStore()
-const { currentContact } = storeToRefs(contactsStore)
+const { loadCurrentAccount } = accountsStore
+const { setCurrentAddressBook } = addressBooksStore
+const { setCurrentContact } = contactsStore
+const { currentAccount } = storeToRefs(accountsStore)
+const { addressBooks, currentAddressBook } = storeToRefs(addressBooksStore)
+const { contacts, currentContact } = storeToRefs(contactsStore)
 const { loadAddressBooks } = useLoadAddressBooks()
 const { loadContacts } = useLoadContacts()
+const { httpAuthenticated } = useClientService()
 
 const isLoading = ref<boolean>(true)
 
 const currentAccountIdQuery = useRouteQuery('accountId')
 const currentAddressBookQuery = useRouteQuery('addressBookId')
+const currentContactQuery = useRouteQuery('contactId')
 
 const loadAddressBooksAndContacts = async () => {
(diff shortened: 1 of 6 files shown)


AlexAndBear (changes requested):
Some first observations:

- While reloading the page when a contact got selected, this contact won't be selected anymore but the `contactId`query stays intact. We want to preselect the contact from the query

- In desktop view the first card with the avatar has not enough spacing to the top

- In the mockups, the card headers have icons in front of the the headline, which is missing in this PR, did we decide to not go with that ? 

- the addresses' card is initially collapsed, but I can't find this in the acceptance criteria

packages/web-app-contacts/src/views/ContactsList.vue
```diff
@@ -49,4 +121,39 @@ const sortedContacts = computed(() => {
...
+  return `contact-actions-toggle-${contactId}`
+}
+
+function getContactActionsDropId(contactId: string) {
+  return `contact-actions-drop-${contactId}`
+}
+
+function getContactMenuSections(contact: Contact): MenuSection[] {
+  return [
+    {
```
AlexAndBear: please use arrow functions for consistency

AlexAndBear (changes requested):

packages/web-app-contacts/src/views/Contacts.vue
```diff
@@ -3,7 +3,22 @@
...
+          class="min-w-0 w-full overflow-y-auto md:w-[22rem] md:shrink-0 md:border-r-2 md:border-role-outline-variant"
+          :class="{ 'hidden md:block': currentContact }"
+        >
+          <ContactsList />
+        </div>
+        <div
+          class="min-w-0 w-full overflow-y-auto bg-role-surface md:flex-1"
+          :class="{ 'hidden md:block': !currentContact }"
+        >
+          <ContactDetails :key="currentContact?.id" />
```
AlexAndBear, line 18: Why do we need :key here, as we are not in a for loop or such ?

AlexAndBear (changes requested):

packages/web-app-contacts/src/views/Contacts.vue
tammi-23: The Key is used to force a remount on a contact change, so the detials view starts with fresh local state.

AlexAndBear:
Great so far!

Can you check if we can use (other than in figma) the line type icons, as I think they will look much more modern.

One additional thing I noticed, and I guess this is the same with the mail app:
If we scroll to the end of the list, select a contact and reload the page, the contacts details list won't scroll automatically to the selected item. 
I guess a simple watcher on the `currentContact` with `immediate : true` as options and `element.scrollIntoView` will do the trick, we don't need complicated checks here if we did a reload or something, as it won't harm if the element is already visible in the view port. 

Please adjust 👍

packages/web-app-contacts/src/views/ContactsList.vue
```diff
@@ -49,4 +122,52 @@ const sortedContacts = computed(() => {
...
+            setCurrentContact(contact)
+          }
+        }
+      ]
+    }
+  ]
+}
+
+watch(
+  [() => currentContact.value?.id, () => isLoading.value],
```
AlexAndBear: Since both values are refs I don't think you need to use arrow functions and access the currentContact.value 

[currentContact, loading] should be enough.  


As well just use the original variables to make it as simple as possible.

AlexAndBear (changes requested):

packages/web-app-contacts/src/views/ContactsList.vue
JammingBen: Hmm I think watching for the contact's `id` is actually better here (if we're only watching for a contact change). No need for expensive watchers on entire objects.

packages/web-app-contacts/src/views/ContactsList.vue
```diff
@@ -1,9 +1,29 @@
...
-    <div v-if="isLoading" class="flex h-full justify-center items-center">
-      <oc-spinner :aria-label="$gettext('Loading contacts')" />
-    </div>
+  <app-loading-spinner v-if="isLoading" />
+  <template v-else>
+    <no-content-message v-if="!currentAddressBook" icon="folder" icon-fill-type="line">
+      <template #message>
+        <span v-text="$gettext('No contact folder selected')" />
+      </template>
+    </no-content-message>
```
AlexAndBear: Please use the image `packages/design-system/src/assets/images/empty-states/empty-contacts.svg`

You can find how to use it in packages/web-app-contacts/src/views/ContactsList.vue:L30

AlexAndBear (changes requested):

AlexAndBear (approved):
Purrfect 🥇

PR #2339, opened 2026-04-10
fix(extension-sdk): add missing color roles

JammingBen:
Adds the missing color roles to the `tailwind.css` file so extensions can use all tailwind color classes with the `ext:` prefix, e.g. `ext:bg-role-surface`.


AlexAndBear (approved):
👍

PR #2341, opened 2026-04-10
Scroll selected mail into view after reload

tammi-23:
Added scrolling to the selcted mail in the mail list view.


AlexAndBear (approved):
👍

PR #2351, opened 2026-04-12
feat(preview): play next audio file automatically after current track…

FrankM64:
## Description

When playing audio files in the preview app, playback now automatically advances to the next file when the current track ends. Playback stops after the last file.

Manual navigation (arrow keys, next/prev buttons) still disables autoplay as before.

## How Has This Been Tested?

- test environment: OpenCloud 6.0.0, Firefox/Chromium on Manjaro KDE
- test case 1: Open a folder with multiple MP3 files → playback advances automatically to next track
- test case 2: Last track ends → playback stops, does not wrap around to first track
- test case 3: Manual navigation via arrow buttons → autoplay stays disabled

## Types of changes

## Notes
This is my first pull request ever — I hope everything is in order. Happy to fix anything if needed!

The implementation was developed with the assistance of Claude (Anthropic).

### DIFF
--- a/packages/web-app-preview/src/App.vue
+++ b/packages/web-app-preview/src/App.vue
@@ -434,12 +434,12 @@ watch(activeMediaFile, (newValue, oldValue) => {
   currentImageRotation.value = 0
 
   if (oldValue !== null) {
-      if (unref(isAutoAdvancing)) {
-        isAutoAdvancing.value = false
-      } else {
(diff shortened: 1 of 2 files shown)


JammingBen (commented):
Thanks, Code LGTM. Formatting looks a bit off though, running `pnpm check:all` makes sure that everything is alright.

JammingBen (approved):
👍

PR #2399, opened 2026-04-20
feat!: remove unmaintained project specific features

### LINKED ISSUE #2267: Removal of CERN specific logic
We have quite a few CERN specific features and logic in the web client codebase:

* Full share owner paths: shows full paths for received shares, revealing the actual location of the shared file or folder to the share receiver.
* CERN build: allows building the web client with a custom CERN build flag, altering some features at build time.
* Share notifications: notifications for creators of shares, using the old OCS API which we don't use anymore.
* Account types: select CERN specific account types when sharing.
* Reva auth token handling: Custom logic in our OIDC implementation to satisfy CERNs needs for auth handling.
* A few small configs, controlled via the `cernFeatures` feature flag.

All these things have been solely kept to ensure compatibility with CERN's needs and the reva backend. Neither of the features are getting tested or used in our product. Some are simple feature flags, other things like the share owner paths reach deep into our codebase, making it more complicated than it needs to be.

In order to keep our code base maintainable we are going to remove the features and flags described above.

JammingBen:
### web-pkg

**Removed methods**

- `isResourceAccessible`. Resources are always accessible if the user has access to the space.

**Changed signatures**

- removed `loading` from `useDriveResolver`
- removed `space` from `breadcrumbsFromPath` (`useBreadcrumbsFromPath` composable)
- removed `configStore` from `getSpacesByType`
- removed `options` from `createFileRouteOptions`

**Removed options from config store**

- `routing`
- `ocm`
- `cernFeatures`

closes https://github.com/opencloud-eu/web/issues/2267


AlexAndBear (approved):
God bless 🙏

PR #2414, opened 2026-04-21
perf: improve initial loading times

### LINKED ISSUE #2335: Optimize initial loading times
I feel like initial loading times could still be a bit better, which gets especially noticeable on slow connections. Chunks being too big is one part of the problem (see https://github.com/opencloud-eu/web/issues/2332). But it's also a matter of when we load things.

The current loading flow:

The first batch of chunks consists of everything `index.html` loads and, as a result, the top-level imports of `packages/web-runtime/src/index.ts`. Next, the actual bootstrap begins via `bootstrapApp()`, loading a second batch of chunks (mainly config + all the apps). After the apps have been loaded, we announce the router, which then triggers the auth flow so the user gets loaded. After that, the bootstrapping process comes to an end by loading the spaces and the authenticated config (if a user is present).

I think the biggest improvement would come from parallelizing the app loading and the auth flow. This needs a bit of tinkering and testing to avoid catch-22 scenarios, but it should be possible. Plus, within `bootstrapApp()` and during user loading there is also room for smaller optimizations.

JammingBen:
This PR mainly does two things to improve loading times, especially on slower network connections:

- Optimize chunk sizes by splitting the large chunks into smaller ones.
- Parallelize some requests that were running synchronously before during bootstrap:
  - webfinger discovery
  - design system js assets
  - capabilities
  - user permissions
  - app provider mime types

Tests on my local machine showed reduced loading times with throttled network:

| Network | Total loading before | After | Decrease |
| -------- | -------- | -------- | -------- |
| Fast 4G | ~4,1 s | ~3,4 s | 0.7 s |
| Slow 4G | ~16.4 s | ~14 s | 2.4 s |

closes https://github.com/opencloud-eu/web/issues/2332
closes https://github.com/opencloud-eu/web/issues/2335


packages/web-pkg/src/composables/actions/files/useFileActionsEmptyTrashBin.ts
```diff
@@ -54,7 +52,7 @@ export const useFileActionsEmptyTrashBin = () => {
         }
       ),
       hasInput: false,
-      onConfirm: () => loadingService.addTask(() => emptyTrashBin({ space: resources[0] }))
+      onConfirm: () => emptyTrashBin({ space: resources[0] })
```
JammingBen, line 55: `loadingService` is already active for modals. This failed in e2e tests (which is good, I just don't know why now).

AlexAndBear (approved):
👍

PR #2423, opened 2026-04-22
feat(search): clear on esc keypress + misc fixes

Cassolette:
## Description

Currently there is no straightforward way to clear the search input. Users have to use traditional means to do so (i.e. select all the text and delete them) which can make file finding by multiple attempts a little troublesome.

To improve search productivity while using keyboard controls, a shortcut keybinding via Esc is introduced to the search bar. While in focus, any search term is cleared when Esc is pressed. ~~Otherwise if there's no search term, the focus on the text input will be rescinded to restore regular operations.~~ (removed per https://github.com/opencloud-eu/web/pull/2423#discussion_r3129948573)

This behaviour is being referenced from GitHub's file finder. OneDrive also has an esc key binding that clears the search input.

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/web-app-search/src/portals/SearchBar.vue
+++ b/packages/web-app-search/src/portals/SearchBar.vue
@@ -528,13 +528,11 @@ export default defineComponent({
           return
         }
         const routeTerm = route?.query?.term
-        const input = this.$el.getElementsByTagName('input')[0]
-        if (!input || !routeTerm) {
+        if (!routeTerm) {
           return
         }
         this.restoreSearchFromRoute = initialLoad
         this.term = queryItemAsString(routeTerm)
-        input.value = queryItemAsString(routeTerm)
       })
     },
     getMoreResultsDetailsTextForProvider(provider: SearchProvider) {
--- a/packages/web-app-search/src/portals/SearchBar.vue
+++ b/packages/web-app-search/src/portals/SearchBar.vue
@@ -24,7 +24,7 @@
       @update:model-value="updateTerm"
       @clear="onClear"
       @click="showPreview"
-      @keyup.esc="hideOptionsDrop"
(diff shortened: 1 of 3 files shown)


packages/web-app-search/src/portals/SearchBar.vue
```diff
@@ -513,6 +513,14 @@ export default defineComponent({
     onKeyUpDown() {
       this.activePreviewIndex = this.findNextPreviewIndex(false)
     },
+    onKeyUpEsc() {
+      if (this.term) {
+        this.updateTerm('')
+        return
+      }
+      const inputElement = this.$el.getElementsByTagName('input')[0] as HTMLElement
+      inputElement.blur()
```
JammingBen: I'd rather remove the `blur` and only keep clearing the term. It feels a bit weird to reset the focus on an input on esc. The `blur` works for GitHub because the search expands into this "overlay", which we don't have.

JammingBen (changes requested):

packages/web-app-search/src/portals/SearchBar.vue
Cassolette: Oh no not that search button.. guess I should've been clearer. My reference point on esc unfocusing an input came from GDrive and this repo file finder:

img

I'll remove the blur later 🙂

Cassolette:
Also added a little fix for mobile dark mode. The search bar cancel button was using the wrong inverse color (role-on-chrome vs role-on-secondary).

#### Dark (Before/After)
image

image

#### Light (No change)
image

JammingBen (approved):
LGTM 👍

PR #2433, opened 2026-04-22
feat: show mobile sidebar and FAB on tablets in portrait mode

AlexAndBear:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/design-system/src/composables/useIsMobile/index.ts
+++ b/packages/design-system/src/composables/useIsMobile/index.ts
@@ -2,17 +2,21 @@ import { computed } from 'vue'
 import { useWindowSize } from '@vueuse/core'
 
 const MOBILE_BREAKPOINT = 640
-const TABLET_BREAKPOINT = 840
+const TABLET_BREAKPOINT = 960
 
-export const useIsMobile = ({ includeTablet = false } = {}) => {
+export const useIsMobile = () => {
   const { width } = useWindowSize()
 
   const isMobile = computed(() => {
-    const breakpoint = includeTablet ? TABLET_BREAKPOINT : MOBILE_BREAKPOINT
-    return width.value <= breakpoint
+    return width.value < MOBILE_BREAKPOINT
+  })
+
+  const isTablet = computed(() => {
+    return width.value <= TABLET_BREAKPOINT
   })
 
   return {
-    isMobile
+    isMobile,
+    isTablet
   }
 }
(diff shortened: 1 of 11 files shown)


kulmann:
What do you think about setting a max-width on the mobile sidebar? Our desktop resolution sidebar has a width of 230px. I find that rather narrow on mobile (see screenshots). Something like 320px feels more natural. I like that better than going 85% width regardless of the viewport size. Last screenshot is showing 85% on a tablet width.

Screenshot 2026-04-23 at 07 48 01

Screenshot 2026-04-23 at 07 48 41

Screenshot 2026-04-23 at 07 49 53

Screenshot 2026-04-23 at 07 44 06

packages/design-system/src/composables/useIsMobile/index.ts
```diff
@@ -1,11 +1,15 @@
 import { computed } from 'vue'
 import { useWindowSize } from '@vueuse/core'
 
-export const useIsMobile = () => {
+const MOBILE_BREAKPOINT = 640
+const TABLET_BREAKPOINT = 840
```
JammingBen: Tailwind defines `960px` as breakpoint, I think we should use the same here.

packages/design-system/src/composables/useIsMobile/index.ts
```diff
@@ -1,11 +1,15 @@
...
+
+export const useIsMobile = ({ includeTablet = false } = {}) => {
   const { width } = useWindowSize()
 
   const isMobile = computed(() => {
-    return width.value < 640
+    const breakpoint = includeTablet ? TABLET_BREAKPOINT : MOBILE_BREAKPOINT
+    return width.value <= breakpoint
   })
```
JammingBen, line 16: Hmm why not add another computed `isTablet`? I would find that a bit more intuitive to use.

AlexAndBear:
> What do you think about setting a max-width on the mobile sidebar? Our desktop resolution sidebar has a width of 230px. I find that rather narrow on mobile (see screenshots). Something like 320px feels more natural. I like that better than going 85% width regardless of the viewport size. Last screenshot is showing 85% on a tablet width.
> 
> Screenshot 2026-04-23 at 07 48 01 Screenshot 2026-04-23 at 07 48 41 Screenshot 2026-04-23 at 07 49 53 Screenshot 2026-04-23 at 07 44 06

I chose a more balanced way:

on mobile (smartphone) 85%, on tablet sizes 320px as it feels more natural that way

JammingBen (approved):

PR #2443, opened 2026-04-27
change: remove or reduce dummy  methods

### LINKED ISSUE #1336: Remove or reduce dummy `$gettext` methods
We have several places where we have dummy `$gettext` methods just to make sure that a string gets synced to Transifex. Many of them are not needed I suppose if we would just refactor those parts into composables. For all other places, the `$gettext` provided by `web-pkg` should be used instead of re-defining the method over and over again.

AlexAndBear:
## Related Issue

- Fixes #1336 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix


packages/web-app-pdf-viewer/src/index.ts
JammingBen, line 1: Or refactor this to use `defineWebApplication` and then use `useGettext` like we do for most of the other apps?

packages/web-app-webfinger/src/index.ts
JammingBen, line 1: Same here

packages/web-pkg/src/helpers/ui/resourceTiles.ts
```diff
@@ -3,7 +3,6 @@ import { SortField } from '../../composables/sort'
 import { Language } from 'vue3-gettext'
 import { SortDir } from '@opencloud-eu/design-system/helpers'
 
-// just a dummy function to trick gettext tools
 function $gettext(msg: string) {
   return msg
 }
```
JammingBen: Can't you import it from the other location here as well?

JammingBen (approved):

PR #2453, opened 2026-04-29
refactor: ensure module federation gets initialized once

JammingBen:
No need for the module federation to get initialized more than once. Also bumps it to `v1.15.1` to fix recent issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "@cucumber/messages": "32.3.1",
     "@cucumber/pretty-formatter": "3.2.0",
     "@module-federation/runtime": "2.3.3",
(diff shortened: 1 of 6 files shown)


dschmidt (approved):

kja999:
Hi, It looks like this PR breaks how web extensions work. Do you need an issue to be created or will there be a change upcoming to how addons work ?

JammingBen:
> Hi, It looks like this PR breaks how web extensions work. Do you need an issue to be created or will there be a change upcoming to how addons work ?

I need a bit more details on what exactly broke. What extensions are affected? Any console errors? Also, did you update the affected extension(s) to their latest version?

Issue #2356, opened 2026-04-13
Move file and space actions to files app


JammingBen:
The file and spaces actions are currently located in `web-pkg`, where they technically not belong. This has historic reasons, but they should be placed inside the files app and registered/retrieved via the extension system. This would a) create a better separation of concerns and b) reduce `web-pkg`'s total size, which should result in better chunk sizing.

Note that this will be a breaking change, meaning we should first think about when we want to have the next major release.

JammingBen:
Follow-ups after first iteration via https://github.com/opencloud-eu/web/pull/2456:

- [x] Remove "Create Space from selection" should not appear in batch actions. The extension point for batch actions should not be reflected in the context menu, but in the top bar only. The actions control whether they are applicable for one or many resources in the context menu via `isVisible`. fixed via https://github.com/opencloud-eu/web/pull/2463
- [x] The "Details" action should be present the context menu when selecting multiple resources. fixed via https://github.com/opencloud-eu/web/pull/2463
- [x] Actions without `category` should default to the `tertiary` category in the context menu. Currently, they are missing completely. fixed via https://github.com/opencloud-eu/web/pull/2464
- [x] Add a dedicated extension point for sidebar actions. fixed via https://github.com/opencloud-eu/web/pull/2466
- [x] Write deprecation notice for upcoming `v7.0.0` (including how to retrieve the moved actions in a third party app)
- [ ] Add new extension points to documentation
- [ ] Move remaining actions from `web-pkg` to files app.
- [ ] The space actions for setting/editing/deleting the space image should be located in the sub-menu like on the spaces list page. This sub-menu is currently missing on other pages like the search result page.

PR #2456, opened 2026-04-29
refactor!: move certain file and space actions to files app

AlexAndBear:
## Related Issue

- Works towards https://github.com/opencloud-eu/web/issues/2356

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix


packages/web-pkg/src/components/AppBar/AppBar.vue
```diff
@@ -193,39 +179,29 @@ export default defineComponent({
...
       if (actionExtensions.length) {
-        actions = [...actions, ...actionExtensions.map((e) => e.action)]
+        const filteredExtensions = actionExtensions.filter((extension) => {
+          if (!isSearchLocation.value) {
+            return true
+          }
+
+          return !extension.id.includes('.spaces.')
+        })
+        actions = [...actions, ...filteredExtensions.map((e) => e.action)]
```
JammingBen: Good for now, but maybe worth a look in the future. Feels a bit hacky to set/exclude some actions depending on the route or if the id includes `spaces`. Could you add a `FIXME` comment?

packages/web-app-admin-settings/src/components/Spaces/ContextActions.vue
```diff
@@ -30,30 +21,35 @@ export default defineComponent({
   },
   setup(props) {
     const filterParams = computed(() => ({ resources: props.items }))
+    const { requestExtensions } = useExtensionRegistry()
+    const getActionExtensions = (extensionPoint: typeof spacesContextActionsExtensionPoint) => {
+      const extensions = requestExtensions ? requestExtensions<ActionExtension>(extensionPoint) : []
```
JammingBen: Why does this check if `requestExtensions` is defined? This should always be a given.

packages/web-app-admin-settings/src/views/Spaces.vue
```diff
@@ -139,18 +137,17 @@ const breadcrumbs = computed(() => [
...
 ])
 
-const { actions: deleteActions } = useSpaceActionsDelete()
-const { actions: disableActions } = useSpaceActionsDisable()
-const { actions: editQuotaActions } = useSpaceActionsEditQuota()
-const { actions: restoreActions } = useSpaceActionsRestore()
+const extensionBatchActions = computed(() => {
+  const extensions = requestExtensions
+    ? requestExtensions<ActionExtension>(spacesBatchActionsExtensionPoint)
+    : []
```
JammingBen: Same here.

packages/web-app-files/src/components/CreateOrUploadMenu.vue
```diff
@@ -174,12 +176,15 @@ const createFileActionsGroups = computed(() => {
   return result
 })
 
-const extensionRegistry = useExtensionRegistry()
-const extensionActions = computed(() => {
-  return [
-    ...extensionRegistry.requestExtensions(uploadMenuExtensionPoint).map((e) => e.action)
-  ].filter((e) => e.isVisible())
-})
+const extraExtensionActions = computed(() =>
```
JammingBen, line 179: I'm struggling a bit with the name, what does `extra` mean? Actions from other apps, registered via the extension system? Would be nice to clarify this, either via a more descriptive name or a comment.

packages/web-app-preview/src/components/MediaControls.vue
```diff
@@ -180,10 +182,23 @@ const emit = defineEmits<{
...
 
 const space = computed(() => getMatchingSpace(files[activeIndex].resource))
+const actionOptions = computed(() => ({
+  space: unref(space),
+  resources: [files[activeIndex].resource]
+}))
+const previewToolbarActions = computed(() => {
+  const extensions = requestExtensions
+    ? requestExtensions<ActionExtension>(previewToolbarActionsExtensionPoint)
+    : []
```
JammingBen: Another unneeded check on `requestExtensions`.

packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
```diff
@@ -617,34 +618,42 @@ const menuItemsActions = computed(() => {
...
+const menuItemsQuaternary = computed(() => {
+  return (
+    requestExtensions<ActionExtension>({
+      id: 'global.files.context-actions',
+      extensionType: 'action'
+    }) || []
+  )
+    .map((e) => e.action)
+    .filter((action) => action.category === 'quaternary')
+    .filter((item) => item.isVisible(unref(actionOptions)))
```
JammingBen, line 628: For performance reasons, I'd put the following part into a `computed` to avoid it getting called several times here:

```
(
    requestExtensions<ActionExtension>({
      id: 'global.files.context-actions',
      extensionType: 'action'
    }) || []
  )
    .map((e) => e.action)
```

packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -437,9 +439,16 @@ const hasTags = computed(
...
 const { actions: renameActions } = useFileActionsRename()
-const { actions: renameActionsSpace } = useSpaceActionsRename()
+const renameActionsSpace = computed(() =>
+  requestExtensions<ActionExtension>({
+    id: 'global.files.context-actions',
+    extensionType: 'action'
+  })
+    .map((e) => e.action)
+    .filter((action) => action.name === 'rename')
+)
```
JammingBen: Just thinking out loud, not for the scope of this PR: this is a case where an app wants to retrieve one specific action that is registered in the files app. We should think about what our recommended approach for this is. It probably comes down to these 2 options:

1. Do it like it's currently done: query for the `global.files.context-actions` extension point and then filter via the `rename` action name (requires knowledge about the extension point *and* the action name).
2. Allow querying extensions via their ID only (requires knowledge about the extension ID).

I tend to favor 2., because it's less boilerplate for the consumer and it only requires knowing one ID. The list of available IDs can easily be retrieved from `packages/web-app-files/src/composables/extensions/useSpaceActions.ts`. It would be kind similar to a microservice pattern where services communicate with each other over a message queue with message IDs.

packages/web-app-admin-settings/src/components/Spaces/SideBar/ActionsPanel.vue
```diff
@@ -31,22 +26,17 @@ const resources = computed(() => {
...
-const { actions: restoreActions } = useSpaceActionsRestore()
+const { requestExtensions } = useExtensionRegistry()
+const contextActions = computed(() => {
+  const extensions = requestExtensions
+    ? requestExtensions<ActionExtension>(spacesContextActionsExtensionPoint)
+    : []
+  return (extensions || [])
+    .filter((extension) => !hiddenContextActionIds.includes(extension.id))
+    .map((extension) => extension.action)
+})
```
JammingBen: Unneeded check for `requestExtensions`.

Also, this will be done via a dedicated extension point for the admin settings sidebar panel in the future, no? Could you please add a `FIXME` if so?

AlexAndBear: added app.admin-settings.spaces.sidebar-actions

packages/web-pkg/src/components/AppBar/AppBar.vue
AlexAndBear: I enabled the space actions

packages/web-pkg/src/components/FilesList/ResourceTable.vue
AlexAndBear: In this special case, I think I opt for number #3 add a sperate extension point (: 
Implemented

JammingBen (approved):
🥳

Issue #2010, opened 2026-02-19
Contacts: Create new contact


kulmann:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

> * As a user, I want to be able to create new contacts, so that I can add people to my address book.

### User Value

Contacts MVP

### Acceptance Criteria

- Add a FAB to the Contacts list view
- Triggering the FAB opens a bottom drawer panel on mobile or a modal on desktop
- initially (first iteration) has only these fields:
  - First name
  - Last name
  - E-Mail
  - Phone number
- Submitting the form...
   - creates a contact
   - navigates into the newly created contact
   - if navigating away causes data loss (= previously open edit contact form with unsaved data) we ask the user if they want to cancel the navigation

NOT part of this story:
- dynamic fields (followup)
- uploading a contact photo

<img width="375" height="1587" alt="Image" src="https://github.com/user-attachments/assets/8ddda541-1dd1-44d6-aff0-b05fd4ad9907" />

PR #2458, opened 2026-04-30
feat: create new contacts

tammi-23:
IMplemets https://github.com/opencloud-eu/web/issues/2010

### DIFF
--- a/packages/web-app-contacts/src/components/ContactCreateEditor.vue
+++ b/packages/web-app-contacts/src/components/ContactCreateEditor.vue
@@ -1,0 +1,211 @@
+<template>
+  <div
+    v-if="variant === 'drawer'"
+    class="fixed inset-0 z-[var(--z-index-modal)] bg-black/20 md:hidden"
+    @click.self="requestClose"
+  >
+    <div
+      class="absolute inset-x-0 bottom-0 flex max-h-[90vh] min-h-[50vh] flex-col rounded-t-2xl bg-role-surface shadow-xl"
+    >
+      <div class="flex justify-center px-4 pt-3">
+        <div class="h-1.5 w-12 rounded-full bg-role-outline-variant" />
+      </div>
+
+      <div class="flex items-center justify-between border-b border-role-outline-variant px-4 py-3">
+        <h2 class="text-lg font-bold" v-text="$gettext('New contact')" />
+        <oc-button appearance="raw" :aria-label="$gettext('Close')" @click="requestClose">
+          <oc-icon name="close" fill-type="line" />
+        </oc-button>
+      </div>
+
+      <div class="flex-1 min-h-0 overflow-auto px-4 py-4 pb-6">
+        <ContactCreateForm
+          :model-value="unref(createFormState)"
+          @update:model-value="onUpdateForm"
+        />
+      </div>
+
+      <div class="border-t border-role-outline-variant">
+        <div class="px-4 pt-3 pb-3">
+          <div class="flex items-center justify-start gap-3">
+            <oc-button
+              appearance="filled"
+              class="min-w-[120px]"
+              :disabled="unref(isSubmitDisabled)"
+              @click="onSubmit"
+            >
+              <span v-text="$gettext('Save')" />
+            </oc-button>
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+
+  <oc-modal
+    v-else
+    element-class="contact-create-modal"
+    :title="$gettext('New contact')"
+    :hide-actions="true"
+    hide-cancel-button
+  >
+    <template #headerActions>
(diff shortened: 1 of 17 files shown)


packages/web-app-contacts/src/components/ContactCreateEditor.vue
```diff
@@ -0,0 +1,210 @@
+<template>
```
AlexAndBear: the name "Editor" is kinda misleading, isn't it ?

AlexAndBear (dismissed):

packages/web-app-contacts/src/components/ContactCreateEditor.vue
tammi-23: Maybe yes. We could use ContactFormPanel.vue

packages/web-app-contacts/src/components/ContactCreateEditor.vue
```diff
@@ -0,0 +1,210 @@
...
+  }
+}
+
+const showSaveError = () => {
+  dispatchModal({
+    title: $gettext('Failed to save contact'),
+    message: $gettext('The contact could not be created. Please try again.'),
+    confirmText: $gettext('OK'),
+    hasInput: false
+  })
```
JammingBen: Do we want to show a modal on error? Usually we use notifications for this. Or did we decide to go that way here? I can't remember 🤔

packages/web-app-contacts/src/components/ContactCreateEditor.vue
```diff
@@ -0,0 +1,210 @@
...
+  if (!shouldLeave) {
+    return false
+  }
+
+  closeCreateContact()
+  return true
+}
+
+onBeforeRouteLeave(confirmRouteNavigation)
+onBeforeRouteUpdate(confirmRouteNavigation)
```
JammingBen: I know the story specifies a confirmation modal on route change, but is there a way the change the route with the modal/drawer showing? Both elements are present as overlay, making it impossible to navigate to another route. I think this is just unnecessary, or am I overlooking something?

packages/web-app-contacts/src/composables/useContactEditor.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  const setCreateFormState = (value: ContactFormState) => {
+    createFormState.value = value
+  }
+
+  const markCreateDirty = () => {
+    isCreateDirty.value = true
+  }
+
+  const confirmDiscardChanges = async () => {
```
JammingBen: Since there is no `await` expression in this method:

```suggestion
  const confirmDiscardChanges = () => {
```

packages/web-app-contacts/src/composables/useCreateContact.ts
```diff
@@ -0,0 +1,44 @@
...
+
+      const directContact = ContactSchema.safeParse(data)
+      if (directContact.success) {
+        return directContact.data
+      }
+
+      return WrappedContactResponseSchema.parse(data).contact
+    } finally {
+      isSaving.value = false
+    }
```
JammingBen, line 40: Seems like it's missing a catch phrase. It did you leave it out intentionally? I think we should at least add a `console.error` log, and probably add a generic notification message that something went wrong.

packages/web-app-contacts/src/components/ContactCreateForm.vue
```diff
@@ -0,0 +1,45 @@
...
+      :model-value="modelValue.lastName"
+      :label="$gettext('Last name')"
+      @update:model-value="(value: string) => updateField('lastName', value)"
+    />
+    <oc-text-input
+      :model-value="modelValue.email"
+      type="email"
+      :label="$gettext('E-Mail')"
+      @update:model-value="(value: string) => updateField('email', value)"
+    />
```
JammingBen, line 22: The `admin-settings` app validates email addresses via the `email-validator` lib, maybe something for here as well?

JammingBen (changes requested):
2 small things I'm noticing in the UI:

- The modal has the "Create"-button on the left side and a divider line at the top. Shouldn't this match the appearance of other modals (button on the right, no divider line)?
- The bottom drawer looks like it's missing a bit of padding to the sides.

packages/web-app-contacts/src/components/ContactCreateEditor.vue
tammi-23: No, you're right.
It was different with the mail widget, you were able to navigate with an open, collapsed widget.

JammingBen (commented):
One last nitpick from my side, the vertical padding of the headline in the bottom drawer is too big, I think it should match the other bottom drawers:

Image

JammingBen (approved):
👍

Issue #2455, opened 2026-04-29
[Epic] Tiptap integration


kulmann:
Pull request https://github.com/opencloud-eu/web/pull/2195 introduces a tiptap editor integration, which replaces the markdown editor we had before. It's a small set of editor components which is based on tiptap and offers support for content types plain-text, markdown, html and tiptap-json. The toolbar is content-type-sensitive. The editor offers slash command actions.

On top of the basic integration we to harden the integration by offering a really useful feature set via the sub issues linked in this epic.

## Sub-issues:

- [ ] Insert and edit Links
- [ ] Render inline base64 encoded images
- [ ] Upload and inline images base64 encoded
- [ ] toggle raw source view mode
- [ ] toggle line numbers (default off, available both in rendered and raw source view mode)
- [ ] drag to move blocks
- [ ] `+` on hover for slash command menu
- new actions for toolbar:
  - [ ] export as PDF
  - [ ] import from Word
  - [ ] text-alignment (only for html and tiptap-json)
- [ ] toolbar: inline font size controls
- [ ] toolbar: inline "turn into" controls (h1, h2, h3, text, code, quote, etc)
- [ ] toolbar: render colors in color dropdowns (background and text color)

db-ot:
related to https://github.com/opencloud-eu/internal/issues/113

lukas-scm:
I already created a web app integration with tiptap, maybe it can be used for further development: [opencloud-markdown-editor](https://github.com/lukas-scm/opencloud-markdown-editor).

JammingBen:
https://github.com/opencloud-eu/web/issues/2757 is the only left item, PR for this is ready and about to be reviewed and merged.

Exporting & Importing will be handled separately.

Hence closing.

PR #2461, opened 2026-05-05
feat(editor): add insert image functionality

JammingBen:
Adds the option to insert images from the local disk or a URL to the toolbar and the slash command menu. Also fixes displaying those images in md files and another issue with the centered placement of the toolbar.

Note that the drop menus of the toolbar are still not looking very good. I want to tackle this in a follow-up.

works towards https://github.com/opencloud-eu/web/issues/2455


AlexAndBear:
GREAT 🤩 

We are missing an icon for the image from url action in the slash menu 

image

AlexAndBear (approved):
Approved, can be merged after the missing icon is present

**edit: fixed the icon**

Issue #2005, opened 2026-02-19
Contacts list: search


kulmann:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

> * As a user looking for a certain contact in my address book, I want to filter the list of contacts by a search term, so that I can find the contact quickly (or learn that I don't have them in my address book)

### User Value

Contacts discoverability

### Acceptance Criteria

- above the list of contacts there is a search input, exactly the same as above the users list in the admin settings app
- entering a search term filters the currently displayed contacts by firstname, lastname and email partial matches (to be discussed: client side or can we do server side?)
- entering a search term has a short debounce if we do server side search
- the search term is highlighted/marked in the filtered results (e.g. if search for "phi", we'd highlight "phi" in "Sophia")
- if no results remain, we show a short message and a nice icon

PR #2472, opened 2026-05-07
feat: adds client-side search to the contact list

tammi-23:
Implement Contact Search https://github.com/opencloud-eu/web/issues/2005


packages/web-app-contacts/src/composables/useContactSearch.ts
JammingBen, line 1: The search implementation looks a bit complicated 🙈 You can check the other places where we have similar functionality via `fuse` and `mark`, e.g.: https://github.com/opencloud-eu/web/blob/main/packages/web-app-admin-settings/src/components/Groups/SideBar/MembersPanel.vue

JammingBen (approved):
👍

PR #2475, opened 2026-05-07
fix: align space action permissions with backend abilities

tiran133:
## Description

Space actions (`canBeDeleted`, `canRename`, `canEditDescription`, `canRestore`)
now honour the abilities reported by the backend when user has "create-all" permission.

According to 
https://github.com/opencloud-eu/reva/blob/main/pkg/storage/utils/decomposedfs/spaces.go#L1115-L1116
The User should be able to disable/restore and delete its managed spaces.

Backend Permission for `create-all` frontend permission.
```
{
        "id": "79e13b30-3e22-11eb-bc51-0b9f0bad9a58",
        "name": "Drives.Create",
        "displayName": "Create Space",
        "description": "This permission allows creating new spaces.",
        "permissionValue": {
          "operation": "OPERATION_READWRITE",
          "constraint": "CONSTRAINT_ALL"
        },
        "resource": {
          "type": "TYPE_SYSTEM"
        }
      },
```

Although it won't fix the underlaying issue that the owner of a space (User role) should be able to disable/delete/restore via the frontend. Currently, only disable is possible. Restoring or deleting a space as manager of that space with the User role is still not possible.

The comment in the functions.ts suggest problem with the permission model
https://github.com/opencloud-eu/web/blob/main/packages/web-client/src/helpers/space/functions.ts#L244

At least this honour `Drives.Create` with `CONSTRAINT_ALL`

## Related Issue

- Fixes 
https://github.com/opencloud-eu/opencloud/issues/1799

## How Has This Been Tested?

- test environment: local dev setup (`pnpm vite` against a local OpenCloud backend via `docker-compose up -d`)
- test case 1: unit tests in `packages/web-client/tests/unit/helpers/space/functions.spec.ts` updated/added to assert action-specific permission handling for `canBeDeleted`, `canRename`, `canEditDescription`, and `canRestore` (verified with `pnpm test:unit --run packages/web-client/tests/unit/helpers/space/functions.spec.ts`)
- test case 2: manual check in the UI for a disabled space — restore/delete/rename/edit-description actions only appear when the backend grants the corresponding ability
- test case 3: `pnpm check:all` passes (types, lint, format, unit tests)

## Types of changes

- [x] Bugfix

### DIFF
--- a/packages/web-client/src/helpers/space/functions.ts
+++ b/packages/web-client/src/helpers/space/functions.ts
@@ -214,21 +214,21 @@ export function buildSpace(
       if (!this.disabled) {
         return false
       }
-      if (ability?.can('delete-all', 'Drive')) {
+      if (ability?.can('delete-all', 'Drive') || ability?.can('create-all', 'Drive')) {
         return true
       }
       // FIXME: server permissions are a mess currently: https://github.com/opencloud-eu/opencloud/issues/10
       return this.graphPermissions?.includes(GraphSharePermission.deletePermissions)
     },
     canRename: function ({ ability }: { user?: User; ability?: Ability } = {}) {
-      if (ability?.can('update-all', 'Drive')) {
+      if (ability?.can('update-all', 'Drive') || ability?.can('create-all', 'Drive')) {
         return true
       }
       // FIXME: server permissions are a mess currently: https://github.com/opencloud-eu/opencloud/issues/10
       return this.graphPermissions?.includes(GraphSharePermission.deletePermissions)
     },
     canEditDescription: function ({ ability }: { user?: User; ability?: Ability } = {}) {
-      if (ability?.can('update-all', 'Drive')) {
+      if (ability?.can('update-all', 'Drive') || ability?.can('create-all', 'Drive')) {
         return true
       }
       // FIXME: server permissions are a mess currently: https://github.com/opencloud-eu/opencloud/issues/10
@@ -238,7 +238,7 @@ export function buildSpace(
       if (!this.disabled) {
         return false
       }
-      if (ability?.can('update-all', 'Drive')) {
+      if (ability?.can('update-all', 'Drive') || ability?.can('create-all', 'Drive')) {
         return true
       }
       // FIXME: server permissions are a mess currently: https://github.com/opencloud-eu/opencloud/issues/10
--- a/packages/web-client/src/helpers/space/functions.ts
+++ b/packages/web-client/src/helpers/space/functions.ts
@@ -221,14 +221,14 @@ export function buildSpace(
       return this.graphPermissions?.includes(GraphSharePermission.deletePermissions)
     },
     canRename: function ({ ability }: { user?: User; ability?: Ability } = {}) {
-      if (ability?.can('update-all', 'Drive') || ability?.can('create-all', 'Drive')) {
+      if (ability?.can('update-all', 'Drive')) {
         return true
       }
       // FIXME: server permissions are a mess currently: https://github.com/opencloud-eu/opencloud/issues/10
       return this.graphPermissions?.includes(GraphSharePermission.deletePermissions)
     },
     canEditDescription: function ({ ability }: { user?: User; ability?: Ability } = {}) {
-      if (ability?.can('update-all', 'Drive') || ability?.can('create-all', 'Drive')) {
+      if (ability?.can('update-all', 'Drive')) {
         return true
       }
       // FIXME: server permissions are a mess currently: https://github.com/opencloud-eu/opencloud/issues/10
--- a/packages/web-client/tests/unit/helpers/space/functions.spec.ts
+++ b/packages/web-client/tests/unit/helpers/space/functions.spec.ts
@@ -34,33 +34,39 @@ describe('buildSpace', () => {
   describe('canBeDeleted', () => {
     it.each([
       {
-        userCan: false,
+        abilities: [],
         permissions: [GraphSharePermission.deletePermissions],
         disabled: true,
         expectedResult: true
       },
       {
-        userCan: false,
+        abilities: [],
         permissions: [],
         disabled: true,
         expectedResult: false
       },
       {
-        userCan: true,
+        abilities: ['delete-all'],
         permissions: [],
         disabled: true,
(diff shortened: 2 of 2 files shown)


JammingBen:
> Restoring or deleting a space as manager of that space with the User role is still not possible.

I'm not 100% sure, but if I remember correctly (and according to [the docs](https://docs.opencloud.eu/docs/user/spaces/disable-enable)) it shouldn't be. Disabling, restoring & deleting spaces is for administrators and space admins, not space managers. In that case this would be an API issue, since it currently allows all these operations for managers on their assigned spaces. There is no such thing as an "owner" of a project space, I believe that's why we went with that decision. I might be wrong though... @tbsbdr @micbar do you remember? 

Anyway, you stated that this PR is not tackling this issue, so to be honest, I'm having a hard time understanding what it actually solves for you. Is it related to the custom role you mentioned in https://github.com/opencloud-eu/opencloud/issues/1799?

tiran133:
When I was writing owner I meant manager, a space member with `can manage` permission.
I'm aware there are no ownerships of a space.

About what the docs say it is a bit wired since currently you are allowed to disable any space are a member of with permission `can manage` regardless of User role.
Either via API or UI if that's not intended as the docs say it's a serious issue. 
I confirmed this on the demo site. I disabled  "Webinars and Talks" as Alana

I then can even delete/enable it straight after. But as soon as I refresh the page I am now unable to enable/delete it again via UI still be able to enable/delete via API.

image

\
\
Yes, it would fix an issue we have with our custom role. Where we took the User Role and added the `Drives.Create` Permission and called it `spaceoperator`
This way we can create users which are allowed to create spaces but don't see the admin menu to manage all space.
Just their “own” they get added as `can manage` to that created space.

That's what the original issue was about.

But yes in essence the issue where you should or shouldn't (I guess this question has to be answered first) be able to disable/enable/delete as a member with `can manage` permission still persists and has nothing to do with our custom role.
When we opened the other issue in Nov. last year I did not dig into this as deep.
Now I can see it is a general problem with the `can manage` permission.

We also could add `Drives.ReadWriteEnabled` to the custom role, but this is a more powerful role which allows to disable/enable/delete for all spaces even though the User doesn't have the `can manage` permission. More like an admin permission.
\
\
I can see why you hesitant to address the issue like this when it comes to permissions since this would probably just solve our needs nothing else in general and probably is just a workaround.
\
But the original issue is over 7 month old and wasn't addressed properly.

Let's use this opportunity to address this now?

I think it first it needs to be answered 
Should or shouldn't a `can manage` member be allowed to `disable/enable/delete`

Based on that it is either a serious issue with the permission model or just a UI issue.

And it should not lay there for 7 months.

Anyway I hope we can find some solution here and I am happy to help where I can.

JammingBen:
> I think it first it needs to be answered
> Should or shouldn't a can manage member be allowed to disable/enable/delete

Yes I agree. I'll check in with the team to get an answer to that, then we can decide how to move forward with this. Sorry for the long wait!

tbsbdr:
needs clarification what the expected behaviour is

tiran133:
To make it easy to understand and to not have to deal with a custom role. Ignore the custom role part.
It's not the issue here.
Also, I'm not even sure if this is the right place to discuss this.
I was just tired of waiting and tried to fix it myself opening this PR but then realising it is an underlying problem which has nothing to do with the custom role here.

---- 

Unfortunately that is not that easy to describe the expected behaviour. As there is one fundamental question to answer first.

As already multiple times described in here and here https://github.com/opencloud-eu/opencloud/issues/1799
there is an inconsistency between the UI and the API.

I'm talking about a user (User role) which has the `Can Manage` permission in any space.
image

This user is allowed to disable the space. Any user with `Can Manage`  is allowed to do this.
This user however cannot enable or delete the space via the UI.
After disabling the space and refreshing the page the two menu items are missing.
image
The refresh is key here without a refresh the user can `enable/delete` the space

After the refresh the user is still able to do this actions `enable/delete` via API.

You can try it on https://demo.opencloud.eu 
Just login as Alan.



Two Option.

Option 1
The `can manage` permission allows `disable/enable/delete` **This is what currently is reflected in the API.**

Expectation:
The UI should show `enable/delete` options after a page refresh.

Option 2
The `can manage` permission should not allow `disable/enable/delete` of a space 
And these actions are reserved for `Space admin` and `admin` according to the docs see above.

Expectation:
Fix the API to not allow these actions and also remove the disable menu item from the UI.
\
\
To voice a preference here. Option 1 is very much preferred here. As this would allow the user to self-manage the spaces.
At least this is what our use case requires.

----

Clarification on this PR!

As stated above this would fix the menu Items with our custom role. It would allow the user with that role to perform these actions, however, this would be a workaround and not a proper fix as also stated above.

The fix would be to have consistent behaviour with the `Can Manage` permission regardless which role the user has.

tiran133:
So Is there any update on this issue?

tiran133:
@tbsbdr @JammingBen
It's been 3 weeks now, and I have not heard anything from this. Is there anything being worked on?
Feels like you are repeating the same you did in https://github.com/opencloud-eu/opencloud/issues/1799
I would really appreciate it if this not sit here for 7 months like the other issue.
I honestly think this is important.

I'm happy to help, but no communication makes it really hard.

micbar:
@tiran133 Sorry for the long time period. This ticket is difficult to understand and it seems that different product views are maybe colliding.

@tbsbdr @JammingBen This needs a deep read and maybe a call to talk about it with @tiran133 

I think we should settle it quickly.

JammingBen:
Okay so we clarified the desired behavior:

- Deleting spaces should only be possible for users with the role "Space Admin", not for managers of a space. This is a final delete without any trash bin involved, hence only users that can create spaces should be able to delete them (= Space Admins).
- Disabling & enabling spaces should be possible for managers as well as Space Admins.

That means there is a mismatch between UI and API that we need to solve. API needs to disallow deleting spaces for managers. UI needs to reflect the situation properly (before and after reloading the page).

JammingBen:
I'm gonna close here, since the bug has been fixed and the feature (https://github.com/opencloud-eu/opencloud/issues/2962) like needs a different approach. Thanks for digging into it and bringing it to our attention!

PR #2492, opened 2026-05-11
Fix mail validation layout issue

tammi-23:
This fix removes an unnecessary extra line in the mail validation, which caused the design to break.

### DIFF
--- a/packages/web-app-contacts/src/components/ContactCreateForm.vue
+++ b/packages/web-app-contacts/src/components/ContactCreateForm.vue
@@ -17,7 +17,6 @@
       :error-message="
         isContactEmailInvalid(modelValue.email) ? $gettext('Please enter a valid email') : ''
       "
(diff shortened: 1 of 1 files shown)


packages/web-app-contacts/src/components/ContactCreateForm.vue
```diff
@@ -17,7 +17,7 @@
       :error-message="
         isContactEmailInvalid(modelValue.email) ? $gettext('Please enter a valid email') : ''
       "
-      :fix-message-line="true"
+      :fix-message-line="false"
```
AlexAndBear: you can just remove this line, as it defaults to false

AlexAndBear (changes requested):

AlexAndBear:
Please also give this PR a meaningful title, as it lands in the release notes

AlexAndBear (approved):

Issue #2210, opened 2026-03-24
Modal header breaks into 2 lines


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

The modal header breaks into 2 lines if too long.

### Reproduction Steps

- create a file with a very long name
- click the rename pencil next to the name

### Expected Outcome

The name gets truncated, the "x" button stays on the very right side.

### Actual Outcome

The "x" button gets pushed into a second line.

<img width="541" height="251" alt="Image" src="https://github.com/user-attachments/assets/bb7a4ae0-2621-42df-ab05-35867e13f22c" />

PR #2495, opened 2026-05-11
fix: prevent modal title actions from wrapping

tammi-23:
This fixes "Modal header breaks into 2 lines"
 https://github.com/opencloud-eu/web/issues/2210


AlexAndBear (approved):
👍

PR #2513, opened 2026-05-16
fix: skip upload conflict check when uploads target a non-current folder

dschmidt:
## Description

`UploadResourceConflict.getConflicts` compares uploads against `resourcesStore.resources`, the resources of the user's **current folder**. When a caller redirects an upload via `setUploadFolder` (e.g. the unzip extension extracting into a freshly created folder), the actual destination is *not* the current folder, so that check produces false positives whenever the upload's first path segment happens to match a resource in the current folder.

Concrete example: extracting `data.zip` (which contains a top-level `data/` directory) in the user's personal root. The unzip extension creates `<personal>/data/` as the extraction target via `setUploadFolder`. Every extracted file has `relativePath = "/data/..."`, so `getConflicts` takes the first segment (`data`), finds the freshly created `data` folder in the current folder's resources, and pops up a *"Folder already exists"* dialog, even though the real upload target is `<personal>/data/data/...` and nothing actually collides.

This PR adds an `isCurrentFolder(uploadFolder)` helper that compares `storageId + path` of the upload target against `resourcesStore.currentFolder`, and only runs the conflict check when they match.

### Known limitation, follow-up needed

This change **suppresses the false positive** but does **not** actually perform conflict detection for non-current target folders. The current `getConflicts` is synchronous and depends on `resourcesStore.resources`, which only ever holds the current folder's contents.

A real conflict check for a non-current upload target would need to:
- async-list the target folder over webdav, and
- reshape `getConflicts` / call site to handle the async result.

That's out of scope for this PR. For now, **callers that use `setUploadFolder` own their own conflict resolution**. The unzip extension already does this (it deduplicates the archive-name folder via `resolveFileNameDuplicate` before creating it). Other future callers should be aware.

## Related Issue

Surfaced while fixing https://github.com/opencloud-eu/web-extensions/issues/293. Extracting a zip that wraps its content in a top-level folder matching the archive name triggered the false-positive dialog.

## Types of changes

- [x] Bugfix

### DIFF
--- a/packages/web-app-files/src/HandleUpload.ts
+++ b/packages/web-app-files/src/HandleUpload.ts
@@ -105,6 +105,16 @@ export class HandleUpload extends BasePlugin<PluginOpts, OcUppyMeta, OcUppyBody>
     return this.resourcesStore.currentFolder
   }
 
+  isCurrentFolder(uploadFolder?: Resource): boolean {
+    const currentFolder = this.resourcesStore.currentFolder
+    if (!uploadFolder || !currentFolder) {
+      return false
+    }
+    return (
+      uploadFolder.storageId === currentFolder.storageId && uploadFolder.path === currentFolder.path
+    )
+  }
+
   /**
    * Converts the input files type UppyResources and updates the uppy upload queue
    */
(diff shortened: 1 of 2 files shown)


JammingBen (approved):
Good catch 👍

Issue #2018, opened 2026-02-20
Some event listeners don't get cleaned up


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

The following event listeners don't get cleaned up an could cause a potential memory leak:

`wheel` in `packages/web-app-preview/src/components/Sources/MediaImage.vue`
`change` in `packages/web-pkg/src/services/uppy/uppyService.ts`

### Reproduction Steps

- add dev logs to track the registering and unregistering of those listeners
- open the preview app and close it
- open the "New"-menu and close it

### Expected Outcome

The registered event listeners should be unregistered.

### Actual Outcome

They stay registered.

JammingBen:
Fixed by https://github.com/opencloud-eu/web/pull/2532

PR #2532, opened 2026-05-18
fix: cleanup event listener

tammi-23:
This is a fix for event listeners don´t get cleaned up https://github.com/opencloud-eu/web/issues/2018

### DIFF
--- a/packages/web-pkg/src/services/uppy/uppyService.ts
+++ b/packages/web-pkg/src/services/uppy/uppyService.ts
@@ -300,6 +300,10 @@ export class UppyService {
       this.uploadInputListeners.delete(el)
     }
 
+    this.unregisterUploadInputElement(el)
+  }
+
+  unregisterUploadInputElement(el: HTMLInputElement) {
     this.uploadInputs = this.uploadInputs.filter((input) => input !== el)
   }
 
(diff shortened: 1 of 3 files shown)


JammingBen (changes requested):
Something broke the upload when clicking the "New" button and then selecting a file via the "Files Upload" action.

packages/web-pkg/src/services/uppy/uppyService.ts
```diff
@@ -277,19 +278,32 @@ export class UppyService {
...
+      const files = Array.from(target.files || [])
+      this.addFiles(files)
     }
+
+    this.uploadInputListeners.set(el, listener)
+    el.addEventListener('change', listener)
+    this.uploadInputs.push(el)
   }
 
   removeUploadInput(el: HTMLInputElement) {
```
JammingBen, line 296: It works now, but the `removeUploadInput` method now never gets called. Isn't there a way to make this work like you had it previously, by having only `removeUploadInput` that handles everything?

packages/web-pkg/src/services/uppy/uppyService.ts
```diff
@@ -277,16 +278,20 @@ export class UppyService {
...
+    if (this.uploadInputListeners.has(el)) {
+      return
+    }
+
+    const listener: EventListener = (event) => {
+      const target = event.target as HTMLInputElement
+      const files = Array.from(target.files || [])
+      this.addFiles(files)
+      this.uploadInputListeners.delete(el)
     }
```
JammingBen: Now it works 🚀  I guess simply adding `{ once: true }` is enough. Can we remove `uploadInputListeners` as well? Basically just have everything as it was before, but with `{ once: true }`.

JammingBen (approved):

PR #2556, opened 2026-05-19
Collaborative markdown

micbar:
## Description

Add an Y.js server to enable the collaborative mode of tiptap.

## Details

- [x] Adds a hocuspocus container (Should be maintained in a separate repo)
- [x] Integrates it into the DEV Stack
- [x] Implements A WebDAV Storage Layer for Hocuspocus
- [x] Uses a webhook config to connect the tiptap editor with the hocuspocus server
- [x] Uses a Tiptap Editor instance on the Hocuspocus Server inside a JSDOM Tree to consolidate the write process to the storage in one location. This avoids that all editor instances need to synchronize and elect a leader who stores the data in markdown. This added layer is needed because of our markdown implementation, because we cannot sync the markdown between the different editor instances. Using an Editor Instance on the hocuspocus server makes the markdown conversion and storage more reliable and keeps the logic in hocuspocus instead of spreading it to all editor instances
- [x] Discovers the Token validation URL via openIDConnect Discovery to validate its access tokens
- [x] Uses the existing token from the pinia Store
- [x] Handles Write Conflicts to the markdown file and notifies the enduser by leveraging the WebDAV "If-Match" logic and by tracking etags
- [x] Implements a random color finder function for the user visualisations (called "collaboration-carets")

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/dev/docker/hocuspocus/.dockerignore
+++ b/dev/docker/hocuspocus/.dockerignore
@@ -1,0 +1,4 @@
+node_modules
+npm-debug.log
+.DS_Store
+*.log
--- a/dev/docker/hocuspocus/Dockerfile
+++ b/dev/docker/hocuspocus/Dockerfile
@@ -1,0 +1,12 @@
+FROM node:22-alpine AS deps
+WORKDIR /app
+COPY package.json ./
+RUN npm install --omit=optional --no-audit --no-fund
+
+FROM node:22-alpine
+WORKDIR /app
+COPY --from=deps /app/node_modules ./node_modules
+COPY package.json tsconfig.json server.ts ./
+ENV NODE_ENV=production
+EXPOSE 9400
+CMD ["npx", "tsx", "server.ts"]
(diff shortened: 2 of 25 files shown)


micbar:
@JammingBen 
There seems to be a bug in my implementation.

Something regarding Token Handling or reconnects.

When my browser Tab is inactive for some time, the Editor Content is appended several times to the end of the document again and again.

micbar:
@dschmidt 
I am interested how your implementation differs from this.

dschmidt:
Quite a bit 😅 

The main difference is that I tried to keep the server as dumb as possible. I want it to do basic authentication and authorization and apart from that I want it to be universally usable by different yjs apps.
That's why most of the logic in my approach is on the client side.

- https://github.com/opencloud-eu/web/pull/2561

I've tried to keep all connection handling, stale detection, client mismatch handling and all that in a reusable generic layer, so actual app implementation can stay very lean.

It's very much still work in progress too - but if you wanna take a first look ... feel free! :) 

Happy to discuss the overall architecture.

FWIW I'm still online in Matrix 😅

micbar:
closing in favour of https://github.com/opencloud-eu/web/pull/2561

PR #2573, opened 2026-05-21
fix(editor): json rendering, plaintext serialization

JammingBen:
Fixes a few issues with the editor:

1. The text editor is not supposed to render tiptap json from any json file. This would mean that the editor doesn't display json files properly. The functionality for rendering tiptap json (= richtext format) is reserved for dedicated apps like the notes app.

2. Fix error when saving and closing plaintext files.

3. Prevent the duplicated blank lines when saving and reloading plaintext files.


tammi-23 (approved):
All 3 issues are fixed 👍

PR #2588, opened 2026-05-22
fix: prevent components from being proxied in Vue's reactivity system

JammingBen:
Vue warns when a component object is made reactive, which can cause performance overhead and interfere with component identity checks. Components passed as `content` to extension registrations and as `customComponent` to `dispatchModal` calls were missing `markRaw`, causing them to be inadvertently proxied by Vue's reactivity system.


AlexAndBear (commented):
👍

kulmann (approved):

PR #2589, opened 2026-05-22
feat: redesign batch actions

### LINKED ISSUE #2022: Batch actions
### Initial Checklist

### User Story

> * As a user, I want to have a consistent location where batch actions appear, so that I know where to find them

### User Value

UX

### Acceptance Criteria

- when files are selected, a batch actions bar appears and completely replaces the "FilterBar" (see https://github.com/opencloud-eu/web/issues/2021 )
- style the batch actions bar like in the attached mockup

BLOCKING: we have a batch action for "Empty trash", which gets displayed when no files are selected. This won't be possible anymore. We need a solution for that.

Image

JammingBen:
Changes the appearance of the batch actions to match the redesign mocks.

This also tidies up the implementation a bit by removing the `CreateAndUpload` component:

- The relevant upload-related parts don't need a template, so they have been moved to a composable that gets invoked by `GenericSpace`.
- The clipboard relevant actions are implemented as proper file actions that are rendered via the existing batch actions extension point.

image

closes https://github.com/opencloud-eu/web/issues/2022


tbsbdr:
i like! could you please double-check with the design - I think we need a bit more whitespace within the bar

image

AlexAndBear:
Hm, somehow I like the original more then the redesign, it doesn't look visually appealing to me. 

Anyways, can we apply the deselect click handler around the x icon and the text, so it's easier to reach ?

AlexAndBear (approved):

PR #2599, opened 2026-05-27
feat: use embed mode for copy and move actions

AlexAndBear:
## Related Issue

- Fixes https://github.com/opencloud-eu/internal/issues/285

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix


AlexAndBear:
When we create a space from within the embed mode and want to copy move a file to this space, the actions fails.
This is because we actually don't have the space is the store. We need sse events for spaces to get this working:

https://github.com/opencloud-eu/opencloud/issues/2844

packages/web-runtime/src/container/sse/spaces.ts
```diff
@@ -0,0 +1,73 @@
...
+    return
+  }
+
+  const space = await clientService.graphAuthenticated.drives.getDrive(sseData.itemid)
+  spacesStore.upsertSpace(space)
+  await spacesStore.loadGraphPermissions({
+    ids: [space.id],
+    graphClient: clientService.graphAuthenticated,
+    useCache: false
+  })
```
JammingBen: Is it really necessary to load the permissions here? Normally it should be enough to do this lazy when selecting the space, since it's expensive.

packages/web-app-files/src/composables/actions/files/useFileActionsCopy.ts
```diff
@@ -1,41 +1,175 @@
...
+      return
+    }
+
+    resourcesStore.setSelection(resources.map(({ id }) => id))
+    const parentFolderLink = getParentFolderLink(resources[0])
+
+    dispatchModal({
+      elementClass: 'location-picker-modal',
+      title: $gettext('Copy to'),
+      customComponent: LocationPickerModal,
```
JammingBen: To prevent it from being proxied.

```suggestion
      customComponent: markRaw(LocationPickerModal),
```

packages/web-app-files/src/composables/actions/files/useFileActionsMove.ts
```diff
@@ -0,0 +1,243 @@
...
+      return
+    }
+
+    resourcesStore.setSelection(resources.map(({ id }) => id))
+    const parentFolderLink = getParentFolderLink(resources[0])
+
+    dispatchModal({
+      elementClass: 'location-picker-modal',
+      title: $gettext('Move to'),
+      customComponent: LocationPickerModal,
```
JammingBen:
```suggestion
      customComponent: markRaw(LocationPickerModal),
```

packages/web-app-files/src/composables/actions/files/useFileActionsCopy.ts
```diff
@@ -1,41 +1,175 @@
...
+    targetFolder,
+    sourceSpace,
+    resources
+  }: {
+    targetSpace: SpaceResource
+    targetFolder: Resource
+    sourceSpace: SpaceResource
+    resources: Resource[]
+  }) => {
+    const targetFolderRef = computed(() => targetFolder)
```
JammingBen: Defining a `computed` inside a method feels a bit finicky. `ref(targetFolder)` would be the better fit, since you don't need any computed reactivity here.

packages/web-app-files/src/composables/actions/files/useFileActionsMove.ts
```diff
@@ -0,0 +1,243 @@
...
+    targetFolder,
+    sourceSpace,
+    resources
+  }: {
+    targetSpace: SpaceResource
+    targetFolder: Resource
+    sourceSpace: SpaceResource
+    resources: Resource[]
+  }) => {
+    const targetFolderRef = computed(() => targetFolder)
```
JammingBen: Same here.

packages/web-pkg/src/components/Modals/LocationPickerModal.vue
```diff
@@ -0,0 +1,129 @@
...
+      class="size-full"
+      :title="iframeTitle"
+      :src="iframeSrc"
+      tabindex="0"
+      @load="onLoad"
+    ></iframe>
+  </div>
+</template>
+
+<script lang="ts">
```
JammingBen: Script setup?

packages/web-runtime/src/container/sse/spaces.ts
AlexAndBear: no

packages/web-pkg/src/components/Modals/LocationPickerModal.vue
AlexAndBear: yep this was a copy of FilePickerModal

JammingBen (approved):
LGTM

PR #2601, opened 2026-05-27
feat: implement user mentions in Collabora docs

JammingBen:
Users can be mentioned with `@` in Collabora documents, triggering a notification to the mentioned user. Also refactors all Collabora postMessage stuff and the task for loading shares to dedicated composables.

Note this this is still missing the actual notification parts from the server. But it doesn't hurt to have the functionality in Web before that.

### DIFF
--- a/packages/web-app-external/src/composables/useCollaboraPostMessages.ts
+++ b/packages/web-app-external/src/composables/useCollaboraPostMessages.ts
@@ -1,4 +1,4 @@
-import { onMounted, onBeforeUnmount, ref, unref, type Ref } from 'vue'
+import { onMounted, onBeforeUnmount, ref, unref, type Ref, markRaw } from 'vue'
 import { useGettext } from 'vue3-gettext'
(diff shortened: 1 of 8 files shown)


packages/web-app-external/src/composables/useCollaboraPostMessages.ts
```diff
@@ -0,0 +1,378 @@
...
+    if (unref(userIdsToMention).length === 0) {
+      return
+    }
+
+    try {
+      await httpAuthenticated.post(urlJoin(configStore.serverUrl, 'collaboration/notify'), {
+        fileID: unref(resource).fileId,
+        userIDs: unref(userIdsToMention),
+        type: 'mention'
+      })
```
JammingBen, line 309: @fschade JFYI this is the notification POST request I'm sending to the server as we discussed.

kulmann (approved):
LGTM 🥇

PR #2618, opened 2026-05-29
fix: prevent loosing mailcompose state

tammi-23:
Prevents loss of unsaved email drafts when navigating away from the mail compose widget or closing the browser.

### DIFF
--- a/packages/web-app-mail/src/components/MailWidget.vue
+++ b/packages/web-app-mail/src/components/MailWidget.vue
@@ -369,7 +369,7 @@ const requestUnsavedComposeSave = ({
 }) => {
   dispatchModal({
     title: $gettext('Unsaved changes'),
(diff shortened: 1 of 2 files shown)


packages/web-app-mail/src/components/MailWidget.vue
```diff
@@ -346,6 +360,23 @@ const doClose = () => {
...
+const requestUnsavedComposeSave = ({
+  onCancel,
+  onSave
+}: {
+  onCancel: () => void
+  onSave: () => void | Promise<void>
+}) => {
+  dispatchModal({
+    title: $gettext('Unsaved changes'),
+    message: $gettext('Your email isn’t finished yet. Save it as a draft before leaving.'),
```
kulmann:
```suggestion
    message: $gettext('Your email isn’t finished yet. Save it as draft before leaving.'),
```

kulmann (approved):
Nice 👍

PR #2649, opened 2026-06-09
fix: delete space action showing for space managers

JammingBen:
When a space manager, that is not a (space) admin, disables a space, the UI was previously showing the delete action. While the API should prevent that, the UI should also not show the action.

Fixes this issue by reloading the space permissions after disabling or enabling a space.

However, there are still questions that need clarification in the future:

- The server currently doesn't allow loading permissions for disabled spaces, hence the client defaults to no permissions. Should this be possible? For a solution like described in https://github.com/opencloud-eu/opencloud/issues/1799#issuecomment-4648116407, this needs to be possible.
- The client still has a lot of implicit permission checks in https://github.com/opencloud-eu/web/blob/main/packages/web-client/src/helpers/space/functions.ts#L220 for space actions. This is quite messy and should be cleaned up at some point. Also see https://github.com/opencloud-eu/opencloud/issues/10.

refs https://github.com/opencloud-eu/opencloud/issues/1799

### DIFF
--- a/packages/web-app-files/src/composables/actions/spaces/useSpaceActionsRestore.ts
+++ b/packages/web-app-files/src/composables/actions/spaces/useSpaceActionsRestore.ts
@@ -45,20 +45,10 @@ export const useSpaceActionsRestore = () => {
             space.spaceQuota = updatedSpace.spaceQuota
           }
           spacesStore.upsertSpace(updatedSpace)
-          return true
+          return space.id
         })
     )
     const results = await Promise.allSettled(promises)
-
-    try {
-      await spacesStore.loadGraphPermissions({ ids: spaces.map((s) => s.id), graphClient: client })
-    } catch (error) {
-      console.error(error)
-      showErrorMessage({
-        title: $gettext('Failed to update permissions for enabled spaces'),
-        errors: [error]
-      })
-    }
 
     const succeeded = results.filter(isPromiseFulfilled)
     if (succeeded.length) {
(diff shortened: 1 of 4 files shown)


AlexAndBear (approved):

packages/web-app-files/src/composables/actions/spaces/useSpaceActionsRestore.ts
```diff
@@ -73,6 +71,11 @@ export const useSpaceActionsRestore = () => {
               succeeded.length,
               { spaceCount: succeeded.length.toString() }
             )
+      await spacesStore.loadGraphPermissions({
```
kulmann: eh... I'm not sure if I fully understand. There already is a `loadGraphPermissions` call a few lines above (currently line 54). Doesn't this call here mean we should remove the call on line 54 with the entire try/catch block?

packages/web-app-files/src/composables/actions/spaces/useSpaceActionsRestore.ts
```diff
@@ -73,6 +71,11 @@ export const useSpaceActionsRestore = () => {
...
+        ids: spaces.map((s) => s.id),
```
kulmann:
```suggestion
        ids: succeeded.map((s) => s.id),
```

packages/web-app-files/src/composables/actions/spaces/useSpaceActionsRestore.ts
kulmann: additionally, should this be wrapped in a try/catch?

kulmann (approved):

PR #2650, opened 2026-06-09
enable translation-sync

v-scharf:
replace `ctx.build.sender` -> `ctx.build.title`


v-scharf:
@kulmann I did't see your https://github.com/opencloud-eu/woodpecker-ci-config-service/pull/20
close it if not needed

PR #2693, opened 2026-06-15
fix: vault route guard for share spaces

JammingBen:
Fixes an issue where reloading a vault as a share receiver would not redirect the user to the unlock vault page.

This happened because share spaces get created via mount points, which are not yet loaded when the `setupVaultUnlockGuard` runs. Hence we need to trigger and await the loading of mount points, fetch the correct one for the given `shareId` and create a share space from that.


kulmann (approved):
LGTM 👍

PR #2700, opened 2026-06-16
fix: new menu entries for external apps in vaults

JammingBen:
Disable entries from external apps in the "New"-menu inside vaults. This feature currently isn't supported.

image


kulmann (approved):
LGTM 👍

PR #2704, opened 2026-06-16
fix: hide restore action in the version sidebar panel if the resource permission is missing

fschade:
## Description

check for permissions in the versions sidebar panel


JammingBen (approved):
LGTM, formatter need another run though.

PR #2726, opened 2026-06-21
fix: include registered extra props in PROPFIND request body

flash7777:
## Summary

`registerExtraProp()` registered properties for response parsing but `buildPropFindBody()` never included them in the PROPFIND request XML. Extensions using `registerExtraProp()` would never receive their custom properties because the server was never asked for them.

Replaces #2721 (which had unrelated files).

## Changes

| File | Change |
|---|---|
| `builders.ts` | Include registered extra props in PROPFIND XML body |

## Test plan

- [x] Extension registers extra prop via `registerExtraProp('oc:my.prop')`
- [x] PROPFIND request XML now includes `<oc:my.prop/>`
- [x] Server returns the property value in response
- [x] Existing props (etag, fileid etc.) still work

### DIFF
--- a/packages/web-client/src/webdav/client/builders.ts
+++ b/packages/web-client/src/webdav/client/builders.ts
@@ -42,6 +42,12 @@ export const buildPropFindBody = (
   }
 
   const object = properties.reduce((obj, item) => Object.assign(obj, { [item]: null }), {})
+  // Include extra props in the request so they appear in PROPFIND
+  for (const ep of extraProps) {
+    if (!(ep in object)) {
+      object[ep] = null
+    }
+  }
   const props = getNamespacedDavProps(object, extraProps)
 
   const xmlObj = {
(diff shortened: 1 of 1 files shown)


packages/web-client/src/webdav/client/builders.ts
```diff
@@ -42,6 +42,12 @@ export const buildPropFindBody = (
   }
 
   const object = properties.reduce((obj, item) => Object.assign(obj, { [item]: null }), {})
+  // Include extra props in the request so they appear in PROPFIND
+  for (const ep of extraProps) {
+    if (!(ep in object)) {
+      object[ep] = null
```
JammingBen, line 51: Type checks now complain because `object` is implicitly typed `any`. You should be able to fix this by typing `object` in the reduce function above: `.reduce<Record<string, unknown>>(...)`.

JammingBen (changes requested):

JammingBen (approved):
Code looks good, thanks for fixing!

Now the formatter complains in CI though. You can run `pnpm check:all` to ensure basic checks succeed (possibly for your other PRs as well).

PR #2727, opened 2026-06-21
feat: support checkbox type for extension preferences

flash7777:
## Summary

Extension points can now declare `userPreference.type: 'checkbox'` to render a checkbox toggle instead of a dropdown select on the account Extensions preferences page.

Replaces #2720 (which had unrelated files).

## Changes

| File | Change |
|---|---|
| `types.ts` | Add optional `type` field to `UserPreference` |
| `ExtensionPreference.vue` | Render checkbox when `type: 'checkbox'` |

## How extensions use it

```typescript
userPreference: {
  type: 'checkbox',
  optionLabel: 'Show file references'
}
```

The extension point should have exactly two extensions (enabled/disabled). The first represents the checked state.

## Test plan

- [x] Checkbox renders on Extensions preferences page
- [x] Toggle persists selection
- [x] Dropdown mode still works for existing extension points

### DIFF
--- a/packages/web-pkg/src/composables/piniaStores/extensionRegistry/types.ts
+++ b/packages/web-pkg/src/composables/piniaStores/extensionRegistry/types.ts
@@ -210,5 +210,6 @@ export type ExtensionPoint<T extends Extension> = {
   userPreference?: {
     label: string
     description?: string
+    type?: 'select' | 'checkbox'
   }
 }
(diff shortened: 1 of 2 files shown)


JammingBen (changes requested):
I'm not liking the checkbox solution, sorry. The main issue I'm having is that it kinda ties together two extensions on one extension point. As soon as some other app registers e.g. a third extension on that extension point, this concept breaks.

I think you can just use the existing select for your extension, no? It will then only have 2 options, which is totally fine though.

flash7777 (commented):
Thanks for the feedback! You are right — the checkbox approach ties two extensions together in a fragile way. Using the existing select with 2 options is cleaner and works fine for the use case. Closing this PR.

PR #2728, opened 2026-06-21
feat: add metadata sidebar panel

flash7777:
## Summary

New sidebar tab "Metadata" that displays custom metadata (`user.oc.md.*`) for the selected file or folder. Fetches data from the Graph API endpoint `GET /drives/{driveID}/items/{itemID}/metadata`.

Replaces #2709 (which had unrelated files).

## Changes

| File | Change |
|---|---|
| `MetadataPanel.vue` | New component: renders key-value pairs with formatted labels |
| `useFileSideBars.ts` | Register metadata panel, visible for single item selection |

## Features

- Strips `oy.` prefix for display
- Converts camelCase keys to Title Case
- Read-only display of all custom metadata

## Test plan

- [x] Set custom metadata via Graph Metadata API
- [x] Select file → Sidebar shows Metadata tab
- [x] Custom metadata values displayed correctly
- [x] Panel hidden when no metadata present

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Metadata/MetadataPanel.vue
+++ b/packages/web-app-files/src/components/SideBar/Metadata/MetadataPanel.vue
@@ -1,0 +1,96 @@
+<template>
+  <div id="files-sidebar-panel-metadata" class="rounded-sm p-4 bg-role-surface-container">
+    <div v-if="loading" class="flex justify-center">
+      <oc-spinner :aria-label="$gettext('Loading metadata')" />
+    </div>
+    <div v-else-if="error" class="text-role-on-surface-variant">
+      {{ $gettext('Could not load metadata.') }}
+    </div>
+    <div v-else-if="isEmpty" class="text-role-on-surface-variant">
+      {{ $gettext('No metadata available.') }}
+    </div>
+    <dl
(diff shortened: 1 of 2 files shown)


packages/web-app-files/src/components/SideBar/Metadata/MetadataPanel.vue
```diff
@@ -0,0 +1,96 @@
...
+      const res = unref(resource)
+      const sp = unref(space)
+      if (!res?.id || !sp?.id) return
+
+      loading.value = true
+      error.value = false
+      try {
+        const httpClient = clientService.httpAuthenticated
+        const response = await httpClient.get(
+          `/graph/v1.0/drives/${sp.id}/items/${res.id}/metadata`
```
kulmann, line 65: Where does this metadata endpoint come from? Is it your own custom thing? If yes, then this PR should not be targeting the core `web` repo, but should just be an extension that you keep in one of your own repos.

kulmann: ah, found it 🤔  https://github.com/opencloud-eu/opencloud/pull/2960

PR #2744, opened 2026-06-24
fix: provide npm token via .npmrc for pnpm 11 publish

JammingBen:
pnpm 11 rewrote publish from scratch and no longer applies the auth token injected via npm_config_* env vars, causing a 404 on publish of scoped packages. whoami still read the env var, which masked the problem. Write the token to a workspace-root .npmrc instead; the token is only written to the file (never echoed), the file is removed on exit via a trap, and the bash -c wrapper keeps it out of the command trace.


kulmann (approved):

JammingBen:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|stable-7.1|[](https://github.com/opencloud-eu/web/pull/2745)|

### Questions ?

PR #2768, opened 2026-06-30
fix: isolate module federation shared scope per app

JammingBen:
Load each external app through its own `ModuleFederation` instance instead of a single shared host instance. All remotes loaded through one instance would share the same `shareScopeMap` by reference, so a faulty app that corrupts that scope during init breaks every other app too. Per-instance isolation contains such failures to the offending app while preserving parallel loading and singleton identity of shared modules.

refs https://github.com/opencloud-eu/opencloud/issues/3043


dschmidt (approved):
💪🏻

kulmann (approved):

JammingBen:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|stable-7.1|[](https://github.com/opencloud-eu/web/pull/2770)|

### Questions ?

PR #2755, opened 2026-06-29
feat(yjs): collaborative editing framework

dschmidt:
> [!WARNING]
> **THIS IS A PoC EXPERIMENT. NOT READY FOR MERGING OR PRODUCTION.**
>
> Original PoC and architecture discussion: https://github.com/opencloud-eu/web-extensions/pull/447
> This is still evaluating the general architectural approach I've taken - interfaces, names, configs etc. are still subject to change
> Obviously I'm not trying to land two more text editor apps - they are just demonstrators the abstraction is usable for more than one app. They were my starting point before I ported the default text-editor app.

## Description

Migrates the realtime collaborative editing PoC from `opencloud-eu/web-extensions` (PR https://github.com/opencloud-eu/web-extensions/pull/447) into this repo.

- New `CollaborativeWrapper` in `@opencloud-eu/web-pkg` with Y.Doc + Awareness + optional Hocuspocus provider, app-version handshake, stale-state recovery, per-app room namespace, and a three-state `realtimeUrl` (derive from `configStore.serverUrl + /realtime`, explicit URL, or `null` for local-only).
- Two new editor apps as thin shells: `web-app-codemirror`, `web-app-tiptap`.
- `web-app-text-editor` refactored onto the wrapper. Toolbar, slash commands, content-type strategies, undo/redo all preserved; only load / save / dirty tracking goes through the wrapper. Strategies flip `StarterKit.undoRedo` to false so `yUndoPlugin` from `@tiptap/y-tiptap` takes over.
- AppWrapper now refetches the file on 412 / 409 save responses and retries with the fresh etag, recovering silently in the typical collab case (peer just saved, our Y.Doc already has their edits).
- Hocuspocus sidecar landed in `dev/docker/hocuspocus/` with a `docker-compose.yml` service entry.

Migration plan and known follow-ups live in `REALTIME_COLLAB_MIGRATION.md` at the repo root.

## Related Issue

- https://github.com/opencloud-eu/opencloud/issues/2542
- Upstream PoC: https://github.com/opencloud-eu/web-extensions/pull/447

## How Has This Been Tested?

- test environment: local docker compose stack (web compose with the new hocuspocus service) + `pnpm build:w`
- test case 1: 5/5 codemirror playwright specs against the running stack
- test case 2: 4/4 tiptap playwright specs
- test case 3: 8/8 realtime-sync integration suite against the hocuspocus sidecar directly
[...]

PR #2785, opened 2026-07-01
feat(yjs): collaborative editing framework

JammingBen:
## Description

Subset of https://github.com/opencloud-eu/web/pull/2755 without the example apps we don't want to have in our monorepo, but with some architectual changes, improvements and bug fixes.

This is quite a lot, I recommend reading `dev/docs/yjs.md` to get more information about the implementation.

Huge thanks to @dschmidt for the heavy lifting groundwork in the original PR!

## Setup

Pull the latest OpenCloud image and run a fresh `docker compose up -d`.

## How has this been tested

* Locally by myself
* Claude via headless browsers (see `dev/docs/yjs-manual-tests.md`)

This of obviously far from being battle tested. It will be especially interesting to see how it performs under load and with more than 2-3 concurrent peers.

## What next?

https://github.com/opencloud-eu/web/issues/3033 lists all the follow-ups and categorizes them into priorities.


dev/docs/yjs-manual-tests.md
```diff
@@ -0,0 +1,143 @@
+# Yjs - manual test cases
+
+Manual test pass for the `feat/realtime-collaboration` branch. Every case below
```
dschmidt, line 3: We likely shouldn't merge this wording to main :)

dschmidt: In general - shouldnt we put this into proper e2e instead of this awkward test plan?

dev/docs/yjs.md
```diff
@@ -0,0 +1,376 @@
...
+    HP -->|"3 . who are you?"| GRAPH
+    HP -->|"4 . may you write this?"| GRAPH
+
+    classDef client fill:#e8f0fe,stroke:#4285f4,color:#111
+    classDef server fill:#e6f4ea,stroke:#34a853,color:#111
+    classDef side fill:#fef7e0,stroke:#f9ab00,color:#111
+    class A,B client
+    class GRAPH,DAV server
+    class HP side
+```
```
dschmidt, line 38: Image

Client Aweb and Client Bweb as names looks odd

dev/docs/yjs.md
```diff
@@ -0,0 +1,376 @@
...
+
+**Every peer autosaves.** A remote edit marks _your_ `AppWrapper` dirty, so with N clients open the same document gets saved N times per autosave interval. There is election for hydration but none for saving.
+
+**Peer edits arm your unsaved-changes guard.** Same root cause: remote CRDT updates flow into `currentContent`, so `beforeunload` and the unsaved-changes modal fire for edits you did not make. Writers only - `isDirty` is hard-wired to `false` for a read-only client, which has nothing to save.
+
+**Hydration election can race.** The 150 ms awareness-settle window is heuristic. If awareness has not propagated in time, two peers can both elect themselves and hydrate, duplicating content. Server-side hydration would remove the whole class.
+
+**`_oc_meta` is writable by any client with write access.** A buggy or hostile client can set `appVersion` or `isStale` and lock or reset every peer in the room. The room's control plane has no server authority beyond the read-only gate.
+
+**`appVersion` is `0.0.0`.** `web-app-text-editor` is a private package with a placeholder version, so the version gate never actually fires for it.
```
dschmidt, line 360: Should we use the web version here instead?

packages/web-app-text-editor/src/App.vue
```diff
@@ -10,59 +10,45 @@
...
+import { detectContentType } from './yjs'
+
+// AppWrapper keeps its loading screen up until the Yjs session has
+// synced and hydrated, so `ydoc` and `awareness` are ready by the time this
+// component mounts. The editor binds to the shared Y.Doc through
+// `@tiptap/extension-collaboration` inside `useTextEditor`.
+// `currentContent` is declared by `YjsEditorSlotProps` but never read
+// here. It makes AppWrapper fetch the file, which the Yjs session
+// uses to seed an empty Y.Doc. Once the session is up the Y.Doc is the source
+// of truth.
```
dschmidt: This is useful while reviewing, but tbh I think the value having this inside the code is limited.
I would prefer to remove the whole block

packages/web-app-text-editor/src/index.ts
```diff
@@ -200,7 +202,8 @@ export default defineWebApplication({
       {
         path: '/:driveAliasAndItem(.*)?',
         component: AppWrapperRoute(TextEditor, {
-          applicationId: appId
+          applicationId: appId,
+          yjs: { appVersion: pkg.version, makeAdapter: makeTextEditorAdapter }
```
dschmidt: multiline maybe?

packages/web-app-text-editor/src/yjs.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+/**
+ * Builds the Y.Doc bridge the AppWrapper's Yjs session runs on.
+ *
+ * Called during the wrapper's setup, before the file is loaded, so the
+ * content type is resolved lazily, per call, from the resource ref. The
+ * strategies themselves must be built eagerly: they call `useGettext()`, which
+ * only works while a setup context is active.
+ */
```
dschmidt, line 28: Could be shorter I guess - the strategies/useGettext info is the most important bit I guess

packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
```diff
@@ -186,30 +202,19 @@ registerExtensions(appBarExtension)
 const { actions: saveAsActions } = useFileActionsSaveAs({ content: currentContent })
 
 const isEditor = computed(() => {
-  return Boolean(wrappedComponent.emits?.includes('update:currentContent'))
+  // A collaborative app drives its content through the Yjs session rather
+  // than through `update:currentContent`, so opting in makes it an editor.
```
dschmidt: 🤯

packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
```diff
@@ -722,18 +894,25 @@ onBeforeRouteLeave((_to, _from, next) => {
...
   space: unref(unref(currentFileContext).space),
   resource: unref(resource),
   activeFiles: unref(activeFiles),
   isDirty: unref(isDirty),
-  isReadOnly: unref(isReadOnly),
+  isReadOnly: unref(effectiveReadOnly),
   applicationConfig: unref(applicationConfig),
   currentFileContext: unref(currentFileContext),
-  currentContent: unref(currentContent),
+  currentContent: unref(currentContent) as string,
```
dschmidt, line 836: looks like a hack 🤓

packages/web-pkg/src/components/AppTemplates/types.ts
```diff
@@ -1,12 +1,102 @@
...
+ * below, or `Pick` the keys you need.
+ */
+export interface AppWrapperSlotProps {
   applicationConfig: AppConfigObject
+  space: SpaceResource
   resource: Resource
-  currentContent: Ref<string>
+  currentFileContext: FileContext
+  /** Fetching this costs a WebDAV GET of the whole file. */
+  currentContent: string
```
dschmidt, line 55: Is this correct? Cant we have different types depending on what the app defines how it wants content to be injected? Or is this just urls not content?

packages/web-pkg/src/editor/composables/strategies/plainText.ts
```diff
@@ -35,9 +38,10 @@ export const useStrategyPlainText = (editorState: TextEditorState): ContentTypeS
     }
   }
 
-  const extensions = (): Extension[] => {
+  const extensions = (options?: ExtensionsOptions): Extension[] => {
     return [
       StarterKit.configure({
+        undoRedo: options?.yjs ? false : undefined,
```
dschmidt, line 44: iirc there's a separate plugin that supports it - if we haven't integrated that within the scope of this pr, we should add it to the follow up issue

packages/web-pkg/src/editor/composables/useTextEditor.ts
```diff
@@ -10,12 +14,50 @@ import type {
...
+            label.setAttribute('style', `background-color: ${user.color ?? '#ffa500'}`)
+            label.insertBefore(document.createTextNode(user.name ?? ''), null)
+            cursor.insertBefore(label, null)
+            return cursor
+          }
+        })
+      ]
+    }
+  })
+}
```
dschmidt, line 56: It's only a few lines of code, so it's not too important - but is this something we should upstream somewhere? in `@tiptap/extension-collaboration` maybe?

packages/web-pkg/src/editor/types.ts
```diff
@@ -40,7 +71,8 @@ export interface TextEditorInstance {
   state: TextEditorState
   editor: ShallowRef<Editor | null>
   contentType: Ref<ContentType>
-  readonly: Ref<boolean>
+  /** Derived from the caller's `readonly` option; follows it while mounted. */
+  readonly: ComputedRef<boolean>
```
dschmidt: Why is it important for the interface it's computed? A ComputedRef is a Ref as well

services/yjs/src/server.ts
```diff
@@ -0,0 +1,300 @@
...
+        user: {
+          id,
+          displayName: 'Dev Fake User',
+          color: deterministicColor(id)
+        }
+      }
+    }
+
+    const me = await validateTokenAgainstOpenCloud(token)
+    const id = me.id ?? me.userPrincipalName ?? 'unknown'
```
dschmidt: maybe another fallback to email address?

dschmidt (commented):
basically: LGTM :) 

Just a bunch of nit picks here and there

dev/docs/yjs-manual-tests.md
JammingBen: I'd like to merge it until we have https://github.com/opencloud-eu/web/issues/3064, just to keep track of it. And it's a nice testing manual for agents until we have e2e tests.

dev/docs/yjs.md
JammingBen: Yes, already listed in the follow-ups.

packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
JammingBen: `currentContent` is typed `unknown` for... unknown reasons 😂 I'd rather not change it here without knowing the effect on other (external) apps.

packages/web-pkg/src/components/AppTemplates/types.ts
JammingBen: Yes, setting `currentContent` results in a WebDAV `GET` because of `if (unref(hasProp('currentContent')))` in the `AppWrapper`.

packages/web-pkg/src/editor/composables/strategies/plainText.ts
JammingBen: I think it's already part of the `Collaboration` extension, because undo/redo works.

kulmann (approved):
Wow, HUGE lift! 🥇 all I could find is listed in followups anyway. let's get this merged and iterate 🤩

PR #2793, opened 2026-07-02
feat(tiptap): add Tiptap toolbar support

tammi-23:
This PR adds toolbar variants to the shared Tiptab text editor. The initial usecase is supporting different toolbar layouts for different mail compose sizes.


JammingBen (commented):
I'm not the biggest fan of this solution because it seems to be heavily tailored to the mail app. What if some other app wants to display a different set of toolbar items on mobile?

I'd rather have a more generalized mechanism of controlling which actions to display. Maybe introduce a `actionsToDisplay` prop on the toolbar which acts as a whitelist and defaults to `undefined` = all actions? The downside would be that the consuming party would require knowledge about the action ids. But it would keep the whole thing more abstract and not tailored to a specific implementation.

packages/web-pkg/src/editor/components/TextEditorProvider.vue
```diff
@@ -5,12 +5,14 @@
...
 
 <script setup lang="ts">
-import { provide } from 'vue'
+import { provide, toRef } from 'vue'
 import type { TextEditorInstance } from '../types'
 
-const { editor } = defineProps<{
+const props = defineProps<{
   editor: TextEditorInstance
+  actionsToDisplay?: string[]
```
JammingBen: Wouldn't it make more sense to put this on the `TextEditorToolbar` component directly?

packages/web-pkg/src/editor/composables/strategies/html.ts
```diff
@@ -102,7 +110,13 @@ export const useStrategyHtml = (editorState: TextEditorState): ContentTypeStrate
     addColumnBefore,
     addColumnAfter,
     deleteColumn
-  } = useEditorActions(editorState)
+  } = useEditorActions(editorState, editorActionOptions)
+
+  const toolbarLink = () => ({
+    ...link(),
+    showInToolbar: true
+  })
```
JammingBen, line 134: I think you don't need this overwrite, since the `actionsToDisplay` ignore the `showInToolbar` setting anyways..? 🤔

JammingBen (approved):
LGTM! One final remark, although I'm not sure if it's accurate.

packages/web-pkg/src/editor/composables/strategies/html.ts
tammi-23: I will handle this in a folow-up US

PR #2803, opened 2026-07-05
fix: Prevent Enter during IME compose in submission dialogs

### LINKED ISSUE #2646: Japanese and other East Asian Languages IME input can prematurely submit forms
### Initial Checklist

### Bug Description

For East Asian languages, text input is generally done via IME (Input Method Editor) which converts text to asian characters. In this process, the return/enter key is used to select a conversion target per group of words.

In the case of Opencloud, the return/enter key will prematurely fire the form submit handler and leave the user with a potentially incomplete inpit. The solution is fairly simple as the `KeyboardEvent` already has a `isComposing` prop likely to address this issue.

Please see: https://dev.to/oikon/improving-japanese-input-ux-in-multilingual-applications-properly-handling-ime-conversion-2ild for background and solution snippet incorporating the `isComposing` prop for Vue.

### Reproduction Steps

Seems to affect MacOS - (could not reproduce in Win 11 but may depend on version windows/IME)

The easiest area to see this is Spaces:

1. Click on + New to create a space
2. With an IME input method on Japanese, start typing `tesuto` with the full desired name being `テストスペース` (Test Space)
3. IME should display recommended converstion to `テスト`  
4. Press Enter to accept conversion
5. Conversion is accepted but the Dialog submits before I can continue to the 2nd word 🐛 

### Expected Outcome

Expected IME conversion to happen without form submission

### Actual Outcome

Dialog prematurely submitted

neverether:
## Description

When typing with an IME (e.g. Japanese, Chinese, Korean), pressing Enter to confirm a character conversion immediately submitted dialogs. For example, in the "New Space" dialog, typing `tesuto` and pressing Enter to accept the テスト conversion created the space right away, before the user could finish typing the intended name.

This happens because the keydown that confirms an IME composition is delivered to the page as a regular Enter keydown on macOS (Windows IMEs swallow it, so the bug only manifests on macOS), and the Enter handlers never checked `KeyboardEvent.isComposing`.

Changes:

- Added an `isComposingEvent` helper to `web-pkg` that treats an event as part of a composition session when `isComposing` is set, or when `keyCode` is 229 for browsers that fire the composition-ending keydown with `isComposing` already unset (e.g. Safari).
- Guarded the Enter-to-confirm handlers with it:
  - `OcModal` input (covers New Space, New Folder, New File, Rename, Edit Description and every other `hasInput` modal) — check inlined here since the design system cannot import `web-pkg`
  - `SetLinkPasswordModal` — browsers suppress IMEs on `type="password"` inputs, but the field's "show password" toggle switches it to `type="text"`, where the IME is active and the same premature submit occurs
  - `FileNameModal` (external app)
  - `CreateShortcutModal` drop selection
- Added regression unit tests for the helper and the guarded components.

Before:
https://github.com/user-attachments/assets/674bca8a-09b0-4229-9e5c-2197616386d8

After:
https://github.com/user-attachments/assets/cb3badf9-cb47-4fc9-ac9e-66be34f2b286

**Note**: This PR also covers the Password Edit dialog – there may be a potential oversight in this area. When typing in a password with the password mask on (input type = password), IMEs and blocked and we're limited to safe alphanumeric password input. However when the mask is off (clicking the eyeball icon, the type = text), it's now possible to utilize the IME to input unicode. I'm not sure this is the intended policy and can cause passwords with unintended values/inputs. This might need addressing, but I'm uncertain of the intent here.

## Related Issue

- Fixes #2646

## How Has This Been Tested?

- test environment: local docker compose setup, macOS Chrome with the Japanese IME
- test case 1: "+ New" → Space, type `tesuto`, press Enter to confirm the テスト conversion → the dialog no longer submits; a subsequent plain Enter submits as before. Same for rename and other input dialogs.
- test case 2: new unit tests — Enter keydown with `isComposing: true` or `keyCode === 229` does not emit confirm, plain Enter does.

## Types of changes

- [x] Bugfix


packages/design-system/src/helpers/isComposingEvent.ts
```diff
@@ -0,0 +1,5 @@
+// Whether the event is part of an IME composition session, e.g. the Enter confirming a CJK
+// conversion. keyCode 229 covers browsers where `isComposing` is already unset (e.g. Safari).
+export const isComposingEvent = (event: KeyboardEvent) => {
+  return event.isComposing || event.keyCode === 229
+}
```
JammingBen, line 5: I'd rather put this helper into the design-system. It fits there are as well and it would remove the duplication, since you can use it from across the entire app.

JammingBen (changes requested):
Thanks for looking into this.

I guess the same is missing in `OcSearchBar.vue` and `SearchBar.vue`, or did you leave these out intentionally?

neverether:
@JammingBen Thanks for reviewing so quickly.

> I guess the same is missing in OcSearchBar.vue and SearchBar.vue, or did you leave these out intentionally?

I did spot those and the issue does manifest there as well; however in the case of the search bar the unintentional submit doesn't seem to have negative consequences - the focus is still in the input so the user can continue typing, it'll just return results early like a faux debounced search.

If you have a preference to keep it entirely consistent across those as well, I'll bundle it in.

JammingBen:
> I did spot those and the issue does manifest there as well; however in the case of the search bar the unintentional submit doesn't seem to have negative consequences - the focus is still in the input so the user can continue typing, it'll just return results early like a faux debounced search.

Sounds reasonable, I'm fine with keeping it as it is then.

neverether:
@JammingBen Have addressed your comment; moving the helper to `design-system` required an update to peerDeps for type resolution, let me know if that matches convention correctly.

JammingBen (approved):
Looks good, thanks again!

PR #2806, opened 2026-07-06
feat: remove unnecessary white spaces

JammingBen:
In the light of the current design refinements: Remove the white space on pages where we have filters and/or a search bar, since we can re-use the same space for the batch actions.

Most of these changes are basically moving filters and search bars from one location to another. Though I also included search bars on the "Shared with Others" and "Shared via link" pages for consistency reasons.


AlexAndBear:
Good stuff!

small nitpick:

In the `Shared with others` view, the search box in on the very left, while it should be typically on the right.

image

image

AlexAndBear (changes requested):
For consistency reasons, I think we should change the search label to 'Search for trash bins' as we also use this wording in the footer

Image

JammingBen:
Thx, fixed those 2!

AlexAndBear (approved):
🥇

PR #2814, opened 2026-07-07
refactor(files): components to script setup (#1)

JammingBen:
Refactor a batch of file-app components to script setup.


JammingBen:
Since this is a bit hard to review, I've let Claude review all the changes locally. I'd expect it to be pretty accurate with that, considering there is only one objectively right outcome.

AlexAndBear (approved):

PR #2843, opened 2026-07-08
fix: name sorting in files table is off when folder name contains a dot

### LINKED ISSUE #2842: [Bug] Dot in folder name sorts unexpected
## Describe the bug
When I have a folder with a dot in it, these are always sorted to the end. Is that expected behavior? Can be reproduced on demo.opencloud.eu

Image

AlexAndBear:
## Related Issue

- Fixes #2842 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/web-client/tests/unit/helpers/resource/functions.spec.ts
+++ b/packages/web-client/tests/unit/helpers/resource/functions.spec.ts
@@ -73,10 +73,20 @@ describe('filterResources', () => {
     })
     it.each([
       { name: 'afolder', isFolder: true },
-      { name: 'afolder', type: 'dir' },
-      { name: 'afolder', type: 'folder' }
+      { name: 'afolder', type: 'folder' },
+      { name: 'afolder', type: 'directory' }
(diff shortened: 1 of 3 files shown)


packages/web-client/src/helpers/resource/functions.ts
```diff
@@ -46,6 +46,10 @@ export const extractNameWithoutExtension = (resource?: Resource): string => {
 }
 
 export const extractExtensionFromFile = (resource: Resource): string => {
+  if (resource.isFolder || ['folder', 'directory'].includes(resource.type)) {
```
kulmann: this was deliberately changed to support apps being registered on "folder name extensions", e.g. notes or vault. need to find a different solution for sorting / must not rely on the extension being present as grouping criterion.

kulmann (changes requested):

kulmann (approved):

PR #2892, opened 2026-07-15
refactor: remove custom resource icon colors via theme.json

JammingBen:
Remove the option to provide custom colors to resource icons via the theme. This is not supported by our new resource icons, hence we decided to kill this completely. This means that we can also remove the `colorPalette` option from the `theme.json`.

As a result of this, also migrate the former default resource type icon colors to the SVGs directly via the `fill` attribute. This only affects the remaining old resource icons. We will replace them with new ones eventually.


packages/design-system/src/assets/icons/resource-type-drawio-fill.svg
kulmann, line 1: would it make sense to use new file type icons for the ones you colored?

JammingBen: Eventually, yes. I just didn't want to do it with this PR, but I can do it as a direct follow-up.

kulmann: sounds god 👍

kulmann (approved):

PR #2907, opened 2026-07-17
feat(admin-settings): enable name sorting for extension overview

AlexAndBear:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -72,6 +72,7 @@ Enforced via ESLint (`packages/eslint-config`). Run `pnpm lint` to check.
 - **TypeScript everywhere** – use types to catch bugs at compile time and support IDE tooling.
 - **Composables over services** – encapsulate reusable logic in composables. Avoid services.
 - **Early returns** – prefer early returns to keep code readable and performant.
+- **Arrow functions** – Prefer arrow functions for concise, modern, and readable code.
 - **Translations** – use `$gettext` (or its variants) for all user-facing strings. Don't manually edit the generated `l10n/translations.json` files inside a package.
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -72,6 +72,7 @@ Enforced via ESLint (`packages/eslint-config`). Run `pnpm lint` to check.
 - **TypeScript everywhere** – use types to catch bugs at compile time and support IDE tooling.
 - **Composables over services** – encapsulate reusable logic in composables. Avoid services.
 - **Early returns** – prefer early returns to keep code readable and performant.
+- **Arrow functions** – Prefer arrow functions for concise, modern, and readable code.
```
JammingBen: This conflicts with

> - **Functions vs arrow functions** – prefer named `function` declarations over arrow functions. Use arrow functions only for callbacks or when the `this` context requires it.

a few lines below. Best practice (and how most modern TypeScript projects handle it) is to use native `function` for functions, unless it's an inline-callback or the `this` scope requires it.

JammingBen (changes requested):

AGENTS.md
JammingBen: Though I'd argue that consistency beats it all at the end of the day for us. So maybe keep the existing bullet point and extend it by something like ("if the file is already using arrow functions, use that")

AlexAndBear: Hmm.. this seems to be a contradiction, I leave it as it is for now

JammingBen (approved):

PR #2909, opened 2026-07-17
chore: remove activities app

AlexAndBear:
## Description

The activities app never made it out an very early WIP state, therefore I think we shouldn't carry it around.
If we want to productize the app later we might want to reuse some of the relevant code, although it's quite outdated.

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix


packages/web-app-admin-settings/src/index.ts
```diff
@@ -206,7 +206,7 @@ export default defineWebApplication({
       name: $gettext('Admin Settings'),
       id: APPID,
       icon: 'settings-4',
-      color: '#2b2b2b'
+      color: '#000000'
```
kulmann: unrelated... is this wanted?

kulmann (approved):
Hm yes, I agree. Giving it a try was worth it though! I liked the idea very much.

packages/web-app-admin-settings/src/index.ts
AlexAndBear: totally, I updated the admin settings app icon in another pr (and used a different color) https://github.com/opencloud-eu/web/pull/2911

PR #2913, opened 2026-07-17
e2e-fix: a11y flaky in webkit

v-scharf:
- fix a11y test flaky in webkit
- increase timeout for a11y test using [tags](https://vitalets.github.io/playwright-bdd/#/writing-features/special-tags)


JammingBen (approved):
👍

PR #2915, opened 2026-07-17
Guest invite

### LINKED ISSUE #2824: [Guests] Invite guests by email address from the Sharing / Members panel
## User Story

> * As a user who owns or manages a resource (a file, a folder, or a Space), I want to invite external people by simply typing their email address into the Sharing or Members panel, so that I can collaborate with people who do not (yet) have an OpenCloud account, using the same flow I already use for internal users.

## User value

Guest sharing with better security than public links; seamless invite flow.

## Acceptance Criteria

- [ ] In the Sharing panel (file/folder) and the Members panel (Space), I can type a free-text value into the collaborator input.
- [ ] While I type, existing users and groups continue to be suggested exactly as they are today (no regression to the current sharee search).
- [ ] As soon as the typed value is a valid email address (validated against a standard email regex), the suggestion list additionally offers that email address as a selectable guest suggestion.
- [ ] The guest suggestion is visually distinguished from account suggestions by:
  - [ ] a globe icon, and
  - [ ] a "(Guest)" suffix on the suggestion label.
- [ ] If the typed value is not a valid email address, no guest suggestion is offered (no invalid entries can be selected).
- [ ] If the typed email address already belongs to a known account, the existing account is suggested as usual and is **not** duplicated as a guest suggestion.
- [ ] Clicking (or selecting via keyboard) the guest suggestion adds it to the invite input as a selected recipient chip, alongside any already-selected recipients.
- [ ] Any number of recipients can be added before submitting, in any combination of existing users, groups, and guests.
- [ ] A role can be assigned to the invite (and an expiration set) using the same controls used for internal shares.
- [ ] Submitting the invite triggers `invite` requests, where a guest has `@libre.graph.recipient.type: "mail"` and `objectId` is the email address
- [ ] The whole guest invite suggestion gets applied in both `internal` and `external` share modes
- [ ] After a successful submit, each invited guest email appears in the list of share receivers (Sharing panel) or Space members (Members panel).
- [ ] A guest entry in that list can be managed exactly like any other share:
  - [ ] change its role,
  - [ ] set / change / remove an expiration,
  - [ ] view its share metadata, and
  - [ ] remove the share entirely.
- [ ] On backend failure for the invite request, the user sees an error toast.

## Visuals

Image

Image

Image

Image

kulmann:
## Description

Adds guest invites from entered, valid email addresses.

## Related Issue

- Fixes https://github.com/opencloud-eu/web/issues/2824

## Types of changes

- [ ] Bugfix


packages/web-app-files/src/components/SideBar/Shares/Collaborators/InviteCollaborator/InviteCollaboratorForm.vue
```diff
@@ -320,34 +323,45 @@ const fetchRecipientsTask = useTask(function* (signal, query: string) {
...
+        client.groups.listGroups({ orderBy: ['displayName'], search: `"${query}"` }, { signal })
+      )
   const groups = (groupData || []).map((u) => ({
     ...u,
     shareType: ShareTypes.group.value
   })) as CollaboratorAutoCompleteItem[]
 
   const isSpace = !unref(resource) || isSpaceResource(unref(resource))
-  const guests = isSpace
+  const contacts = isSpace
```
kulmann: renaming OX addressbook users to `contacts` because our own email invite flow is supposed to be the `guests`.

AlexAndBear:
It looks like the icons in the invite box are black for guest users 
image

Also I am not a fan of globe icon in general, do we have other options ?

AlexAndBear:
image

The orange color of the globe icon is visually not very pleasing

AlexAndBear:
When I set the share type to external and add a guest user with the role `can edit`, the share will be created with the role `can view`, is this somehow expected ?

AlexAndBear:
Typing an email inside the Share with People box is very hard, when there is already an account present, without the auto-complete for guest users, it's even harder - nearly to "unsuable" :(

image

packages/web-app-files/src/components/SideBar/Shares/Collaborators/InviteCollaborator/InviteCollaboratorForm.vue
```diff
@@ -307,6 +311,7 @@ const createSharesConcurrentRequests = computed(() => {
 const {
   autocompleteResults,
   fetchRecipients,
+  fetchRecipientsTask,
```
JammingBen, line 314: This seems to be unused here. I see it getting referenced in the unit tests, can't you stub this globally there?

packages/web-app-files/src/components/SideBar/Shares/Collaborators/InviteCollaborator/InviteCollaboratorForm.vue
```diff
@@ -375,7 +397,26 @@ const share = async () => {
...
+            shareType === ShareTypes.group.value ? ShareTypes.group.key : ShareTypes.user.key
+        }
+
+    // guests are internal-style shares and must never receive a federated role, so fall back to
+    // the first internal role when a guest is invited from the external share mode
+    // FIXME: clean up this internal vs external share type selection mess :-(
+    let roleId = unref(selectedRole).id
+    if (isGuest && unref(isExternalShareRoleType)) {
+      roleId = unref(availableInternalRoles)[0]?.id
+    }
```
JammingBen, line 418: Maybe it's better to not offer guest suggestions at all when external is set? 🤔 This would also solve the role downgrade reported in https://github.com/opencloud-eu/web/pull/2915#issuecomment-5775535738.

packages/web-app-files/src/components/SideBar/Shares/Collaborators/InviteCollaborator/InviteCollaboratorForm.vue
```diff
@@ -319,25 +324,42 @@ const {
...
-
-    if (selected || exists) {
-      return false
+  const emailBelongsToAccount = (query: string) => {
+    return collaborators.some((c) =>
+      [c.mail?.toLowerCase(), c.onPremisesSamAccountName?.toLowerCase()].includes(
+        query.toLowerCase()
+      )
+    )
+  }
```
JammingBen, line 335: Hmm can we do something to make this check better? What I'm noticing:

Set `OC_SHOW_USER_EMAIL_IN_RESULTS: false`. Now as user alan, search for `admin@example.org`. It will suggest to add this user as guest, whereas it should be admin.

The server also responds with admin in the `/graph/v1.0/users` response, so we know about this. Can we somehow include this in the check here? I guess if the user enters an email address, and the server responds with exactly one user, then we can safely assume it belongs to the user, no matter the value of `c.mail` or `c.onPremisesSamAccountName`.

Issue #2841, opened 2026-07-08
[tiptap] Insert and edit Links


kulmann:
## User story

> * As a user I want to be able to insert and edit links in the text editor, so that I can link to external websites.

## User value

Text editing feature completeness.

## Acceptance criteria

- [ ] Inserting a link is possible for markdown, html and tiptap-json content strategies
- [ ] Pasting a link into the text editor automatically makes it a rendered link
- [ ] Clicking the link opens an inline link panel (see below)
- [ ] Clicking the link does NOT follow the link
- [ ] The toolbar has a link icon, which opens an inline link panel (see below)
- [ ] The slash command menu has a link entry, which opens an inline link panel (see below)
- [ ] The inline link panel has two text fields: url and text, a button for following the link and a button for unlinking the text (similar to how platejs is doing it)

PR #2930, opened 2026-07-21
feat: tiptap - Insert and edit Links

tammi-23:
Implements https://github.com/opencloud-eu/web/issues/2841
Insert and edit Links (tiptap)


packages/web-pkg/src/editor/helpers/link.ts
JammingBen, line 1: This doesn't look like a composable, but a set of helper functions. Maybe also put this in `packages/web-pkg/src/editor/extensions/link.ts`?

packages/web-pkg/src/editor/extensions/link.ts
```diff
@@ -0,0 +1,52 @@
+import Link from '@tiptap/extension-link'
+
+const allowedProtocols = new Set(['http:', 'https:', 'mailto:'])
```
JammingBen, line 4: This might be a bit too strict 🤔 I feel like we should at least allow fragment URLs to link to headlines (e.g. `#foo`). Also, since `isAllowedUri` is not only being used when creating links, but also in read-only mode when parsing & displaying them, a md file with `[text](#foo)` will now loose `#foo` in the link.

Out of scope here, but at some point we should also think about relative links to reference local documents.

JammingBen (commented):
When inserting a link, I can't really "exit" it without going into a new line. Platejs e.g. does this when typing a space, which feels a bit more natural. What do you think?

JammingBen (approved):
👍

Issue #2937, opened 2026-07-22
File selection is slow with many files


JammingBen:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

File selection is slow with many files in the files list.

### Reproduction Steps

- upload 100 files
- set the "Items per page" to 100
- select a file

### Expected Outcome

There is no noticeable delay when selecting.

### Actual Outcome

There is a noticeable delay.

PR #2939, opened 2026-07-22
perf: improve file selection in tiles view

JammingBen:
`ResourceTile` now derives its own selected state from the resources store instead of receiving it as a prop and exposes it to the selection slot as a scoped prop. This removes all `selectedIds` reads from `ResourceTiles`' render, so a selection toggle no longer re-renders the whole list.

Further improves selection performance by using a Set-based O(1) membership checks instead of `Array.includes`/`indexOf` in `useResourceViewHelpers` (`isResourceSelected`) and the resources store.

refs https://github.com/opencloud-eu/web/issues/2937

### DIFF
--- a/packages/web-pkg/src/components/FilesList/ResourceTable.vue
+++ b/packages/web-pkg/src/components/FilesList/ResourceTable.vue
@@ -21,7 +21,6 @@
     :drag-drop="dragDrop"
     :hover="hover"
     :item-dom-selector="resourceDomSelector"
-    :selection="selectedResources"
     :sort-by="sortBy"
     :sort-dir="sortDir"
     :lazy="lazy"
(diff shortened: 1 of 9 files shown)


packages/web-pkg/src/components/FilesList/ResourceTable.vue
```diff
@@ -21,7 +21,6 @@
     :drag-drop="dragDrop"
     :hover="hover"
     :item-dom-selector="resourceDomSelector"
-    :selection="selectedResources"
```
JammingBen, line 24: This prop doesn't exist on `OcTable`.

AlexAndBear (approved):
Didn't measure the improvement but altogether this makes totally sense 👍

PR #2940, opened 2026-07-22
perf: improve file selection in table view

### LINKED ISSUE #2937: File selection is slow with many files

JammingBen:
Basically the same as https://github.com/opencloud-eu/web/pull/2939 but for the table view. Unfortunately, the setup is a bit more complicated here because we have ResourceTable > OcTable (design-system) > Tr (design-system).

Selecting/deselecting resources in the table view re-rendered the whole list, replaying all per-row helpers (`getResourceLink`, `isVisible`) once per row. This gets quite expensive with >100 files.

- Extract `OcTableRow` so each row computes its own memoized highlight/disabled state. This ensures that a change of this state only re-renders the affected rows instead of the entire table.
- Read the selection from the store in `useResourceViewHelpers` instead of a `selectedIds` prop; drop the prop from `ResourceTable`/`ResourceTiles` and the `:selected-ids` input from the views. Keeping this input and changing it also causes a re-render of the entire component.
- Extract `ResourceTableSelectAll` leaf, use Set-based O(1) lookups for highlight- and disabled-checks (when still using the `highlighted` prop) and remove the dead `rowMounted` emit and checkbox outline handling.

closes https://github.com/opencloud-eu/web/issues/2937


AlexAndBear (approved):
🥇

PR #2951, opened 2026-07-24
fix: retry silent token renewal before logout on auth error

### LINKED ISSUE #3165: Web: a single failed silent renew permanently kills the session — no retry, no signinSilent-on-401
A transient network interruption (e.g., a VPN reconnect) at the moment silent renewal fires causes the web session to die permanently even though the IdP session remains valid.

Server logs show the tab's SSE stream auto-reconnecting with the expired access token (repeated 401s — `failed to verify access token: token has invalid claims: token is expired` — same request-id, over several minutes) with no renewal re-attempt, then the user is dropped to the "Not logged in" page. Clicking "Log in again" re-enters **without a credential prompt** — the IdP session was healthy throughout; only the client-side renewal path died.

**Expected behavior**: retry silent renewal with backoff on failure, and/or attempt `signinSilent` on the first 401 before surfacing the logout page.

**Version**: 7.3.0 (rolling). Standard `WEB_OIDC_*` config with `offline_access` scope and `WEB_OIDC_SILENT_REDIRECT_URL` wired.

JammingBen:
## The fix

Attempt a silent renewal with exponential backoff on any auth error (not only when the client considers the token expired) before logging the user out. A `401` from the server is authoritative even when the client still thinks its token is valid (e.g. clock skew after waking a device from sleep), so gating renewal on the client-side expiry flag could kill an otherwise-valid session on a transient interruption.

Also close the SSE stream on logout so it stops reconnecting with the removed access token.

## How to test

- Reduce the token expiration time to e.g. 60 seconds by setting `IDP_ACCESS_TOKEN_EXPIRATION: 60`
- Add the `offline_access` scope by setting `WEB_OIDC_SCOPE: openid profile email offline_access`
- Recreate the OpenCloud container
- Login with Web
- Put your computer in standby
- Wait a few minutes
- Wake up from standby and use Web (keep navigating some folders in your personal space)

You shouldn't get logged out at any point. You *may* see a `404` error, which is related to an internal Docker clock-hiccup, since you've also put the containers into sleep.

Without this change, there is a high chance that you get logged out at some point (maybe on the first request, maybe after navigating a few folders). 

fixes opencloud-eu/opencloud#3165

### DIFF
--- a/packages/web-client/src/sse/index.ts
+++ b/packages/web-client/src/sse/index.ts
@@ -150,3 +150,15 @@ export const sse = (baseURI: string, fetchOptions: FetchEventSourceInit): SSEAda
 
   return eventSource
 }
+
+/**
+ * Closes the current SSE connection and clears the singleton, so the next `sse()`
+ * call creates a fresh connection. Used on logout to stop the stream from
+ * reconnecting with a stale (removed) access token.
+ */
+export const resetSSE = (): void => {
+  if (eventSource) {
+    eventSource.close()
+    eventSource = null
+  }
+}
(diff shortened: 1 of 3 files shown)


tammi-23 (approved):
It looks good now after the final adjustment. 👍

PR #2956, opened 2026-07-25
fix: resolve hardcoded absolute image paths under a subpath deployment

kost:
Every asset reference on these pages: index.html's own favicon, manifest.json, config.json, the design-system stylesheet --is relative, and relies on the server-templated <base href> to resolve under whatever root the app is deployed at (see the companion opencloud-eu/opencloud subpath-deployment-support patch, which is what makes that server-side templating work correctly under a subpath in the first place). 28 image references across 25 files didn't follow that convention and used an absolute, domain-root path instead:

- Every <no-content-message img-src="/images/empty-states/*.svg"> (every empty folder/trash/users/groups/contacts/mail/search-results/ spaces/shares/app-tokens view, plus the 404 page itself).
- The login page's emblem (layouts/Plain.vue: icon-lilac.svg / icon-petrol.svg, chosen by theme).
- The rclone-crypt vault-not-found empty state and vault icon (UnlockVault.vue).

Under a subpath deployment these always resolved to the domain root instead of the deployment root, 404ing regardless of <base href> (absolute paths ignore it entirely). Dropped the leading slash on all of them so they resolve the same way as everything else on the page. No component-level or build-level changes needed: both consumers of these paths (native  for non-SVG, and vue-inline-svg's fetch()-based loading for .svg, which is all of these) resolve relative URLs against the document's base URL the same way, so the fix is purely mechanical.

Verified end to end: deployed a from-source build of this branch under a real subpath (/test/opencloud) behind a real nginx passthrough.

Testing: eslint clean on all changed files (0 errors; pre-existing, unrelated warnings only). Full unit suite (pnpm vitest run --config ./tests/unit/config/vitest.config.ts) passes, 2912 tests, one snapshot updated to match the corrected path.

## Types of changes

- [ ] Bugfix
- [ X ] Enhancement (a change that doesn't break existing code or deployments)

### DIFF
--- a/packages/web-app-admin-settings/src/components/Extensions/ExtensionsList.vue
+++ b/packages/web-app-admin-settings/src/components/Extensions/ExtensionsList.vue
@@ -2,7 +2,7 @@
   <no-content-message
     v-if="!filteredExtensions.length"
     id="admin-settings-extensions-empty-filtered"
-    img-src="/images/empty-states/empty-extensions.svg"
+    img-src="images/empty-states/empty-extensions.svg"
   >
     <template #message>
       <span v-text="$gettext('No extensions found')" />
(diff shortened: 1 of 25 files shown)


JammingBen (approved):
Thanks!

Issue #1232, opened 2025-09-22
Performance: Loading all groups makes AdminUI unusable


micbar:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

# Description

The opencloud User and group management UI displays a lot of information. To gather that, we are doing very expensive queries to the backend, which makes the UI completely unusable.

## Root causes

1. `https://cloud.keycloak-daily.opencloud.rocks/graph/v1.0/groups?%24orderby=displayName&%24expand=members` when opening the User Management.
2. Same expensive request to `https://cloud.keycloak-daily.opencloud.rocks/graph/v1.0/groups?%24orderby=displayName&%24expand=members` when opening the Group Management

These queries should be avoided. Fetching the groups is not problematic (1 sec or less) , but the `expand=members` kills it. That is only used to add a number to the groups in the list which does not add a lot of value because the assignments of a user can not be seen. We should avoid that in total.

The next problematic issue is the filter dropdown. The UX with more that 10 or 15 groups is not good. Imagine to have several thousand groups. Other UIs like the Keycloak UI use an autocomplete filter query field which scales better.

### Reproduction Steps

1. Create 50.000 users and 10.000 groups
2. Browse to the Admin UI to "Users" or "Group" management
3. White Screen, the Client does not respond within seconds.

### Expected Outcome

The UI should be usable.

### Actual Outcome

UI is not usable.

JammingBen:
In the users list, I think we're fetching the groups only for the filter and for the "Add To/Remove from Groups"-action. So I guess we could stop fetching them initially and only do it when opening the filter or a filter is applied (which can also be the case initially because of the URL query params), or when using one of these actions.

In the groups list, we would either need to kill the "Members" column, or the server can give the client a member count property via API.

JammingBen:
@micbar Are you 100% sure the request to `/graph/v1.0/groups` (+ `expand`) is what's taking so long?

I did some tests with ~6000 users and groups. The request itself is still fast (milliseconds), and it shouldn't be blocking the UI. However, there is an issue in the UI table rendering process, causing *all* items to render initially, before rendering the paginated amount. I expect this to be the real issue for the slow loading and especially the unresponsiveness of the browser.

[...]

PR #2958, opened 2026-07-27
perf: improve render times with lots of paginated items

JammingBen:
## Description

The `items-per-page` route query is written asynchronously, so on the first render `perPage` was NaN and `usePagination` returned the full list. With 6000 items (e.g. groups, users) that mounted all rows (~90k DOM nodes) before re-rendering with 50, blocking the main thread for ~5s. Fall back to the default when the query is not available yet.

Also makes sure that the navigate and scrollTo event handlers get unsubscribed on unmount.

## How has this been tested

- Create e.g. ~1000 groups
- Access the groups page in the admin-settings
- Watch it loading

## Benchmarks

Loading a list with ~6000 groups:

| Before | After |
|---|---|
| 5172 ms | 509 ms|

refs https://github.com/opencloud-eu/web/issues/1232


AlexAndBear (approved):
Good finding 🕵️‍♀️

PR #2963, opened 2026-07-28
perf: improve user and groups list loading in admin-settings

JammingBen:
## Description

* Parallelize requests for fetching users, groups and roles on the users list page and stop expanding group members since we don't use them anywhere in this list.
* Add a hint instead to use the filter to narrow down the list of items. This ensures that the UI stays responsive and doesn't lag.
* Stop auto-expanding members when listing groups. This also removes the members count in the list, which was not that helpful anyways. The members panel in the right sidebar now loads users on mount to display them for each group.

## How has this been tested

- Create ~1000 groups
- Open the "Groups"-filter on the users list page -> opening the filter should be notably faster
- Open the groups list page and have a look at the time for the `/graph/v1.0/groups` request in the network tab -> should be marginally faster
- Add 50 users to a group, then open the "Members" sidebar panel for this group -> rendering the members list doesn't cause a delay

## Benchmarks

With ~6000 users and groups on the system:

Rendering the users list on the users list page:

| Before | After |
|---|---|
| 1795 ms | 1492 ms |

Opening the groups filter on the users list page:

| Before | After |
|---|---|
| 1219 ms | 25 ms |

Loading all groups on the groups list page (request to `/graph/v1.0/groups`):

| Before | After |
|---|---|
| 360 ms | 330 ms |

refs https://github.com/opencloud-eu/web/issues/1232


tammi-23 (approved):
👍 Looks good! A really great update!
You save even more time with a large group.

PR #2964, opened 2026-07-28
feat: allow apps to provide themes

dschmidt:
Lets apps provide selectable themes to the theme switcher, generically (no
app-specific code in the core).

- `initializeThemes` merges app-provided theme configs alongside light/dark.
- `announceTheme` loads them before the first apply (no flash) and resolves
  asset paths relative to the app's own URL (works in dev and prod).
- `setAndApplyTheme` sets `html[data-theme]` (so themes can add structural CSS)
  and clears props a theme does not define (no leak across switches, e.g. fonts).

First consumer: the Windows 95 theme extension, opencloud-eu/web-extensions#511.

**Interim / draft:** themes are read from `external_apps[].config.themes`,
piggybacking on the per-app `config` bag so it works without a backend change
(marked `INTERIM` in `announceTheme`). The clean shape is a first-class
`external_apps[].themes` field, needing coordinated changes in the opencloud
backend and extension-sdk, then a one-line switch here.

### DIFF
--- a/packages/web-pkg/src/composables/piniaStores/theme.ts
+++ b/packages/web-pkg/src/composables/piniaStores/theme.ts
@@ -84,13 +84,33 @@ export const useThemeStore = defineStore('theme', () => {
 
   const availableThemes = ref<WebThemeType[]>([])
 
-  const initializeThemes = (themeConfig: ThemeConfigType) => {
+  // Custom prop keys (without the `--oc-` prefix) applied by the active theme,
+  // so a following theme can clear the ones it doesn't define instead of
+  // leaking them (applyCustomProp only sets, never clears).
+  const appliedCustomProps = ref<string[]>([])
+
+  const initializeThemes = (
+    themeConfig: ThemeConfigType,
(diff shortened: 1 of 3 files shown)


dschmidt:
Of course this collides a bit with the intention of unifying themes via a central theme.json, obviously this doesnt help for mobile client e.g. :/ 

So not sure this is a really good idea after all, my idea was just that we have an established approach to distribute apps (unlike themes)

dschmidt:
Superseded by #2971 and #2972.

The app-distribution approach (apps providing selectable themes via `external_apps[].config.themes`) added too much complexity for the value. Instead, the two generally useful, self-contained changes from this draft are split out:

- #2971 - expose the active theme as `html[data-theme]` (structural-CSS hook)
- #2972 - reset custom props not defined by the active theme (prop-leak fix)

Together with the existing global `theme.json` customization and global stylesheets (`options.styles`), a deployment can still ship theme-scoped structural CSS, without any app-specific machinery. Closing in favor of the two focused PRs.

PR #2971, opened 2026-07-29
feat: expose the active theme as html[data-theme]

dschmidt:
Sets the active theme's label on the `html` element (`html[data-theme="..."]`) so deployment CSS can attach structural styles (borders, radii, ...) that the design-token system alone cannot express.

Together with the existing global theme.json customization and global stylesheets (`options.styles`), a deployment can ship theme-scoped structural CSS without any app-specific machinery.

Split out of #2964 (which is superseded): this is the single, generally useful change from that draft, without the app-distribution complexity.


JammingBen (approved):
👍

PR #2988, opened 2026-07-30
fix: use Inter as the default font

### LINKED ISSUE #3206: Select greek letters dont have lowercase characters
## Describe the bug
In web UI, when anything in main UI or folder - file names includes the greek letters ω and δ, these are wrongly displayed as capital Ω and Δ. This is tested to be a purely css fonts issue. If the user uses browser dev tools to remove the applied font-face, the letters return to their correct lowercase form. 

## Steps to reproduce
1. Select Greek language from the user settings (this is easier for people not having greek keyboard in their systems)
2. Observe the now in greeks folder "Personal" being translated to "ΠρΩσΩπικα" instead of "Προσωπικα"

## Expected behavior
Letters ω and δ appearing (lowercase) instead of uppercase Ω and Δ

## Actual behavior
I tried to create a folder with greek letter ω and it appears capitalized no matter what I do. This is purely a web UI css issue as the folder is actually saved in the system with the correct letters. Using dev-tools I edited on the fly the css rule for font-face and the letters was then correctly displayed.

## Setup
I am using docker compose following the guidelines of official docs and version 7.3.0

tbsbdr:
fix: use Inter as the default font

The bundled OpenCloud webfont renders some Greek lowercase glyphs (e.g. ω, δ) as their uppercase counterparts, so text like "Προσωπικα" showed up as "ΠρΩσΩπικα". This is a font issue the stored data is correct. Drop the OpenCloud font from the web ui as we don't know which other issues can occur with the OpenCloud font. Falling back to Inter as default font. keeping the OpenCloud font files as they are referenced in the docs.

Fixes https://github.com/opencloud-eu/opencloud/issues/3206


AlexAndBear (approved):
🪦👍

JammingBen (commented):
Please also remove the OpenCloud fonts from `packages/design-system/src/assets/fonts/` and in the design-system docs: `packages/design-system/docs/.vitepress/theme/custom.scss` You might need to add the Inter font there as well (similar to how it's done in `packages/design-system/src/styles/fonts.scss`, mind the path though).

AlexAndBear:
> Please also remove the OpenCloud fonts from `packages/design-system/src/assets/fonts/` and in the design-system docs: `packages/design-system/docs/.vitepress/theme/custom.scss` You might need to add the Inter font there as well (similar to how it's done in `packages/design-system/src/styles/fonts.scss`, mind the path though).

Thought this was intentional ?

> keeping the OpenCloud font files as they are referenced in the docs.

🤔

Which would makes sense, since docs.opencloud.eu also uses the "official" font linking to the ods docs.

JammingBen:
> Which would makes sense, since docs.opencloud.eu also uses the "official" font linking to the ods docs.

True, I didn't think of that. But please move them to the docs folder in `packages/design-system/docs/.vitepress/theme/` then, since they don't belong to the bundled web source anymore.

tbsbdr:
@JammingBen 
I'd only change the font in the application (web UI) as it needs to support a much wider range of characters btw. in the mobile and desktop app we also dont have the custom opencloud font. For the rest of the ecosystem such as the documentation, website and presentations etc. we should keep the OpenCloud font because we have much more control over how it's used there. this is just a remedy as we can not guarantee that the opencloud font is robust enough for our needs. do you agree?

JammingBen:
> @JammingBen I'd only change the font in the application (web UI) as it needs to support a much wider range of characters btw. in the mobile and desktop app we also dont have the custom opencloud font. For the rest of the ecosystem such as the documentation, website and presentations etc. we should keep the OpenCloud font because we have much more control over how it's used there. this is just a remedy as we can not guarantee that the opencloud font is robust enough for our needs. do you agree?

Yep, makes sense!

JammingBen (approved):

PR #2992, opened 2026-07-31
fix: getting resource link for folder extensions

JammingBen:
Make sure that `useResourceLink().getResourceLink` returns the correct link for folders with extensions that can be opened a registered action. E.g. `.ocnb` folders should be opened with the notes extension per default.

fixes a regression from ff3be40e9a4d577ebc44f169d556393fd6840965


AlexAndBear (approved):
👍 Please add a regression test

PR #3018, opened 2026-08-03
feat(vault): use integrity id as passphrase verifier

### LINKED ISSUE #3009: Save encrypted string as webdav prop on vault
### Initial Checklist

### User Story

As a user, I want to set a vault's passphrase once so that I don't get confused and don't need to repeat that step.

### User Value

- UX

### Acceptance Criteria

- when creating a vault, a random encrypted string gets created with `uuidV4()` and saved as an arbitrary webdav prop `vault-integrity-id` on the folder
  - the property gets saved in the `oc` webdav namespace
- when decrypting a vault, instead of checking a random file, this webdav prop gets used for validating the passphrase instead

JammingBen:
Stores a random uuid, encrypted with the vault's content cipher, as the `ocrclone:integrity-id` property on the vault root when a passphrase is first set. This means an empty vault no longer re-offers setup. It also has the added benefits of a) having a cryptographically safe way to validate a passphrase (it wasn't before) and 2) less server load because we don't need to list all files inside a vault.

The old mechanism is kept as fallback however to ensure full compatibility with vaults created outside of Web (e.g. via rclone-crypt cli).

The webdav property lives in its own `ocrclone` namespace rather than the `oc` one because the server answers unknown `oc` local names with `404`. Also, the client has been extend to support extra properties on a per-request basis as well a `setProperties` method for writing arbitrary WebDAV properties via `PROPPATCH`.

closes https://github.com/opencloud-eu/web/issues/3009

### DIFF
--- a/packages/design-system/src/components/OcSwitch/OcSwitch.vue
+++ b/packages/design-system/src/components/OcSwitch/OcSwitch.vue
@@ -3,7 +3,7 @@
     <span class="inline-flex items-center gap-1">
       <!-- dim only the label + toggle when disabled, not the slot: an opacity on the whole
            switch would create a stacking context that traps slotted popovers (e.g. a helper) -->
-      <span :id="labelId" v-text="label" :class="{ 'opacity-40': disabled }" />
+      <span :id="labelId" :class="{ 'opacity-40': disabled }" v-text="label" />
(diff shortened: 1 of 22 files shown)


AlexAndBear (approved):
🥇

PR #3031, opened 2026-08-05
feat: rework vault setup process

### LINKED ISSUE #3032: Rework vault creation process
We want to add the vault creation to the regular create folder modal instead of having its own menu item, to make it look more "integrated" in our UI. Also, the process for spaces will be the same in the future.

Image

Image

JammingBen:
Reworks the setup process so creating vaults is now possible via a 2-step process in the regular modal for creating folders.

The way it works technically:

The folder modal checks if there is an encryption extension available that offers creating of vaults via a custom component. If given, display a UI toggle in the modal and render the custom component of the extension as second step of the folder creation process. This component is used by the rclone-crypt extension to let the user enter a password which then gets saved on the vault via a finalize callback.

Also, the UI for unlocking vaults has been reduced quite a lot.

closes https://github.com/opencloud-eu/web/issues/3032


AlexAndBear (approved):
She's a beauty 🥇

PR #3053, opened 2026-08-07
correct resource id in restore file version link

maki5:
the restore link was not created correct resulting in 404 response for shared resources, the id was not pointing to the correct resource but to the parrentFolderId instead, for non shared resources it was working because backend was geting the file id from the key itself and ignoring this part.

`/ dav / meta / <id> / v / <version-key>`

for non shared resource the file id is from `version-key` for shared from `ìd`


micbar:
@maki5 We should ping the web team for review.

@AlexAndBear @JammingBen

aduffeck:
Not gonna comment on the code change itself, but I can confirm that this fixes the issue of restoring versions on file shares.

JammingBen (approved):
Looks good, thanks! 👍

PR #3056, opened 2026-08-08
fix(web-pkg): keep sidebar panels mounted while loading

dschmidt:
The sidebar loading state replaced all root panels with a spinner (`v-if`/`v-else` in `SideBarPanels`), so any stateful panel was unmounted and recreated on every resource change. For the maps location panel this destroyed the map, so switching between geotagged files re-created it instead of animating (no fly-to).

Render the panels always and show the spinner as an overlay instead, so panels keep their state while loading.

Regressed by #2676, which started toggling the sidebar loading state on every resource change.


JammingBen (approved):
Thx!

Issue #3012, opened 2026-08-03
[tiptap] Mention users


JammingBen:
### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

As a user, I want to be able to mention other users in the text editor so that I have a quick and easy way to notify them.

### User Value

UX

### Acceptance Criteria

- typing `@` in a new line or after a blank opens a drop menu that lists all the users that have access to the current document (space members & share receivers)
- selecting a user puts it into a temporary list
- saving the document calls the collaboration/notification endpoint on the server with selected users from that list (see `packages/web-app-external/src/composables/useCollaboraPostMessages.ts:305`)

PR #3060, opened 2026-08-10
feat: add mention users for TipTap

tammi-23:
This will add https://github.com/opencloud-eu/web/issues/3012


packages/web-pkg/src/composables/mentions/useMentionUsers.ts
```diff
@@ -0,0 +1,169 @@
...
+}
+
+export function useMentionUsers({
+  space,
+  resource,
+  loadSharesTask: providedLoadSharesTask
+}: {
+  space: Ref<SpaceResource>
+  resource: Ref<Resource>
+  loadSharesTask?: LoadSharesTask
```
JammingBen: Do we really need to pass `loadSharesTask`? I think the composable can retrieve it from `useLoadShares` by itself.

packages/web-app-text-editor/src/App.vue
```diff
@@ -57,12 +61,30 @@ const placeholder = computed(() => {
...
 const textEditor = useTextEditor({
   contentType: unref(parsedContentType),
   currentResource: toRef(() => resource),
   modelValue: toRef(() => currentContent),
   readonly: isReadOnly,
   placeholder: unref(placeholder),
+  mentions: {
+    items: getMentionUsers,
+    onSelect: selectMention
+  },
```
JammingBen: This should be guarded by `useAuthStore().userContextReady`, otherwise it will try to fetch users on public links as well.

packages/web-pkg/src/composables/mentions/useMentionUsers.ts
```diff
@@ -0,0 +1,169 @@
...
+      return Promise.resolve(unref(collaborators))
+    }
+
+    if (!pendingCollaborators) {
+      const taskInstance = loadSharesTask.perform({
+        space: unref(space),
+        resource: unref(resource),
+        updateStore: false,
+        includeInheritedShares: true
+      })
```
JammingBen, line 49: I noticed a mean bug: when having the right sidebar sharing panel open in the editor, then typing `@` to mention a user, it will remove all the loaded shares. This is because `loadSharesTask` calls

```
sharesStore.setLoading(true)
sharesStore.removeOrphanedShares()
```

and never re-seeds the store because of `updateStore: false`. I think we should stop calling `setLoading` and `removeOrphanedShares` when `updateStore === false`.

packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
```diff
@@ -501,11 +503,15 @@ const saveFileTask = useTask(function* () {
...
     }
+    return false
   }
 }).drop()
 
 const save = async () => {
-  await saveFileTask.perform()
+  const saved = await saveFileTask.perform()
+  if (saved) {
+    await appOnSaveCallback?.()
```
JammingBen, line 617: Worth a `try...catch` and at least log the error?

JammingBen (commented):
Do we want to highlight the mentioned users with a special color, like on tiptap.dev?

Image

AlexAndBear:
> Do we want to highlight the mentioned users with a special color, like on tiptap.dev?
> 
> Image

Yes 🙌

fschade:
@tammi-23, FYI: https://github.com/opencloud-eu/libre-graph-api/pull/64

SRY

packages/web-pkg/src/editor/components/MentionMenu.vue
```diff
@@ -0,0 +1,157 @@
...
+
+onMounted(async () => {
+  await nextTick()
+  dropRef.value?.show?.({ anchorElement: anchorElement(), noFocus: true })
+})
+
+defineExpose({ onUpdate, onKeyDown })
+</script>
+
+<style>
```
JammingBen: Can we `scope` this?

packages/web-pkg/src/editor/types.ts
```diff
@@ -15,6 +15,16 @@ export type ContentType = 'plain-text' | 'markdown' | 'html' | 'tiptap-json'
  */
 export const DEFAULT_YDOC_FRAGMENT = 'default'
 
+export interface MentionItem {
+  id: string
+  label: string
+}
+
+export interface TextEditorMentionsOptions {
+  items: (query: string) => MentionItem[] | Promise<MentionItem[]>
```
JammingBen: Since this is a callback, maybe `getItems` would be more fitting?

packages/web-pkg/src/editor/extensions/mentions.ts
```diff
@@ -0,0 +1,183 @@
...
+        return mentionDecorations(transaction.doc, labels)
+      }
+    },
+    props: {
+      decorations: (state) => MentionHighlightPluginKey.getState(state)
+    },
+    view: (view) => {
+      let destroyed = false
+
+      Promise.resolve(items(''))
```
JammingBen: This gets called when the editor is mounted, which means it loads all shares each time a user opens a document. Can we ensure that this a) happens only when typing `@` and b) only for the user actually typing it, not for every connected peer?

packages/web-app-text-editor/src/App.vue
```diff
@@ -45,13 +43,35 @@ const placeholder = computed(() => {
   return $gettext('Write or type / for formatting options...')
 })
 
+const { getMentionUsers, notifyMentionedUsers, resetMentionState, selectMentionUser } =
+  useMentionUsers({
+    space: toRef(() => space),
+    resource: toRef(() => resource)
+  })
```
JammingBen: I think you should pack this inside a `if (mentionsEnabled)` block. Otherwise, the composable gets mounted and sets up `sharesStore.$onAction`, even if mentions are not enabled.

JammingBen (changes requested):
A few issue I noticed:

1. The own name is missing the highlight. For `alan` it looks like this after getting notified:

Image

2. When you type `@`, select a user, remove the line again and then save, the user still gets a notification.

3. Should we limit the functionality to `ocnote` and `md`? I don't think you want to mention users e.g. in a code document. The same inside code blocks, I think we should disable it as well for that.

JammingBen (approved):
Looks good! 🚀

PR #3061, opened 2026-08-10
fix: reset shift selection anchor when deselecting all files

AlexAndBear:
## Summary

Fixes a bug where the shift selection anchor was not properly reset after deselecting all files, causing incorrect file selection on subsequent shift+click operations.

## Description

When users deselected all files (either via the checkbox or the "Clear selection" button), the visual state was correctly updated, but the internal `shiftSelectionAnchorId` was not reset. This caused the next shift+click selection to incorrectly select files starting from the old anchor point instead of from the newly clicked file.

## Changes

- Added `eventBus.publish('app.files.shiftAnchor.reset')` to `resetSelection()` in the resources store
- Added `eventBus.publish('app.files.shiftAnchor.reset')` to `toggleSelectionAll()` in useResourceViewSelection composable
- Added eventBus import to resources.ts

## Test Plan

1. Select multiple files in tiles or table view
2. Click "Deselect All" or use the clear selection button
3. Verify all files are visually deselected
4. Click on a file, then shift+click on another file
5. Verify only the files between the two clicks are selected (not including files from the previous selection)

### DIFF
--- a/packages/web-app-files/src/composables/keyboardActions/useKeyboardFileMouseActions.ts
+++ b/packages/web-app-files/src/composables/keyboardActions/useKeyboardFileMouseActions.ts
@@ -11,7 +11,7 @@ export const useKeyboardFileMouseActions = (
   viewMode: Ref<string | QueryValue>
 ) => {
   const resourcesStore = useResourcesStore()
-  const { latestSelectedId } = storeToRefs(resourcesStore)
+  const { latestSelectedId, selectedIds } = storeToRefs(resourcesStore)
 
   let fileListClickedEvent: string
   let shiftAnchorResetEvent: string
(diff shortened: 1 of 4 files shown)


JammingBen (approved):

tammi-23 (dismissed):
There is still one issue. 
If you single selected all files via checking first and than last file and afterwards deselect with "select all" the anchor get not reseted.
Just click on an other file with pressed shift-key.

tammi-23:
This now also fix:
A shift click on the checkbox of an already selected file leaves that file
selected. Because the selection state doesn't change, the checkbox doesn't get
re-rendered and keeps the state that the browser toggled on click: the file was
selected, but its checkbox appeared unchecked.

kulmann (approved):

Issue #3074, opened 2026-08-11
Hashtag in Folder/file Name breaks Tiptap Editor


micbar:
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

Hashtags in parent folder names or file names break the loading of the Tiptap Editor.

### Reproduction Steps

1. Create Folder `Github#Ticket`
2. Create File `Test.md` inside that folder
3. Try to open the file in Tiptap

### Expected Outcome

File Content should be visible

### Actual Outcome

The Editor opens but no content is present.

v-scharf:
Please check that files with name can be open:
- test?.txt
- test#.txt
- test%20.txt
- tëst_日本.txt

PR #3081, opened 2026-08-11
fix(text-editor): update editor content when loaded via recovery

AlexAndBear:
## Summary
The text editor now properly updates when file content is loaded after initial mount, e.g., through the AppWrapper recovery mechanism. This ensures the editor displays content even when it's loaded asynchronously after the editor component has already been initialized.

## Changes
- Added a watcher in `packages/web-app-text-editor/src/App.vue` that detects when `currentContent` transitions from `undefined` to a value
- Only triggers during recovery scenarios to avoid interfering with normal editing or file switches

## Test plan
- [ ] Open a file in the text editor
- [ ] Verify content loads correctly in normal scenarios
- [ ] Test with files that trigger the recovery mechanism
- [ ] Verify editor updates properly when content loads after mount
- [ ] Ensure typing/editing is not affected by the watcher

Works towards #3074

### DIFF
--- a/packages/web-app-text-editor/src/App.vue
+++ b/packages/web-app-text-editor/src/App.vue
@@ -8,7 +8,7 @@
 </template>
 
 <script setup lang="ts">
-import { computed, toRef, unref } from 'vue'
+import { computed, toRef, unref, watch } from 'vue'
 import { useGettext } from 'vue3-gettext'
 import {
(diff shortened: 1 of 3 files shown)


JammingBen (approved):

kulmann (approved):
🙏

PR #3082, opened 2026-08-11
fix: add cache-buster for provided images

AlexAndBear:
## Description

image

## Related Issue

Update 

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix


AlexAndBear:
## Why this approach?

This solution uses query parameters (`?v=VERSION`) for cache busting, which is the most pragmatic choice for our use case:

### ✅ Advantages:
- **Simple and maintainable**: No complex build configuration changes needed
- **Standard practice**: Used by major CDNs (Google Fonts, jQuery, Bootstrap) 
- **No breaking changes**: URLs stay the same, just a parameter added
- **Works immediately**: No manifest system or lookup tables required
- **Package-based versioning**: Only invalidates cache on actual releases, not every dev build

### 🤔 Why not hash-based filenames?

We considered using Vite's built-in asset hashing (`icon-abc123.svg`), but:

1. **`vite-plugin-static-copy` doesn't support hashing** - files are copied as-is
2. **Would require custom tooling**: Need to build a manifest mapping `icon-name.svg` → `icon-name-abc123.svg`
3. **Breaking change**: All icon URLs would change, requiring lookup system
4. **Complex for 1000+ icons**: Dynamic access pattern (`icons/${name}.svg`) would need refactoring
5. **Vite's `public/` directory also doesn't hash** - same problem, different location

### 📊 Comparison:

| Approach | Complexity | Breaking Change | Works Now |
|----------|-----------|----------------|-----------|
| Query parameters (this PR) | Low | No | ✅ |
| Hash in filenames | High | Yes | ❌ |
| Import as modules | High | Yes | ❌ |

The query parameter approach solves the user-reported issue (stale icons after updates) with minimal changes and maximum compatibility.

fschade:
clever!

AlexAndBear:
> clever!

🤷‍♀️

packages/design-system/src/helpers/assets.ts
```diff
@@ -0,0 +1,17 @@
...
+export const addVersionToAssetUrl = (url: string): string => {
+  const version = process.env.PACKAGE_VERSION
+  if (!version) {
+    return url
+  }
+
+  const urlObj = new URL(url, window.location.origin)
+  urlObj.searchParams.set('v', version)
+
+  return `${urlObj.pathname}${urlObj.search}`
```
JammingBen: I think we should just append the version at the end of the string. Because right now, you construct a `URL` object with each recompute inside `OcIcon`, which gets quite expensive because we have a lot of them. Also, this transforms potential foreign origins to the OpenCloud origin as well as always making relative URLs absolute.

JammingBen (approved):

PR #3115, opened 2026-08-14
feat: e2ee vault spaces

### LINKED ISSUE #3029: E2EE vault spaces
### Initial Checklist

### User Story

As a user, I want to be able to work with e2e encrypted spaces so that can ensure that *all* data inside a space is encrypted on the server and only decrypted on the client.

### User Value

UX, privacy

### Acceptance Criteria

- `Drive`s have a new writable property `@libre.graph.contentType`. This will also be the identifier for E2EE spaces.
- The modal when creating a space has a new switch toggle "Enable end-to-end encryption" (1:1 like regular folder create)
- Creating an E2EE space:
  - Doesn't auto-create a `.space` folder (`template: none`)
  - Sets `@libre.graph.contentType: 'application/vnd.opencloud.vault'` (graph)
  - Sets the integration token via a proppatch, similar to how it's done with folders (webdav)
- Everything inside an E2EE space is encrypted
  - That means E2EE spaces have the same limitations as folder vaults (no sharing inside, space is not sharable via public link etc.)
- E2EE spaces don't support images, emojis or descriptions (for now)
- The spaces list shows a lock indicator (open/closed) for E2EE spaces, similar to the file list

JammingBen:
Add support for e2ee vault spaces.

This change looks bigger than it actually is: a lot is just removing the word "folder" from function and file names, since it applies to spaces as well now.

The main changes:

- spaces can read and be updated with the new `@libre.graph.contentType` property.
- the `rclone-crypt` app claims spaces that have this property set to `application/vnd.opencloud.vault`.
- UX is pretty much 1:1 like it is for folders.

closes https://github.com/opencloud-eu/web/issues/3029


packages/design-system/src/assets/icons/resource-type-space-vault-fill.svg
kulmann, line 1: this image is very bright in dark mode 🙈  but we have the same issue with the image for regular spaces (without the lock), so maybe that's something for a followup.

kulmann (approved):
fantastic 😍

packages/design-system/src/assets/icons/resource-type-space-vault-fill.svg
JammingBen: We have a `hasDarkVariant` prop on resource icons, so we could have a different icon there. Although it wouldn't work for regular spaces because the default image comes from the server. I can't remember why we did that and not let Web handle the fallback 🤔 Anyway, I agree it's something for a followup rather.

PR #3147, opened 2026-08-17
fix: render action buttons in move/copy dialogs

### LINKED ISSUE #2975: Unrendered "abort/confirmation" buttons in "copy to" and "move to" dialogue
### Initial Checklist

### Bug Description

The "move to" and "copy to" dialogues do not render the abort or confirmation buttons correctly. They should be rendered like in other dialogues, too.

### Reproduction Steps

Open either "Move to" or "Copy to" dialogue

### Expected Outcome

Same rendering like e.g. in "Rename" dialogue:

Image

### Actual Outcome

"Move to" dialogue:

Image

"Copy to" dialogue:

Image

junkerderprovinz:
## Description

The "Move to" and "Copy to" dialogs open the file list in embed mode. Their action buttons come from `EmbedActions.vue`, which is teleported into `#app-runtime-footer`. That footer sits on the chrome colored area of the application layout (`bg-role-chrome` in `Application.vue`).

The buttons used `color-role="chrome"` themselves:

- the confirming button (`Move here` / `Copy here` / `Choose` / `Save` / `Attach as copy` / `Share link(s)`) was `appearance="filled"` with the chrome role, so it was filled with exactly the color of the area it is placed on and became invisible,
- the cancel button was `appearance="raw-inverse"` with `no-hover`, so it was rendered as plain text without any hover feedback.

Both therefore looked unrendered, as reported.

This changes the actions to the color role and appearances that the other dialogs use (see `UnsavedChangesModal.vue`, `UnlockVault.vue`, `OcModal.vue`): the default `secondary` color role, `outline` for cancel and `filled` for the confirming action. The buttons now render like the ones in the "Rename" dialog in both the light and the dark theme.

Note that the default themes shipped by OpenCloud do not define a `chrome` role, so it falls back to `surfaceContainer` / `onSurface` in the theme store, which is why the filled button was painted in the same color as the surrounding area.

## Related Issue

- Fixes #2975

## How Has This Been Tested?

- test environment: local monorepo checkout, Node 24, pnpm 11
- test case 1: extended `packages/web-app-files/tests/unit/components/EmbedActions/EmbedActions.spec.ts` with a `button appearance` block asserting that the cancel action renders as an outlined button, that the select and share actions render as filled buttons, and that none of them carries the chrome color role class. The three new assertions fail on `main` and pass with this change.
- test case 2: `pnpm test:unit --run packages/web-app-files/tests/unit/components/EmbedActions/EmbedActions.spec.ts` passes (19 tests)
- test case 3: `pnpm check:types` and `pnpm lint` pass without errors, `prettier --check` is clean for the touched files. The full unit test run has 6 pre-existing failures in unrelated files (conflict handling, `ResourcePreview`, `CreateFolderModal`) which also fail on an unmodified `main`.

## Types of changes

- [x] Bugfix

### DIFF
--- a/packages/web-app-files/src/components/EmbedActions/EmbedActions.vue
+++ b/packages/web-app-files/src/components/EmbedActions/EmbedActions.vue
@@ -11,14 +11,7 @@
     />
 
     <div class="flex items-center ml-auto">
-      <oc-button
-        class="mr-4"
-        data-testid="button-cancel"
-        color-role="chrome"
-        appearance="raw-inverse"
-        no-hover
-        @click="emitCancel"
-      >
+      <oc-button class="mr-4" data-testid="button-cancel" appearance="outline" @click="emitCancel">
         {{ $gettext('Cancel') }}
       </oc-button>
       <oc-button
@@ -26,7 +19,6 @@
         key="btn-share"
         class="mr-4"
         data-testid="button-share"
-        color-role="chrome"
         appearance="filled"
         :disabled="isShareLinksButtonDisabled"
         @click="createLinkAction.handler({ resources: selectedFiles, space })"
@@ -38,7 +30,6 @@
           v-if="isLocationPicker"
           data-testid="button-select"
           appearance="filled"
-          color-role="chrome"
           :disabled="isChooseButtonDisabled"
           @click="emitSelect"
         >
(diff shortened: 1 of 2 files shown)


JammingBen (approved):
Thanks!

PR #3152, opened 2026-08-18
refactor: use svg for space image fallback

JammingBen:
Before, the fallback image for spaces without image data was a PNG asset that got loaded via an authenticated request. Change it to an SVG so it's smaller in size and aligns with our general resource icon loading.

Also, use the new space images in the admin-settings.


AlexAndBear (approved):
nice

PR #3157, opened 2026-08-18
fix(design-system): avoid duplicated SVG ids

JammingBen:
Avoid duplicated ids in SVGs by either redrawing them and removing their ids, or by passing a `uniqueId` param to the `OcIcon` component to ensure they stay unique and don't conflict.


JammingBen:
~Converted to a draft for now - maybe we should just remove those ids from the SVGs.~

kulmann (approved):

PR #3173, opened 2026-08-19
feat: make space header more compact

### LINKED ISSUE #588: [FR] Space description optional / compact space view
### Initial Checklist

- [x] I understand this is a feature request and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)

### Idea

Currently, the header section of a space is always rendered, even if no description is set. This results in unnecessary use of vertical screen space without providing any content value.

Suggested improvement:

* Do not display the header area if the description is empty or not set.
* Optionally, introduce a configurable "trigger word" (e.g., #hide) in the description field.
If this is detected, the header area should also be hidden.

This would improve layout efficiency and provide more space for meaningful content.

JammingBen:
closes https://github.com/opencloud-eu/web/issues/588

image

### DIFF
--- a/packages/web-app-files/src/components/Spaces/SpaceHeader.vue
+++ b/packages/web-app-files/src/components/Spaces/SpaceHeader.vue
@@ -1,13 +1,13 @@
 <template>
   <div
-    class="space-header p-4"
+    class="space-header py-4"
     :class="{ flex: !imageExpanded && !isMobile, 'space-header-squashed': isSideBarOpen }"
   >
     <div
-      class="space-header-image mr-6 min-w-xs aspect-[16/9]"
+      class="space-header-image mr-6 aspect-[16/9]"
       :class="{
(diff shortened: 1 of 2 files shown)


JammingBen:
@kulmann I'd argue that the batch actions should be placed beneath the space header. The current position only made sense because we had the "New"-buttons in the past. Agree? Though this is a bigger change because of the sticky app bar, so nothing for this PR.

kulmann:
> @kulmann I'd argue that the batch actions should be placed beneath the space header. The current position only made sense because we had the "New"-buttons in the past. Agree? Though this is a bigger change because of the sticky app bar, so nothing for this PR.

yep, fully agree with that. the "ugly whitespace" should move below the space header entirely, including the batch actions and in the future the inline notes.

kulmann (approved):

Issue #3166, opened 2026-08-19
File list: rendering one row or tile triggers 428 router.resolve() calls


JammingBen:
Part of the file list scroll performance work for #3159.

## Current situation

Rendering a single file list row or tile triggers **428 `router.resolve()` calls**.

`getResourceLink()` calls `getDefaultAction()`, which evaluates `isVisible()` on every registered file action. A default instance has 209 registered file extensions, so `editorActions` alone produces 209 actions. Each of those `isVisible()` implementations starts with `isLocationTrashActive(router, 'files-trash-generic')` before doing the cheap extension/mimeType comparison, and `isLocationActive()` performs two `router.resolve()` calls - one of them on the full current route.

`router.resolve()` is additionally expensive because the router uses `qs.stringify` as its `stringifyQuery`.

Measured on a 500-image folder (items per page 500):

- revealing 4 tiles → 1712 `router.resolve()` calls
- revealing 22 table rows → 9636 calls, 48 ms
- scrolling over already-rendered items → 0 calls (the cost is strictly per item, there is no re-render cascade)
- a full pass over 500 items ≈ 214,000 calls ≈ 1.1 s of pure route resolution in Chromium

A Chromium CPU profile of a scroll shows the hottest stack as `getResourceLink → isVisible → router.resolve → stringifyQuery → qs encode`.

Stubbing `router.resolve` entirely brings the Chromium tiles view from 19 ms to 13 ms median frame time.

## Why this is bad

Since #3152/7.3.0 this work happens per item while scrolling instead of once on folder load, so it competes directly with the frame budget. It is the largest single contributor to sluggish scrolling in large folders, and it scales with the number of installed apps.

## Proposed solution

1. In `editorActions.isVisible` (`packages/web-pkg/src/composables/actions/files/useFileActions.ts`), move the extension/mimeType comparison to the top so 208 of 209 actions bail out before touching the router. Keep the trash-location check, just after it. https://github.com/opencloud-eu/web/pull/3177
2. Memoize `isLocationActive` (`packages/web-pkg/src/router/utils.ts`): cache `router.resolve(currentRoute).href` and each comparative's `href`, keyed on the current route, invalidated on navigation. https://github.com/opencloud-eu/web/pull/3182
[...]

PR #3177, opened 2026-08-19
perf: remove location checks from resource actions

JammingBen:
Remove the route location checks from `isVisible` on resource actions. Each check called `router.resolve` internally, which gets quite expensive in big file lists. Also, it much cleaner to check the resources or their spaces instead.

refs https://github.com/opencloud-eu/web/issues/3166 (steps 1 & 3)


packages/web-app-files/src/composables/actions/files/useFileActionsClearClipboard.ts
JammingBen, line 1: Was unsued.

packages/web-app-files/src/composables/actions/files/useFileActionsDisableSync.ts
```diff
@@ -87,21 +86,12 @@ export const useFileActionsDisableSync = () => {
...
+      isVisible: ({ resources }) => {
         if (resources.length === 0) {
           return false
         }
 
-        if (
-          isLocationSpacesActive(router, 'files-spaces-generic') &&
-          (space?.driveType !== 'share' || resources.length > 1 || resources[0].path !== '/')
-        ) {
+        if (!resources.every(isIncomingShareResource)) {
```
kulmann:
```suggestion
        if (resources.some(!isIncomingShareResource)) {
```
^ this should be more efficient, right?

kulmann (approved):

PR #3182, opened 2026-08-20
perf: introduce cache for active location checks

### LINKED ISSUE #3166: File list: rendering one row or tile triggers 428 router.resolve() calls

JammingBen:
## Description

The `isLocationActive` util gets called several times on each page, which can get expensive since it does a `router.resolve` internally. This needs to be done for the current route and the given route we're comparing it to.

Add a cache for the resolved routes so calling it multiple times on/for the same routes hits the cache instead of resolving them again.

Also, remove the legacy deprecated route handling. While technically a breaking change, IMO it's fine not to announce it as that, because there aren't any OpenCloud instances out there relying on these.

## Benchmarks

Note that the absolute milliseconds were measured in Node/jsdom, not Chromium. They are primarily useful as a ratio, not as absolut numbers.

| metric | before  | after  |
|--|--|--|
| router.resolve per isLocationTrashActive() call       | 2       | 0      |
| time per isLocationTrashActive() call                 | 2679 ns | 105 ns |
| router.resolve for the first revealed item on a route | 420     | 2      |
| router.resolve per further revealed item              | 420     | 0      |
| router.resolve for revealing 22 table rows            | 9240    | 0      |
| time for revealing 22 table rows                      | 14.2 ms | 2.0 ms |
| router.resolve for a full pass over 500 items         | 210 000 | 0      |
| time for a full pass over 500 items                   | 313 ms  | 32 ms  |

closes https://github.com/opencloud-eu/web/issues/3166


packages/web-pkg/src/router/utils.ts
```diff
@@ -6,47 +6,75 @@ export interface ActiveRouteDirectorFunc<T extends string> {
...
+
+  return nextCache
+}
+
+const resolveComparativeHref = (
+  router: Router,
+  cache: ResolvedHrefCache,
+  comparative: RouteLocationNamedRaw
+): string => {
+  const cachedHref = cache.comparativeHrefs.get(comparative)
```
kulmann, line 50: shouldn't we use `comparative.name.toString()` as keys for the map? Seems to be a little bit safer than relying on object identity...

packages/web-pkg/src/router/utils.ts
```diff
@@ -6,47 +6,75 @@ export interface ActiveRouteDirectorFunc<T extends string> {
...
+const getHrefCache = (router: Router): ResolvedHrefCache => {
+  const currentRoute = unref(router.currentRoute)
+  const cache = hrefCaches.get(router)
+
+  if (cache && cache.path === currentRoute.path) {
+    return cache
+  }
+
+  // FIXME: router.resolve cleans the path. we don't need it, if we can rely on
+  // router.currentRoute to not have slashs encoded for paths
```
kulmann:
```suggestion
  // router.currentRoute to not have slashes encoded for paths
```

kulmann (approved):

packages/web-pkg/src/router/utils.ts
JammingBen: A `WeakMap` only works with objects as keys though. And `comparative.name` is optional, so it's not a given that we even have it.

PR #3189, opened 2026-08-20
test(e2e): add e2ee space tests

JammingBen:
Follow-up for https://github.com/opencloud-eu/web/pull/3115.


v-scharf (commented):
maybe we also need check that resources inside the space cannot be shared? 
like:`{string} should not be able to share following resource(s) from the space {string}` 

expect that share btn is not visible
Image

v-scharf (approved):
Cool! thank you!

Issue #3167, opened 2026-08-19
Tiles view: previews are fetched at 1000x1000 and never released


JammingBen:
Part of the file list scroll performance work for #3159.

## Current situation

`ImageDimension.Tile` is `[1000, 1000]` (`packages/web-pkg/src/constants.ts`). Each tile thumbnail is a ~816 KB response, decoded into roughly 4 MB of bitmap, and displayed in a 316x178 px box.

On top of that:

- `previewService.privatePreviewBlob` calls `window.URL.createObjectURL()` but nothing ever revokes those URLs, so no preview is ever freed
- `itemVisible` queues a preview for every item the user scrolls past, and nothing cancels queued work for items that have left the viewport
- `cacheService.filePreview` has a capacity of 250 for a 500-item page, so the second half evicts the first and previews are re-fetched when scrolling back
- `Cache.evict()` walks the whole map on every `get` and `set`
- `useLoadPreview` creates its own `PQueue` per caller, so concurrency is not bounded globally

Measured on a 500-image folder (items per page 500):

| Scenario | Engine | Median frame | Tiles filled in 4 s |
| --- | --- | --- | --- |
| Tiles, baseline | WebKit | 80 ms | 108 |
| Tiles, preview requests blocked | WebKit | 32 ms | 348 |
| Tiles, baseline | Chromium | 18 ms | 448 |
| Tiles, preview requests blocked | Chromium | 10 ms | 500 |

Sampled response: `processor=fit&x=1000&y=1000`, 836,221 bytes, 37-41 ms each. 500 tiles is roughly 400 MB of image data.

The table view is much less affected (WebKit 47 ms → 39 ms) because table thumbnails are 36x36.

## Why this is bad

Preview loading is the single biggest cost in the tiles view on Safari - roughly 60% of the frame time. The unbounded memory growth also degrades long sessions, and re-fetching on scroll-back wastes bandwidth and puts avoidable load on the thumbnails service.

## Proposed solution

1. Derive the tile preview size from the rendered tile size multiplied by `devicePixelRatio` and capped - roughly 640x640 instead of 1000x1000. Leave `ImageDimension.Preview` alone for the preview app and `SpaceHeader`. https://github.com/opencloud-eu/web/pull/3192
2. Revoke object URLs when a preview cache entry is evicted or replaced, and when the resource list is cleared. https://github.com/opencloud-eu/web/pull/3194
[...]

PR #3192, opened 2026-08-20
perf: dynamic tile preview dimensions

JammingBen:
Derive the requested tile preview dimensions from the current tile size setting and the device pixel ratio. This way we avoid fetching large previews for small tile sizes.

Also reduce the max tile dimension from 1000 to 768 px because this feels sufficient, even on the largest tile size setting, and optimize the requested preview sizes in the right sidebar and the photo roll.

### DIFF
--- a/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
+++ b/packages/web-app-files/src/components/SideBar/Details/FileDetails.vue
@@ -358,7 +358,7 @@ watch(
       preview.value = await loadPreview({
         space: unref(space),
         resource: unref(resource),
-        dimensions: ImageDimension.Preview,
+        dimensions: ImageDimension.Medium,
         cancelRunning: true,
         updateStore: false
       })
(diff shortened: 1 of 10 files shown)


JammingBen:
The net win of this PR is very small as of now, because the server only supports a very limited number of resolutions. There is nothing between `128x128` and `1920x1080` (or `1080x1920`). This needs to be tackled in the server.

AlexAndBear:
> The net win of this PR is very small as of now, because the server only supports a very limited number of resolutions. There is nothing between `128z128` and `1920x1080` (or `1080x1920`). This needs to be tackled in the server.

I was already wondering 😅

JammingBen:
See https://github.com/opencloud-eu/opencloud/pull/3386 and https://github.com/opencloud-eu/opencloud/pull/3377. No need to wait for these though, this change prepares Web for this.

kulmann (approved):

guruz:
Input appreciated here for iOS and Android: https://github.com/opencloud-eu/android/pull/217#issuecomment-5436309312

PR #3194, opened 2026-08-21
perf: preview cache optimizations

JammingBen:
* Revoke preview URLs when cache entries get evicted to ensure we don't keep any stale preview URLs for files that are long gone.
* Introduce a byte limit to the preview cache, instead of a plain capacity limit. This makes more sense and generally increases the limit in file lists, which should improve performance in big file lists with lots of previews.
* Calling `get()` and `set()` on the cache stops looping the entire cache to evict and instead does a lookup.

### DIFF
--- a/packages/web-pkg/src/composables/piniaStores/resources.ts
+++ b/packages/web-pkg/src/composables/piniaStores/resources.ts
@@ -49,15 +49,8 @@ export const useResourcesStore = defineStore('resources', () => {
     resources.value = file
   }
 
-  /** Drops cached previews of resources that are gone, so their object URLs get revoked. */
-  const releasePreviews = (values: Resource[]) => {
-    for (const resource of values) {
-      releaseFilePreviews(resource.id)
-    }
-  }
-
   const removeResources = (values: Resource[]) => {
-    releasePreviews(values)
+    releaseFilePreviews(values.map(({ id }) => id))
     resources.value = unref(resources).filter((file) => !values.find(({ id }) => id === file.id))
   }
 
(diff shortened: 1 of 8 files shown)


packages/web-pkg/src/composables/piniaStores/resources.ts
```diff
@@ -49,7 +49,15 @@ export const useResourcesStore = defineStore('resources', () => {
     resources.value = file
   }
 
+  /** Drops cached previews of resources that are gone, so their object URLs get revoked. */
+  const releasePreviews = (values: Resource[]) => {
+    for (const resource of values) {
```
kulmann: iterating here and then iterating again in `releaseFilePreviews` has pretty high computational complexity. And although `releaseFilePreviews` is exported, it's never used anywhere but here.

I'd recommend building `releaseFilePreviews` with an array as argument, e.g. like this:

```js
export function releaseFilePreviews(resourceIds: string[]): void {
  const prefixes = resourceIds.map((id) => buildFilePreviewCacheKey(id))
  for (const key of filePreviewCache.keys()) {
    if (prefixes.some((p) => key.startsWith(p))) {
      filePreviewCache.delete(key)
    }
  }
}
```

packages/web-pkg/src/services/cache.ts
```diff
@@ -1,13 +1,59 @@
...
+}) => {
+  const { src } = replacedValue
+  if (src?.startsWith('blob:') && src !== value?.src) {
+    window.URL.revokeObjectURL(src)
   }
->({ ttl: 10 * 1000, capacity: 250 })
+}
+
+/** Build a cache key for a file, including its dimensions if provided. */
+export function buildFilePreviewCacheKey(
```
kulmann, line 35: what about the `provider`? shouldn't that also be part of the cache key?

kulmann (changes requested):

packages/web-pkg/src/composables/piniaStores/resources.ts
JammingBen: Yep good catch :) More performaaaaance

packages/web-pkg/src/services/cache.ts
JammingBen: As stated in chat, I think this is not really necessary for now, so I'd leave it out at the moment.

kulmann (approved):
🚤 🏇 🏎️

PR #3198, opened 2026-08-21
perf: drop queued previews leaving the viewport

JammingBen:
Drop previews from the loading queue when they have not yet been loaded and leave the viewport.

This boosts performance when scrolling over a big file list fast, since there is no need to load a preview for a resource the user just scrolled by fast.


AlexAndBear (commented):
Great 👍

AlexAndBear (approved):

Issue #3033, opened 2026-08-05
Collaboration via Yjs


JammingBen:
This is a collection of tasks for bringing collaboration via Yjs to OpenCloud.

### Groundwork

- [x] Bring https://github.com/opencloud-eu/web/pull/2785 into a mergable state, then review and merge it @JammingBen 

### Direct follow-ups

These should happen *before* we officially announce this as productized.

- [x] Add e2e tests #3064
- [x] #3100
- [x] #3101
- [x] #3103
- [x] #3104
- [x] #3105
- [x] #3106
- [ ] #3108
- [ ] #3107

#### Features

- [x] #3109
- [x] #3110
- [ ] #3111
- [ ] #3112
- [ ] Handling folders in the `AppWrapper` (notes extension via ocnb)

#### Hardening

- [ ] #3132
- [ ] #3133
- [x] #3141
- [ ] #3113
- [ ] #3217
- [ ] #3239

### Needs concept

- [ ] Version handling
- [ ] Auto save intervals -> possible with auto save election
- [ ] Check compatibility in vaults, i.e. can we make it work in vaults?
- [ ] Make it work on public links -> worth the efforts?
- [ ] Another peer's edit messes with your local dirty state

### Decisions

- ~Make "Show source" work for markdown files despite collaboration enabled on the system (might need a concept)~ Known limitation: "Show source" doesn't work on instances with collaboration

dschmidt:
> Check compatibility in vaults, i.e. can we make it work in vaults?

Yes, absolutely - but the question is rather: can we encrypt the Yjs traffic using the vault passphrase so the content is not leaked to the server?
Is that "secure" enough or do a lot of small messages allow easier decryption of the passphrase?

AlexAndBear:
> Make "Show source" work for markdown files despite collaboration enabled on the system (might need a concept) Known limitation: "Show source" doesn't work on instances with collaboration

I was at least able to re-enable it but in read-only view

https://github.com/opencloud-eu/web/pull/3405

PR #3199, opened 2026-08-21
feat(editor): add lowlight code blocks

AlexAndBear:
## Description

image

This PR adds syntax-highlighted code blocks to the Tiptap editor using `@tiptap/extension-code-block-lowlight` and a dedicated code block node view with language selection.

It also switches Highlight.js themes based on the active app theme (`isDark`):
- light mode: `atom-one-light`
- dark mode: `atom-one-dark`

## Related Issue

- Fixes N/A

## How Has This Been Tested?

- test environment: local development environment
- test case 1: `./node_modules/.bin/vue-tsc --declaration --noEmit --project packages/web-pkg/tsconfig.build.json`
- test case 2: verified code block language selector rendering and position in the editor node view
- test case 3: verified theme stylesheet switches between atom-one-light and atom-one-dark when theme changes

## Types of changes

- [ ] Bugfix


AlexAndBear:
## Build Size Comparison

I compared the build sizes between `main` and this PR branch to assess the impact of adding lowlight for syntax highlighting.

### Editor Chunk Size

**`main` branch:**
- `dist/js/chunks/editor-*.mjs`: **783.32 kB** (gzip: 240.58 kB)

**This PR (with lowlight):**
- `dist/js/chunks/editor-*.mjs`: **962.76 kB** (gzip: 297.01 kB)

### Impact

- **Uncompressed:** +179.44 kB (+22.9%)
- **Gzipped:** +56.43 kB (+23.4%)

The lowlight package (including the highlight.js themes) adds approximately **56 kB (gzipped)** to the editor bundle. This is a noticeable but reasonable increase for syntax highlighting functionality.

JammingBen:
I think it's fine for the scope of this PR, but we're already getting a warning about the biggest chunk, and now it's even more big. So we should make the editor components async and/or put them into dedicated chunks (kinda already a to-do in https://github.com/opencloud-eu/web/issues/3033).

packages/web-pkg/src/editor/components/CodeBlockComponent.vue
```diff
@@ -0,0 +1,90 @@
...
+      stylesheet = document.createElement('link')
+      stylesheet.id = hljsThemeStylesheetId
+      stylesheet.rel = 'stylesheet'
+      document.head.appendChild(stylesheet)
+    }
+
+    stylesheet.href = isDark ? atomOneDarkThemeUrl : atomOneLightThemeUrl
+  },
+  { immediate: true }
+)
```
JammingBen: It feels a bit wrong to have this per code block, especially since it doesn't get cleaned up. Maybe put it in `TextEditorContent.vue` (+ proper cleanup on unmount), since it's something global per editor?

JammingBen (changes requested):
A few findings from my side:

- Inline blocks via single backticks `like this` lost their styling.
- The select is editable in read-only mode.
- The select overflows the code block initially, which looks a bit off. Maybe reduce its size? Or more general: do we really need this? 🙈 I feel like this feature gets mostly used by technical users that now how to work with code blocks.

Image

packages/web-pkg/src/editor/extensions/imageFileHandler.ts
kulmann, line 1: good that you moved it 👍

AlexAndBear:
> A few findings from my side:
> 
> - Inline blocks via single backticks `like this` lost their styling.
> - The select is editable in read-only mode.
> - The select overflows the code block initially, which looks a bit off. Maybe reduce its size? Or more general: do we really need this? 🙈 I feel like this feature gets mostly used by technical users that now how to work with code blocks.
> 
> Image

Hmm I would keep the select, as it also indicates which languages are supported for code hilighting 🤔

JammingBen (approved):

PR #3204, opened 2026-08-24
perf: decode thumbnails asynchronously

JammingBen:
Add a [decoding prop](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#decoding) to `OcImage` and set `decoding="async"` on thumbnails in `ResourceTile`, `ResourceListItem`, `Projects`, `FileDetails`, `SpaceHeader` and `SpaceDetails`.

### DIFF
--- a/packages/design-system/src/components/OcImage/OcImage.vue
+++ b/packages/design-system/src/components/OcImage/OcImage.vue
@@ -1,5 +1,12 @@
 <template>
-  <img :src="src" :alt="alt" :aria-hidden="`${ariaHidden}`" :title="title" :loading="loadingType" />
+  <img
+    :src="src"
+    :alt="alt"
+    :aria-hidden="`${ariaHidden}`"
+    :title="title"
+    :loading="loadingType"
+    :decoding="decoding"
+  />
 </template>
 
 <script setup lang="ts">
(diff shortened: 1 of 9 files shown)


packages/design-system/src/components/OcImage/OcImage.vue
```diff
@@ -23,9 +30,14 @@ export interface Props {
...
   title?: string
+  /**
+   * @docs The decoding hint of the image.
+   * @default auto
+   */
+  decoding?: 'sync' | 'async' | 'auto'
 }
 
-const { src, alt = '', title, loadingType = 'eager' } = defineProps<Props>()
+const { src, alt = '', title, loadingType = 'eager', decoding = 'auto' } = defineProps<Props>()
```
kulmann: wouldn't `decoding = undefined` let the browser fall back to `auto` anyway while keeping the DOM cleaner?

kulmann (commented):
what about photoroll in preview app, FileDetails, SpaceHeader, SpaceDetails?

packages/design-system/src/components/OcImage/OcImage.vue
JammingBen: Yep good idea!

kulmann (approved):

PR #3210, opened 2026-08-24
fix(design-system): useIsVisible observer look-ahead

### LINKED ISSUE #3168: File list lazy loading has no look-ahead: rootMargin is ineffective

## Current situation

`useIsVisible` (`packages/design-system/src/composables/useIsVisible/index.ts`) creates its `IntersectionObserver` with the default root - the viewport - and `rootMargin: '100px'`.

The file list does not scroll in the viewport. It scrolls inside `.files-view-wrapper`, which has `overflow-y-auto` (`packages/web-app-files/src/components/FilesViewWrapper.vue`). Per spec, an intermediate scroll container clips the target before the root margin is applied, so **the root margin has no effect at all**.

Measured on a 500-row table, same scroll position, four observer configurations:

| Configuration | Rows reported visible |
| --- | --- |
| `root: null`, `rootMargin: 100px` (current) | 19 |
| `root: null`, `rootMargin: 1000px` | 19 |
| `root: wrapper`, `rootMargin: 0px` | 19 |
| `root: wrapper`, `rootMargin: 1000px` | 43 |

Container height is 930 px and rows are 42 px, so about 22 rows fit on screen. 19 rendered means there is zero buffer.

## Why this is bad

Rows and tiles are only marked visible at the exact moment they enter the viewport. There is no look-ahead, so scrolling faster than the render can keep up shows blank space instead of content - which is the main symptom reported in #3159.

## Proposed solution

1. Add a `root` option to `useIsVisible` and pass the scroll container from the list components (or resolve the nearest scrollable ancestor). Without this, `rootMargin` stays inert no matter what value it has.
2. Raise the look-ahead to roughly one container height in both directions.
3. Confirm empirically that the number of observed rows grows - the four-configuration probe above is the check.

Worth doing after #3166 and #3167 land, because a bigger look-ahead multiplies whatever the per-item render cost still is.

## Verification

Fling to the bottom of a 500-item folder: no frame where more than half the on-screen items are still placeholders.

JammingBen:
Fixes the look-ahead of the `useIsVisible` composable by passing it the parent scroll container as `root` property and creating the observer lazily to ensure the target element is mounted.

You can test this by adjusting your browser window size so it fits e.g. 15 tiles in the file list. Now open a folder with >30 images and look at the network tab. It should load ~30 thumbnails because of the fixed look-ahead, which is now about one viewport.

### Breaking for devs

The `root` prop of the `useIsVisible` composable is mandatory now.

fixes https://github.com/opencloud-eu/web/issues/3168

### DIFF
--- a/packages/design-system/src/composables/useIsVisible/index.ts
+++ b/packages/design-system/src/composables/useIsVisible/index.ts
@@ -2,13 +2,28 @@ import { Ref, onBeforeUnmount, ref, watch } from 'vue'
 
 export const useIsVisible = ({
   target,
+  root,
   mode = 'show',
-  rootMargin = '100px',
+  rootMargin = '100% 0%',
   onVisibleCallback,
   onHiddenCallback
 }: {
   target: Ref<Element>
-  mode?: string
+  /**
+   * The scrollable element the look-ahead is measured against. Until it resolves,
+   * the viewport is used, which gives no look-ahead inside a scroll container.
+   */
+  root: Ref<Element | null | undefined>
+  /**
+   * The mode determines whether the target should be observed for visibility
+   * changes after it has been visible once.
+   * @default 'show'
+   */
+  mode?: 'show' | 'showHide'
+  /**
+   * Percentages resolve against the root, so `100% 0%` gives a look-ahead of one
+   * root height above and below without any horizontal buffer.
+   */
   rootMargin?: string
   onVisibleCallback?: () => void
   onHiddenCallback?: () => void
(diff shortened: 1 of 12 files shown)


packages/design-system/src/composables/useIsVisible/index.ts
```diff
@@ -2,13 +2,28 @@ import { Ref, onBeforeUnmount, ref, watch } from 'vue'
 
 export const useIsVisible = ({
   target,
+  root,
   mode = 'show',
-  rootMargin = '100px',
+  rootMargin = '100% 0%',
```
kulmann:
```suggestion
  rootMargin = '100% 0',
```

css `0` values typically don't hold a unit

packages/web-pkg/src/composables/filesList/useFilesViewScrollContainer.ts
```diff
@@ -0,0 +1,20 @@
+import { Ref, onMounted, ref, unref } from 'vue'
+
+const scrollContainerId = 'files-view-wrapper'
+
+/**
+ * Resolves the scrollable element the files views are rendered in. The lookup is
+ * repeated on mount, because on the very first render the container might not in
```
kulmann:
```suggestion
 * repeated on mount, because on the very first render the container might not be in
```

kulmann (approved):

packages/web-pkg/src/components/FilesList/ResourceTile.vue
```diff
@@ -178,6 +183,7 @@ const isResourceSelected = computed(() => resourcesStore.selectedIdsSet.has(reso
 
 const observerTarget = useTemplateRef<InstanceType<typeof OcCard>>('observerTarget')
 const observerTargetElement = computed<HTMLElement>(() => unref(observerTarget)?.$el)
+const scrollContainer = useFilesViewScrollContainer()
```
kulmann, line 193: oh wait, why don't you pass down the scrollcontainer from the ResourceTiles component like you did with the table?

kulmann (changes requested):

packages/design-system/src/composables/useIsVisible/index.ts
```diff
@@ -2,13 +2,28 @@ import { Ref, onBeforeUnmount, ref, watch } from 'vue'
 
 export const useIsVisible = ({
   target,
+  root,
```
kulmann, line 5: strictly speaking, since `root` is required this is a breaking change for the design-system

packages/web-pkg/src/components/FilesList/ResourceTile.vue
JammingBen: It needs to be passed to `useIsVisible`, not to the component. The problem with the table is that `useIsVisible` lives deep inside the `tr`s, hence we need to pass it down further in that case.

packages/design-system/src/composables/useIsVisible/index.ts
JammingBen: Good thing that a major release is around the corner 🤓

packages/design-system/src/composables/useIsVisible/index.ts
JammingBen: No this doesn't work here. It's not a css value, it's an option of the `IntersectionObserver`.

kulmann (approved):

PR #3229, opened 2026-08-25
Reload previews after view mode changes to avoid blurry tiles

### LINKED ISSUE #3228: Regression: images stay blurry after switching from Table to Tiles view in Spaces and Personal
## Initial Checklist

## Bug Description

Regression: image previews remain low-resolution after switching view modes.  
When starting in Table view and then switching to Tiles view, thumbnails in Spaces stay blurry and are not reloaded in higher resolution.  
The same behavior is also observable in Personal.

## Reproduction Steps

1. Navigate to **Spaces**.
2. Initial view option must be **Table**.
3. Switch the view option to **Tiles**.
4. Observe that images are extremely blurry and are not reloaded in higher resolution.
5. Repeat the same in **Personal** (same result).

## Expected Outcome

After switching to Tiles view, image thumbnails should be loaded/rendered at an appropriate higher resolution for the tile size, resulting in sharp previews.

## Actual Outcome

After switching from Table to Tiles, thumbnails remain very blurry and are not refreshed with higher-resolution image data.

AlexAndBear:
## Description

This bugfix resolves a regression where previews could remain low-resolution after switching view modes (for example from table to tiles) in Spaces and Personal.

`useLoadPreview` now tracks the loaded preview profile per resource (processor + dimensions) and only reuses an existing thumbnail if the requested profile matches.  
When the profile changes due to a view-mode change, the preview is loaded again with the correct dimensions/processor.

The test suite was extended with regression coverage for:
- reusing cached thumbnails only for the same profile
- reloading previews when the view mode changes and requires a different profile

## Related Issue

- Fixes #3228

## How Has This Been Tested?

- test environment: local development environment (pnpm workspace)
- test case 1: `pnpm test:unit --run packages/web-pkg/tests/unit/composables/resources/useLoadPreview.spec.ts`
- test case 2: `pnpm test:unit --run packages/web-app-files/tests/unit/views/spaces/Projects.spec.ts`

## Types of changes

- [x] Bugfix

### DIFF
--- a/packages/web-pkg/src/composables/resources/useLoadPreview.ts
+++ b/packages/web-pkg/src/composables/resources/useLoadPreview.ts
@@ -34,9 +34,15 @@ type LoadPreviewOptions = {
   updateStore?: boolean
 }
 
+type PreviewProfile = {
+  dimensions: [number, number]
+  processor: ProcessorType
+}
+
 type QueuedPreview = {
   controller: AbortController
   isRunning: boolean
+  profile: PreviewProfile
 }
 
 export const useLoadPreview = (viewMode?: Ref<string>) => {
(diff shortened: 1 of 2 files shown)


JammingBen (changes requested):
Hmm I don't like this solution because it adds another cache layer on top of the cache layer we already have. And it's scoped to the composable, which gets invoked more than once.

I think we should just remove those 4 lines and let the preview cache handle things:

```
// store-backed previews are loaded once, direct callers may want other dimensions
if (updateStore && resource.thumbnail) {
  return resource.thumbnail
}
```

This also solve another nasty bug where previews would be completely broken once the cache budget gets full and you re-sort the file list. Then the initially loaded but now removed previews are broken.

PR #3232, opened 2026-08-26
feat: redesign plain layout pages

JammingBen:
Redesign the plain layout pages and reduce the flicker after logging in and out and when resolving private links by:

- matching the initial loading spinner and the `AppLoadingSpinner`.
- matching the background colors of the initial page render and the loaded theme.
- introducing a new bare layout that only renders the `AppLoadingSpinner` to avoid rendering some text like `You are being logged in` for a split second.

Also, make using `dark:` prefixes on Tailwind classes work by correctly setting `document.documentElement.dataset.colorScheme`.

image

### DIFF
--- a/packages/web-runtime/src/composables/layout/useLayout.ts
+++ b/packages/web-runtime/src/composables/layout/useLayout.ts
@@ -1,3 +1,4 @@
+import LayoutBare from '../../layouts/Bare.vue'
 import LayoutPlain from '../../layouts/Plain.vue'
 import LayoutApplication from '../../layouts/Application.vue'
 import { computed, unref } from 'vue'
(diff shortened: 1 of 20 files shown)


JammingBen:
@kulmann I'm a bit undecided on the `You are being logged in` part. It previously appeared directly after login for a split second (depending on the network). But it doesn't really add much value and just states the obvious (if the user is fast enough to read it in the first place). So I just expanded the loading spinner a bit longer instead. What do you think?

kulmann:
> @kulmann I'm a bit undecided on the `You are being logged in` part. It previously appeared directly after login for a split second (depending on the network). But it doesn't really add much value and just states the obvious (if the user is fast enough to read it in the first place). So I just expanded the loading spinner a bit longer instead. What do you think?

Yes, much better. Having a text that is only briefly there might cause the typical confusion of "I was not fast enough to read it, what was that? did something bad happen?". loading spinner is sufficient. 👍

packages/web-runtime/src/composables/layout/useLayout.ts
```diff
@@ -9,24 +10,18 @@ export interface LayoutOptions {
...
-      'oidcCallback',
-      'resolvePublicLink',
-      'accessDenied'
-    ]
-    if (
-      !unref(router.currentRoute).name ||
-      plainLayoutRoutes.includes(unref(router.currentRoute).name as string)
-    ) {
+  const layoutType = computed<'bare' | 'plain' | 'application'>(() => {
+    const bareLayoutRoutes = ['login', 'oidcCallback']
```
kulmann, line 26: only looking at the names, what's the difference between "Bare" and "Plain"? Why do they co-exist? When do I use the one and when the other? Can you add a brief explanation here, which one is responsible for what?

packages/web-runtime/src/layouts/Plain.vue
```diff
@@ -1,51 +1,24 @@
 <template>
-  <div class="oc-login h-screen" :style="backgroundImgStyle">
-    <div v-if="pathIsRoot" class="flex items-center justify-center h-full bg-role-surface">
-      <oc-spinner size="large" :aria-label="$gettext('Loading')" />
-    </div>
-    <template v-else>
+  <plain-shell class="oc-login" :title="pageTitle">
```
kulmann: the `oc-login` class doesn't seem to be used as a test selector and it is not defined anywhere anymore. Remove?

kulmann (approved):
really nice improvement 🤓

PR #3235, opened 2026-08-26
feat(text-editor): show collaboration-ready indicator in toolbar

AlexAndBear:
## Description

image

- Adds a collaboration-ready indicator to the text editor toolbar.
- The indicator is shown as the last toolbar icon (with spacing) when collaboration is active and the Yjs session status is "connected".
- Adds tooltip and accessibility label text: "Collaboration ready".
- Threads "yjsStatus" from AppWrapper slot props into useTextEditor, and exposes "collaborationStatus" on the injected text editor instance.
- Updates unit tests for AppWrapper, useTextEditor, TextEditorToolbar, and web-app-text-editor app wiring.

## Related Issue

- Fixes N/A (follow-up enhancement related to collaborative editing UX)

## How Has This Been Tested?

- test environment: local development environment (macOS), pnpm + Vitest
- test case 1: ran unit tests for affected specs in AppWrapper, useTextEditor, TextEditorToolbar, and web-app-text-editor app
- test case 2: verified the toolbar indicator appears only for "connected", includes tooltip text, includes aria-label, and reacts to status changes
- ...

## Types of changes

- [ ] Bugfix

### DIFF
--- a/packages/web-pkg/src/editor/types.ts
+++ b/packages/web-pkg/src/editor/types.ts
@@ -4,6 +4,7 @@ import type { Resource } from '@opencloud-eu/web-client'
 import type { Editor } from '@tiptap/vue-3'
 import type * as Y from 'yjs'
 import type { Awareness } from 'y-protocols/awareness'
+import type { YjsStatus } from '../composables/yjs'
 import type { EditorActionGroup } from './composables'
 
 export type ContentType = 'plain-text' | 'markdown' | 'html' | 'tiptap-json'
@@ -52,6 +53,8 @@ export interface TextEditorOptions {
    * is not also set.
    */
   awareness?: Awareness
+  /** Current transport status of the hosting Yjs session, if any. */
+  yjsStatus?: MaybeRefOrGetter<YjsStatus | null>
 }
 
 export interface TextEditorLinkPanelRequest {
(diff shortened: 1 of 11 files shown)


AlexAndBear:
This is a first iteration of the collaboration-ready indicator. 

**Open questions for follow-up:**
- Should we show an error state when the connection drops (e.g. status transitions from 'connected' to 'disconnecting' or 'error')?
- How should the indicator behave when an editor is opened without Yjs support (e.g. in vaults where collaboration is unavailable)?

**Current behavior:**
- Icon is shown only when status is 'connected'
- Icon is hidden for all other states (connecting, disconnecting, error, null)

JammingBen:
> Should we show an error state when the connection drops (e.g. status transitions from 'connected' to 'disconnecting' or 'error')?

How about "Disconnected"?

> How should the indicator behave when an editor is opened without Yjs support (e.g. in vaults where collaboration is unavailable)?

Good question, showing the red icon with e.g. "not connected" might give the impression that something is wrong when it really isn't. Because of that, I think I'd favor not showing an indicator at all then 🤔

packages/web-pkg/src/editor/types.ts
```diff
@@ -73,6 +76,8 @@ export interface TextEditorInstance {
   contentType: Ref<ContentType>
   /** Derived from the caller's `readonly` option; follows it while mounted. */
   readonly: Ref<boolean>
+  /** Current transport status of the hosting Yjs session, if any. */
+  collaborationStatus: Ref<YjsStatus | null>
```
JammingBen: I decided to name all collaboration related things `yjs`-something to be clear and consistent throughout the code, why not name it `yjsStatus` as well?

AlexAndBear:
Updated the indicator to also show the disconnected state:
- **Connected**: green wifi icon with "Collaboration ready" label
- **Disconnected**: red wifi-off icon with "Collaboration disconnected" label
- **Connecting**: gray wifiicon with "Collaboration connecting..." label
- **Hidden** for: `local`, and `null` states

Also introduced `YjsStatus` enum to replace string literals for better type safety.

JammingBen (approved):

PR #3244, opened 2026-08-27
Persist selected language for plain layout pages

AlexAndBear:
## Description

This fixes language fallback behavior for pages rendered via the plain layout when no user context is available.

The runtime now resolves the initial language in this order:
1. user setting (applied when available and persisted)
2. stored language from local storage
3. browser language

This keeps the UI language stable on plain-layout pages such as /access-denied after users previously selected English.

## Related Issue

- Fixes N/A

## How Has This Been Tested?

- test environment: local development on macOS
- test case 1: pnpm test:unit --run packages/web-runtime/tests/unit/helpers/language.spec.ts packages/web-runtime/tests/unit/container/bootstrap.spec.ts packages/web-runtime/tests/unit/pages/account/accountPreferences.spec.ts
- test case 2: pnpm exec eslint packages/web-runtime/src/helpers/language.ts packages/web-runtime/src/container/bootstrap.ts packages/web-runtime/tests/unit/helpers/language.spec.ts
- test case 3: verified helper behavior for stored-vs-browser language fallback via unit tests

## Types of changes

- [x] Bugfix
- [ ] Enhancement (a change that does not break existing code or deployments)

### DIFF
--- a/packages/web-runtime/src/helpers/language.ts
+++ b/packages/web-runtime/src/helpers/language.ts
@@ -2,21 +2,64 @@ import { ApplicationInformation } from '@opencloud-eu/web-pkg'
 import { merge } from 'lodash-es'
 import { Language, Translations } from 'vue3-gettext'
 
+export const currentLanguageLocalStorageKey = 'oc_language'
+
+function normalizeLanguage(languageSetting: string): string {
+  const trimmed = languageSetting.trim()
+  if (!trimmed) {
+    return ''
+  }
+  return trimmed.includes('-') ? trimmed.split('-')[0] : trimmed
+}
+
+function getStoredLanguage(): string {
+  const storedLanguage = window.localStorage.getItem(currentLanguageLocalStorageKey) ?? ''
+  if (!storedLanguage) {
+    return ''
+  }
+  return normalizeLanguage(storedLanguage)
+}
+
+function storeLanguage(language: string): void {
+  window.localStorage.setItem(currentLanguageLocalStorageKey, language)
+}
+
+function setDocumentLanguage(languageSetting: string): void {
+  const currentLanguage = normalizeLanguage(languageSetting)
+  if (!currentLanguage) {
+    return
+  }
+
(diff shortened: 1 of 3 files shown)


packages/web-runtime/src/helpers/language.ts
```diff
@@ -1,22 +1,59 @@
...
+function normalizeLanguage(languageSetting: string): string {
+  const trimmed = languageSetting.trim()
+  if (!trimmed) {
+    return ''
+  }
+  return trimmed.includes('-') ? trimmed.split('-')[0] : trimmed
+}
+
+function getStoredLanguage(): string {
+  const languageStorage = useLocalStorage<string>(currentLanguageLocalStorageKey, '')
```
JammingBen: Better work with `window.localStorage` directly here, the composable is for usage inside components. It spawns a watcher and also writes asynchronously via the watcher.

packages/web-runtime/src/helpers/language.ts
```diff
@@ -1,22 +1,59 @@
...
+  const storedLanguage = unref(languageStorage)
+  if (!storedLanguage) {
+    return ''
+  }
+  return normalizeLanguage(storedLanguage)
+}
+
+function storeLanguage(language: string): void {
+  const languageStorage = useLocalStorage<string>(currentLanguageLocalStorageKey, '')
```
JammingBen: Same here.

packages/web-runtime/src/helpers/language.ts
```diff
@@ -1,22 +1,59 @@
...
+  languageStorage.value = language
+}
+
+export const resolveInitialLanguage = ({
```
JammingBen, line 36: You need to set `document.documentElement.lang` here as well, otherwise the plain layout pages might be e.g. German, but the `html` tag says `lang="en"`.

JammingBen (changes requested):

JammingBen (approved):

PR #3246, opened 2026-08-27
feat(yjs): improve handling of external file conflicts

### LINKED ISSUE #3103: Collaborative editing: handle external file updates
### Initial Checklist

### User Value
Data integrity, UX

### Acceptance Criteria
- When a peer X joins a room whose document holds unsaved changes and the server etag differs from the room's `nativeEtag`, the external content is **not** hydrated into the document. The room gets flagged as "stale" (metadata change which gets broadcasted to all peers)
- All peers with unsaved changes (everbody except X) in the room are shown a message that the document has changed externally and that they should copy their editor content into the clipboard before reloading the document (check if saving anyway causes a 412 conflict, if not we must prevent saving)
- Covered by an e2e test: edit without saving → change the file externally → second peer joins → conflict message
- E2E commented steps in `tests/e2e/features/yjs/conflicts.feature` can be un-commented

JammingBen:
## Description

Disconnect users from a yjs room if they have unsaved changes that would be overwritten by an external update. This ensures they don't lose their unsaved work and can properly save or store it somewhere, before reconnecting to the room via a page reload.

Also reflect the disconnected status in the indicator.

The second commit tackles an edge case where a user that was offline during a recovery would lose their local work upon reconnect. See down below for explanations.

## Scenarios explained

**`user a` has no unsaved changes**

- `user a` is in a room for `file.txt`
- `file.txt` gets external update
- `user b` joins room and fetches the updated file:
  -> flags the room with `isStale: true`
  -> initiates recovery
- `user a` detects the `isStale` change, stays in the room
-  Recovery finished and every peer in the room gets the updated file

**`user a` has unsaved changes**

- `user a` is in a room for `file.txt`
- `file.txt` gets external update
- `user b` joins room and fetches the updated file:
  -> flags the room with `isStale: true`
  -> initiates recovery
- `user a` detects the `isStale` change and leaves the room to avoid that their changes get overwritten
-  Recovery finished and every peer in the room gets the updated file (note that `user a` is not in the room anymore, hence no update for them)

**`user a` has unsaved changes and is offline during recovery**

- `user a` is in a room for `file.txt` but goes offline
- `file.txt` gets external update
- `user b` joins room and fetches the updated file:
  -> flags the room with `isStale: true`
  -> initiates recovery
-  Recovery finished and every peer in the room gets the updated file
- `user a` goes online again
  -> the guard on `isStale`, that would remove them from the room, never ran, meaning they stayed in the room during recovery and have the updated content now
  -> they now leave the room
  -> they recover their content from the last known state

closes https://github.com/opencloud-eu/web/issues/3103

### DIFF
--- a/packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
+++ b/packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
@@ -518,7 +518,7 @@ function showExternalUpdateConflict(response: Response = null) {
   errorPopup(
     new HttpError(
       $gettext(
-        'This file was updated outside this window. Please copy your changes or save the file under a new name (»Save As...«).'
+        'This file was updated outside this window. Please copy your changes, save the file under a new name (»Save As...«) or reload the page to discard your changes.'
       ),
       response
     )
--- a/packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
+++ b/packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
@@ -229,16 +229,25 @@ const {
   applicationId
 })
 
+const isDirty = computed(() => {
+  if (unref(isReadOnly)) {
+    return false
+  }
+  return unref(currentContent) !== unref(serverContent)
+})
+
 const {
   session: yjsSession,
   isSessionReady,
   effectiveReadOnly,
+  isConflicted,
   reconcileConflict
 } = useAppWrapperYjs({
   yjs,
   applicationId,
   resource,
   isReadOnly,
+  isDirty,
   loading,
   loadingError,
   contentResourceId,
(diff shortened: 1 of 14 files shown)


packages/web-pkg/src/components/AppTemplates/AppWrapper.vue
```diff
@@ -516,6 +514,17 @@ const autosavePopup = () => {
   showMessage({ title: $gettext('File autosaved') })
 }
 
+function showExternalUpdateConflict(response: Response = null) {
+  errorPopup(
+    new HttpError(
+      $gettext(
+        'This file was updated outside this window. Please copy your changes or save the file under a new name (»Save As...«).'
```
kulmann: Also add something like `Reload this page to discard your changes.`?

kulmann (approved):
nice 🥳

Issue #3105, opened 2026-08-13
y.js container hygiene


kulmann:
### Initial Checklist
- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story
> As an administrator, I want the y.js server container to behave like a production-grade service so that I can operate, monitor and restart it without losing user data.

### User Value
Operability, data integrity

### Acceptance Criteria
- The image defines a `HEALTHCHECK` that reflects actual readiness (process up and websocket endpoint reachable)
- A `SIGTERM` handler shuts down gracefully: stop accepting new connections, persist/flush open documents, close websockets within a documented grace period
- The compose setup declares `depends_on` with `condition: service_healthy` where applicable
- Logs go to stdout/stderr
- Restart policy and grace period are documented for administrators (in README.md of the service)

PR #3247, opened 2026-08-27
Harden yjs container readiness and shutdown behavior

AlexAndBear:
## Description

This change upgrades the `services/yjs` container behavior for production-style operation.

It adds:
- container image `HEALTHCHECK` based on actual readiness (`/healthz/ready`) and WebSocket reachability
- graceful signal handling (`SIGTERM`/`SIGINT`/`SIGQUIT`) with a documented grace period
- shutdown draining behavior: readiness flips to unhealthy, new upgrades are rejected, then active connections are closed and document unload/flush paths run
- compose-level `stop_grace_period` for `yjs`, aligned with graceful shutdown behavior
- administrator documentation for restart policy, grace period, health semantics, and logging

## Related Issue

- https://github.com/opencloud-eu/web/issues/3105

## How Has This Been Tested?

- test environment: local development on macOS
- test case 1: `pnpm --filter opencloud-yjs-server check:types`
- test case 2: `pnpm exec eslint services/yjs/src/server.ts services/yjs/src/healthcheck.ts`
- test case 3: `pnpm exec prettier --check services/yjs/src/server.ts services/yjs/src/healthcheck.ts services/yjs/README.md docker-compose.yml`
- test case 4: `docker compose config >/tmp/docker-compose.resolved.yml`

## Types of changes

- [ ] Bugfix
- [x] Enhancement (a change that does not break existing code or deployments)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -147,6 +147,12 @@ services:
     depends_on:
       collabora:
         condition: service_healthy
+      yjs:
+        condition: service_healthy
+      traefik:
+        condition: service_started
+      tika-service:
+        condition: service_started
 
   opencloud-federated:
     <<: *opencloud-service
@@ -176,6 +182,13 @@ services:
       - ./dev/docker/opencloud/app-registry.yaml:/etc/opencloud/app-registry.yaml
       - opencloud-federated-config:/etc/opencloud
       - ${OC_WEB_CONFIG:-./dev/docker/opencloud.web-federated.config.json}:/web/config.json:ro
+    depends_on:
+      yjs:
+        condition: service_healthy
+      traefik:
+        condition: service_started
+      tika-service:
+        condition: service_started
 
   collabora:
     image: collabora/code:26.04.2.4.1
@@ -451,6 +464,7 @@ services:
       # model identical to what CI uses (CI has no Traefik).
       - traefik
     restart: unless-stopped
+    stop_grace_period: 20s
 
 volumes:
   uploads:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 x-opencloud-server: &opencloud-service
   image: ${OC_IMAGE:-opencloudeu/opencloud-rolling:daily}
   entrypoint: /bin/sh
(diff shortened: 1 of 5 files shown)


services/yjs/src/server.ts
```diff
@@ -135,14 +141,42 @@ async function probeFileAccess(token: string, documentName: string): Promise<Fil
...
+    }
+    socket.destroy()
+    // Reject upgrade during shutdown by throwing (see onRequest comment above)
+    throw undefined
+  },
+
   async onAuthenticate({ token, documentName, connectionConfig }) {
+    if (isShuttingDown) {
+      throw new Error('server is shutting down')
+    }
```
JammingBen: I think closing the socket is better here, because throwing will send the client a wrong signal ("Permission denied").

```
socket.close(ResetConnection.code, ResetConnection.reason)
```

`ResetConnection` lives in `@hocuspocus/common` (this already is a transitive dep, needs to be included in `package.json` though so we can import it).

services/yjs/Dockerfile
```diff
@@ -31,4 +31,7 @@ COPY --from=builder /app ./
 USER node
 EXPOSE 1234
 
+HEALTHCHECK --interval=15s --timeout=5s --start-period=10s --retries=3 \
```
JammingBen: `--timeout=5s` is not enough because the health check runs `checkReadinessEndpoint` and `checkWebSocketEndpoint` sequentially, each having a `5s` timeout. So this should either be double here or the timeout for each check in `healthcheck.ts` should be halved.

AlexAndBear: Let's increase it to 10 + add 2 sec buffer

docker-compose.yml
```diff
@@ -451,6 +451,7 @@ services:
       # model identical to what CI uses (CI has no Traefik).
       - traefik
     restart: unless-stopped
+    stop_grace_period: 20s
```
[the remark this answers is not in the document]
AlexAndBear: picky af 🙄

JammingBen (approved):

PR #3253, opened 2026-08-28
refactor!: drop mark.js

### LINKED ISSUE #2966: Replace mark.js with render-based filter highlighting
## Why

Filter highlighting currently uses mark.js, which mutates Vue-rendered DOM. This caused a crash in the sidebar member panels (`Cannot read properties of null (reading 'insertBefore')`): unwrapping highlights called `Node.normalize()`, which removes the empty text nodes Vue uses as `v-for`/`v-if` fragment anchors. That was fixed in `useFilterHighlight`, but the underlying conflict remains at 13 call sites.

Additional problems with the current approach:

- Untestable because of `happy-dom`'s `NodeIterator`.
- Silent DOM corruption. mark.js' `unwrapMatches` calls `parent.normalize()`, which detaches the text node Vue may hold.
- mark.js is frozen since 2022-06 and we use one of its features.

Proposal: highlight during render instead. `defaultFuseOptions` uses `threshold: 0`, so matches are always literal case-insensitive substrings, meaning a ~20 line split helper plus a small component replaces the library and roughly 200 lines of imperative glue (template refs, Mark instances, post-flush watchers).

## Scope

| group | sites | work |
| --- | --- | --- |
| local markup|  app-store AppTile/AppTags, ContactsList, CreateShortcutModal, AutocompleteItem, admin-settings, Users/Groups/Spaces lists (via OcTable field slots) | drop-in |
| files list| trash/Overview, SharedViaLink, SharedWithMe, SharedWithOthers, spaces/Projects | one optional term prop through ResourceTable/ResourceTiles → ResourceListItem → ResourceName |
| slot/extension content| ItemFilter (expose term to its item slot), SearchBar previews (already receive term, so previews highlight themselves) | see blow |

- `ItemFilter` exposes the term to its item slot (`:term="filterTerm"`); the 6 consumer templates in `files/Search/List.vue`, `SharedWithMe.vue`, `SharedWithOthers.vue` and `admin-settings/Users.vue` use it explicitly. No provide/inject: highlighting must not depend on ambient context, or `ResourceName` starts highlighting where nobody asked.
- Search preview components highlight themselves via the term prop they already receive; `SearchBar` stops highlighting centrally. Built-in `ResourcePreview` is covered by the files list work. Third-party providers opt in; document it for extension authors.

JammingBen:
Drop `mark.js` from the codebase and highlight via the new `FilterHighlight` component. This has several advantages:

  - Hardened highlighting; the old implementation had a few quirks where highlighting would not work reliably (e.g. when entering a filter term that yields results, then adding and removing random letters).
  - Less boilerplate when using marks in components.
  - Less external dependencies.

### Breaking for developers

- `mark.js` has been removed: use `FilterHighlight.vue` from design-system instead.
- `useFilterHighlight` has been removed: use `FilterHighlight.vue` from design-system instead.

fixes https://github.com/opencloud-eu/web/issues/2966


AlexAndBear (approved):

AlexAndBear:
LGTM, we could possibly move the FilterHighlight component to the design-system, no ? 🤔

PR #3254, opened 2026-08-28
fix: remove system actions from getAllOpenWithActions

JammingBen:
They don't belong there, especially since they included delete and restore actions. The fallback to the download action in case no other action was found is already present in `getDefaultAction`.

Also remove the deprecated `useOpenWithDefaultApp` composable in the course of this and in preparation of the upcoming major release.


packages/web-pkg/src/composables/actions/files/useFileActionFallbackToDownload.ts
```diff
@@ -23,20 +23,18 @@ export const useFileActionFallbackToDownload = () => {
...
-      label: () => {
-        return $gettext('Download')
-      },
-      isVisible: (options) => {
-        return unref(downloadFileActions)[0].isVisible(options)
-      },
-      class: 'oc-files-actions-fallback-to-download-trigger'
-    }
-  ])
+  const actions = computed((): FileAction => ({
```
kulmann: you should rename it to "downloadAction" then if it's only the one action and the return type also is not an array anymore.

kulmann: oh wait, that's the object that gets returned from the composable. then it should still be `FileAction[]` as return type, without the rename.

AlexAndBear (approved):

PR #3261, opened 2026-08-29
fix: share the tooltip escape listener

### LINKED ISSUE #3171: OcTooltip registers a document keydown listener per element
Found while investigating #3159.

## Current situation

The `v-oc-tooltip` directive (`packages/design-system/src/directives/OcTooltip.ts`) registers a `keydown` listener on `document` for every element it is bound to, and only removes it on unmount.

File list rows are never unmounted once revealed, so the listener count grows with every item scrolled past.

Measured on a 500-row table:

| | `document` keydown listeners |
| --- | --- |
| after initial load | 135 |
| after scrolling through all 500 rows | 3021 |

That is about 6 per row.

## Why this is bad

Every keystroke anywhere in the app fans out to three thousand handlers, each of which only needs to check for `Escape`. It gets worse the longer a session runs and the more large folders are visited. There is no upper bound.

## Proposed solution

Register a single shared `keydown` handler for the directive that closes whichever tooltip is currently open, instead of one handler per bound element. The per-element `mouseenter`/`mouseleave`/`focus`/`blur`/`click` listeners can stay as they are.

## Verification

`document` keydown listener count stays constant across a full scroll pass through a large folder.

fredrikblau:
## Description

Each tooltip instance currently registers its own document-level Escape key listener. This adds one global listener per mounted tooltip and removes it only with that instance.

This change shares one Escape listener across all mounted tooltips and removes it after the last tooltip is unmounted. Tooltip behavior is unchanged.

## Related Issue

- Fixes https://github.com/opencloud-eu/web/issues/3171

## How Has This Been Tested?

- test environment: Node.js 26.7.0, pnpm 11.22.0
- test case 1: `pnpm test:unit --run packages/design-system/src/directives/OcTooltip.spec.ts`
- test case 2: `pnpm check:types`
- test case 3: `pnpm lint`
- test case 4: `pnpm format:check`

## Types of changes

- [x] Bugfix
- [x] Tests (adding or updating tests)

### DIFF
--- a/packages/design-system/src/directives/OcTooltip.ts
+++ b/packages/design-system/src/directives/OcTooltip.ts
@@ -40,6 +40,10 @@ const showTooltip = async (el: HTMLElement) => {
 
   document.body.appendChild(tooltipEl)
   data.tooltipEl = tooltipEl
+  tooltipElements.add(el)
+  if (tooltipElements.size === 1) {
+    document.addEventListener('keydown', documentEscapeHandler)
+  }
 
   const { x, y, placement, middlewareData } = await computePosition(el, tooltipEl, {
     placement: 'top',
@@ -80,6 +84,10 @@ const hideTooltip = (el: HTMLElement) => {
 
   data.tooltipEl.remove()
   data.tooltipEl = null
+  tooltipElements.delete(el)
+  if (tooltipElements.size === 0) {
+    document.removeEventListener('keydown', documentEscapeHandler)
+  }
 }
 
 const destroy = (el: HTMLElement) => {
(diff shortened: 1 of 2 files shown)


packages/design-system/src/directives/OcTooltip.ts
```diff
@@ -7,10 +7,18 @@ interface TooltipData {
   showHandler: () => void
   hideHandler: () => void
   clickHandler: () => void
-  escapeHandler: (e: KeyboardEvent) => void
 }
 
 const tooltipMap = new WeakMap<HTMLElement, TooltipData>()
+const tooltipElements = new Set<HTMLElement>()
```
JammingBen, line 13: This seems to hold all tooltip elements, not just the visible ones, meaning ESC always loops through all of them. I think it would be better to only track visible tooltips (move tracking and event (un-)registering to `showTooltip` and `hideTooltip`).

JammingBen (changes requested):

fredrikblau:
Updated as suggested: tooltip elements are tracked only while visible, with the shared Escape listener registered on show and removed after the last visible tooltip is hidden.

Validated with the focused OcTooltip test, formatting check, and ESLint.

JammingBen (approved):
Thanks!

PR #3263, opened 2026-08-29
fix: define the shimmer animation start frame

### LINKED ISSUE #3169: File list: one IntersectionObserver and one shimmer animation per item

## Current situation

Every lazy list item creates its own `IntersectionObserver` and runs its own shimmer animation.

In a 500-item folder that means 500 live observers and 481 running CSS animations, while only about 20 placeholders are actually on screen. All animations start at page load, so they sweep in unison instead of reading as per-item loading.

The shimmer also barely reads as animated in Safari:

- `@keyframes shimmer` (`packages/design-system/src/styles/tailwind.css`) only declares the `100%` state and relies on an implicit start frame taken from the `after:transform-[translateX(-100%)]` utility class
- the gradient is white at `opacity-10` (table) / `opacity-20` (tiles), which is very low contrast
- the sweep spends roughly half its cycle translated off the box

Measured on a 500-item folder, disabling the shimmer animation only:

| View | Engine | Baseline | Animation off |
| --- | --- | --- | --- |
| Tiles | WebKit | 80 ms | 72 ms |
| Table | WebKit | 48 ms | 40 ms |

## Why this is bad

The observers and animations cost roughly 15% of the frame time in both views while scrolling, and none of that work benefits the user. On top of that, the loading indicator that is supposed to explain the delay is effectively invisible in Safari, so the list just looks broken - which is what was reported in #3159 ("renders whitespace for seconds", "not animated").

## Proposed solution

1. Replace the per-item observers in `useIsVisible` with one shared observer per list, keyed by element, so 500 items cost one observer instead of 500.
2. Add an explicit `0%` keyframe to `@keyframes shimmer`. https://github.com/opencloud-eu/web/pull/3264
3. ~Only animate placeholders that are on screen, so the animation count tracks the viewport rather than the page size.~ The impact is low, and this is basically the same as removing the shimmer completely. I'd only do this if it had real noticeable impact.

## Verification

- one `IntersectionObserver` per list regardless of item count
- `document.getAnimations().length` scales with visible placeholders, not with total items
- shimmer visibly moving in Safari

fredrikblau:
## Description

The shared shimmer animation only declares its ending transform and relies on a utility class to provide the initial state. That implicit starting frame is not consistently retained by browsers, and the loading indicator can fail to animate correctly.

Define the `0%` transform in the keyframes so the shimmer has an explicit, portable start state.

## Related Issue

- Fixes https://github.com/opencloud-eu/web/issues/3169

## How Has This Been Tested?

- test environment: Node.js 26.7.0, pnpm 11.24.0
- test case 1: `pnpm lint`
- test case 2: `pnpm format:check`

## Types of changes

- [x] Bugfix
- [ ] Tests (adding or updating tests)

### DIFF
--- a/packages/design-system/src/styles/tailwind.css
+++ b/packages/design-system/src/styles/tailwind.css
@@ -85,6 +85,10 @@
   }
 
   @keyframes shimmer {
+    0% {
+      transform: translateX(-100%);
+    }
+
     100% {
       transform: translateX(100%);
     }


JammingBen:
It seems you also included this in https://github.com/opencloud-eu/web/pull/3264, so I'm closing here.

PR #3264, opened 2026-08-29
fix: preserve lazy tile height

### LINKED ISSUE #3170: Tiles view: lazy placeholder is 4px shorter than a hydrated tile, causing layout shift

## Current situation

In the tiles view a lazy placeholder tile is 4 px shorter than a hydrated one.

Measured on a 500-image folder, tiles size 3:

- placeholder tile: 220.5625 px
- hydrated tile: 224.5625 px
- grid: 4 columns x 125 auto-sized rows

The placeholder markup in `packages/web-pkg/src/components/FilesList/ResourceTile.vue` does not reproduce the hydrated tile's box exactly.

## Why this is bad

Every reveal grows its grid row and pushes everything below it down. While scrolling, content moves under the user's finger and the scroll position drifts. Each reveal also invalidates layout for the whole 500-item grid.

## Proposed solution

1. Make the placeholder exactly the height of a hydrated tile.
2. Pin the grid row height so revealing a tile cannot resize its row.

## Verification

Placeholder and hydrated tile report the same `getBoundingClientRect().height`. Scroll position stays stable while tiles fill in - no cumulative shift.

fredrikblau:
## Description

Lazy tile placeholders are shorter than hydrated tiles by the height of the tile content's line box. Revealing a tile therefore expands its grid row and shifts the items below it while scrolling.

Give the placeholder a minimum height based on the configured tile width, the 16:9 preview ratio, and the 44px content area used by hydrated tiles.

## Related Issue

- Fixes https://github.com/opencloud-eu/web/issues/3170

## How Has This Been Tested?

- test environment: Node.js 26.7.0, pnpm 11.24.0
- test case 1: `pnpm test:unit --run packages/web-pkg/tests/unit/components/FilesList/ResourceTile.spec.ts`
- test case 2: `pnpm check:types`
- test case 3: `pnpm lint`
- test case 4: `pnpm format:check`

## Types of changes

- [x] Bugfix
- [ ] Tests (adding or updating tests)

### DIFF
--- a/packages/web-pkg/src/components/FilesList/ResourceTile.vue
+++ b/packages/web-pkg/src/components/FilesList/ResourceTile.vue
@@ -21,7 +21,7 @@
     >
       <div class="w-full aspect-[16/9]" />
       <div class="p-2">
-        <div class="h-6" />
+        <div class="h-7" />
       </div>
     </div>
     <template v-else>
@@ -243,10 +243,6 @@ if (!lazy) {
 
   .oc-tile-card-lazy-shimmer::after {
     background-image: linear-gradient(90deg, #4c5f7900 0, #4c5f7933 20%, #4c5f7980 60%, #4c5f7900);
-  }
-
-  .oc-tile-card-lazy-shimmer {
-    min-height: calc(var(--oc-size-tiles-actual) * 0.5625 + 44px);
   }
 }
 </style>
--- a/packages/web-pkg/src/components/FilesList/ResourceTile.vue
+++ b/packages/web-pkg/src/components/FilesList/ResourceTile.vue
@@ -244,5 +244,9 @@ if (!lazy) {
   .oc-tile-card-lazy-shimmer::after {
     background-image: linear-gradient(90deg, #4c5f7900 0, #4c5f7933 20%, #4c5f7980 60%, #4c5f7900);
   }
+
+  .oc-tile-card-lazy-shimmer {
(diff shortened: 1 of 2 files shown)


packages/web-pkg/src/components/FilesList/ResourceTile.vue
```diff
@@ -244,5 +244,9 @@ if (!lazy) {
   .oc-tile-card-lazy-shimmer::after {
     background-image: linear-gradient(90deg, #4c5f7900 0, #4c5f7933 20%, #4c5f7980 60%, #4c5f7900);
   }
+
+  .oc-tile-card-lazy-shimmer {
+    min-height: calc(var(--oc-size-tiles-actual) * 0.5625 + 44px);
```
JammingBen: This doesn't seem to be accurate, I still get a slight jump. Inspecting the lazy tile height, it also doesn't match the loaded one.

I think you can remove the `calc` here and just replace `h-6` with `h-7` in the lazy tile's bottom section.

JammingBen (changes requested):

fredrikblau:
Updated as suggested: removed the calculated shimmer minimum height and changed the lazy tile bottom placeholder from h-6 to h-7.

Validated with the ResourceTile and ResourceTiles test suites (26 tests passed), formatting, ESLint, and diff checks.

JammingBen (approved):
Thanks!

PR #3266, opened 2026-08-29
Fix default Open action for .url files and add Text Editor support

AlexAndBear:
## Description

Before this fix, clicking "Open" on a `.url` file in the file-list context menu could result in no action at all.

Root cause: the rendered action was treated as a route link as soon as a `route` callback existed, even when that callback returned no target. In that case, the handler-backed action ("Open shortcut") was never executed.

What changed:
- In `ActionMenuItem`, route/href targets are resolved first.
- A route link is rendered only when `route(...)` returns a concrete target.
- If no route target exists, the action falls back to button behavior and executes the handler.
- `.url` is now also registered for the Text Editor, so shortcut files can be opened as plain text via "Open with".

## Related Issue

- Fixes N/A

## How Has This Been Tested?

- test environment: local dev environment (macOS, Node/pnpm workspace)
- test case 1: `pnpm --filter @opencloud-eu/web-pkg check:types`
- test case 2: `pnpm test:unit --run packages/web-pkg/tests/unit/components/ContextActions/ActionMenuItem.spec.ts`
- test case 3: `pnpm test:unit --run packages/web-app-text-editor/tests/unit/yjs.spec.ts packages/web-app-text-editor/tests/unit/app.spec.ts`

## Types of changes

- [x] Bugfix
- [x] Enhancement (a change that does not break existing code or deployments)

### DIFF
--- a/packages/web-app-text-editor/src/index.ts
+++ b/packages/web-app-text-editor/src/index.ts
@@ -51,6 +51,7 @@ export default defineWebApplication({
         { extension: 'asm', label: () => $gettext('Assembler source file') },
         { extension: 'log', label: () => $gettext('Log file') },
         { extension: 'ics', label: () => $gettext('Calendar file') },
+        { extension: 'url', label: () => $gettext('Internet shortcut file') },
         { extension: 'rtf', label: () => $gettext('Rich Text Format file') },
         { extension: 'dockerfile', label: () => $gettext('Dockerfile') },
         { extension: 'makefile', label: () => $gettext('Makefile') },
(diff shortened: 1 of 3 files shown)


AlexAndBear:
@kulmann this bug has likely been around for quite a while. It appears to have been introduced when we improved the `Open / Open with...` context-menu actions.
At that time, `Open` was not shown for `.url` files in the context menu (see demo). 

On current main, `Open` now appears for .url files (which is good), but we still need to make `ActionMenuItem` resolve the correct execution path when both handler and route are present.

That combination is an edge case coming from `useFileActionsOpenWithDefault`, and right now it can lead to Open not executing the expected handler.

Also added text editor support **because I just felt like it** 😎

JammingBen (approved):

PR #3272, opened 2026-08-31
feat(web-pkg): support markdown YAML frontmatter in the text editor

### LINKED ISSUE #3256: text-editor corrupts YAML frontmatter on save — even when nothing was edited
### Describe the bug

The built-in Markdown editor in OpenCloud Web rewrites files through Tiptap on save. YAML frontmatter is not preserved: the closing `---` delimiter is dropped, the first key becomes an `##` heading, and the remaining keys are left as body text.

No warning is shown — the file still opens fine afterwards, so the loss is easy to miss.

### Steps to reproduce

1. Upload a `.md` file with YAML frontmatter
2. Open it in the web Markdown editor
3. Change any word in the body
4. Save

**Before**

```
---
status: Bereit
bereich: test
frist: 2026-08-31
---

## Kontext
Some text.
```

**After**

```
---

## status: Bereit
bereich: test
frist: 2026-08-31

## Kontext
Some text.
```

### Expected behavior

Frontmatter should be passed through untouched.

### Impact

Frontmatter is the standard metadata format for Markdown — Obsidian properties, Hugo/Jekyll/Astro page metadata, Zettelkasten setups. Any such file edited once through the web editor loses its metadata silently. In my own storage this affects 418 files, including a task board that filters on those fields: a task edited in the browser simply disappears from the board.

Tables, headings and body text survive the round-trip correctly — only frontmatter is affected. Likely related: Tiptap's Markdown extension does not support frontmatter (ueberdosis/tiptap discussion #7155).

### Environment

- OpenCloud server 7.4.0, Web UI, built-in Markdown editor
- Reproduced 2026-08-28

kulmann:
## Description

Adds a frontmatter custom node and integrates it into markdown toolbar and slashcommand-menu:

- Frontmatter only valid on markdown content
- Frontmatter can be toggled off (= delete with confirmation) or on (= if already present: jump cursor into the box, if not present yet: create one at the top of the document)
- pressing delete in front of or after the frontmatter just moves the cursor into the frontmatter-box to avoid accidental deletion
- all the markdown goodies are disabled when the cursor is in the frontmatter-box, only plain text (and emojis) are allowed in there
- frontmatter content is not interpreted as YAML (at least not in this PR for the sake of simplicity), just rendered and manipulated as plain text

## Related Issue

- Fixes #3256 

## How Has This Been Tested?

- unit tests
- manually standalone
- manually with y.js session

## Types of changes

- [x] Bugfix

### DIFF
--- a/packages/web-pkg/package.json
+++ b/packages/web-pkg/package.json
@@ -99,6 +99,7 @@
     "lodash-es": "^4.17.23",
     "lowlight": "^3.3.0",
     "luxon": "^3.7.2",
+    "marked": "catalog:tiptap",
     "oidc-client-ts": "^3.5.0",
     "p-queue": "^9.1.0",
     "password-sheriff": "^2.0.0",
--- a/packages/web-pkg/src/editor/components/CodeBlockComponent.vue
+++ b/packages/web-pkg/src/editor/components/CodeBlockComponent.vue
@@ -1,9 +1,10 @@
 <template>
   <node-view-wrapper class="code-block">
     <div class="text-editor-code-block-language" contenteditable="false">
+      <oc-icon name="code-box" size="xsmall" fill-type="line" />
       <select
         v-model="selectedLanguageValue"
-        class="text-editor-code-block-select font-mono"
+        class="text-editor-code-block-select"
         :aria-label="$gettext('Code language')"
         :disabled="isReadonly"
       >
@@ -78,6 +79,9 @@ const isReadonly = computed(() => props.editor.isEditable === false)
   position: absolute;
   right: 12px;
   top: 4px;
+  display: inline-flex;
+  align-items: center;
+  color: var(--oc-role-on-surface-variant);
 }
 
 .text-editor-code-block-select:focus {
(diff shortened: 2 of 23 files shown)


JammingBen (changes requested):
One somewhat critical finding: having something between two `---` inside a quote gets erased. E.g. a markdown file with:

```
> ---
> a: 1
> ---

Foo
```

Will result in `Foo`.

Some smaller findings, still worth looking into IMO:

- Keybinding keep working inside frontmatter. E.g. `ctrl/cmd + alt + 1` destroys the frontmatter and makes it a headline.
- Claude found that not passing a custom `marked` instance to `MarkdownManager` will make it use a singleton. This singleton gets appended every extension carrying a `markdownTokenizer` and never gets cleaned up. This leads to more and more tokenizers getting appended and running. The solution is to create a `marked` instance ourselves per strategy and pass it to the `MarkdownManager` (and possibly the markdown extension?).

AlexAndBear (approved):
😍

JammingBen (approved):

PR #3290, opened 2026-09-02
feat: improve shares navigation, checkbox contrast, search drop

JammingBen:
Slightly improve the visual appearance of the shares navigation, reduce the contrast of checkboxes and input placeholders to be less distracting while still being a11y compliant and align the search drop with other drop menus.

image


JammingBen:
@tbsbdr as discussed last week

tbsbdr (approved):
looking diamond!

PR #3293, opened 2026-09-02
fix: new button flickering during navigation

### LINKED ISSUE #3289: New button flickers
### Initial Checklist

### Bug Description

The "New" button sometimes flickers when loading a folder with lots of files and/or the system or the connection is slow.

### Reproduction Steps

- create a folder in your personal space
- throttle your network to e.g. "Slow 4G"
- open the folder

### Expected Outcome

The "New" button doesn't switch its enabled/disabled state upon navigation.

### Actual Outcome

The "New" button switches from being enabled to disabled to enabled within ~1s, causing a flicker.

JammingBen:
Do the permission checks of the FAB on the space instead of the current folder. A space stays stable while navigating its folders, so this makes the FAB not flicker during navigation.

fixes https://github.com/opencloud-eu/web/issues/3289

### DIFF
--- a/packages/web-app-files/src/extensions.ts
+++ b/packages/web-app-files/src/extensions.ts
@@ -10,6 +10,7 @@ import {
   useRouter,
   useSearch,
   useSpaceActionsCreate,
+  useSpacesStore,
   useUserStore
 } from '@opencloud-eu/web-pkg'
 import { computed, markRaw, unref } from 'vue'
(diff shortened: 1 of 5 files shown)


packages/web-app-files/src/extensions.ts
```diff
@@ -78,6 +79,19 @@ export const extensions = (appInfo: ApplicationInformation) => {
           return false
         }
 
+        // permission checks on the space are preferred over the current folder,
+        // as the current folder resets on every navigation, causing a button flicker.
+        const space = unref(currentSpace)
+        if (space) {
+          if (!isPublicSpaceResource(space)) {
+            return space.canUpload({ user: userStore.user }) !== true
```
tammi-23, line 92: The new logic always prioritizes currentSpace when it is available. However, currentSpace is also set on the Trash route and still represents the writable personal or project space. As a result, canUpload() returns true, although the current Trash resource explicitly disallows creating or uploading resources.

tammi-23 (approved):
👍

PR #3313, opened 2026-09-05
fix: load indirect shares in editors

### LINKED ISSUE #3298: Editor doesn't show indirect shares in sidebar
### Initial Checklist

### Bug Description

Being in an editor doesn't show indirect shares in the right sidebar.

### Reproduction Steps

- create a folder and share it with another user
- enter the folder and create a note
- inside the editor, open the right sidebar sharing panel

### Expected Outcome

The sidebar lists the indirect shares.

### Actual Outcome

The sidebar doesn't show any shares.

fredrikblau:
## Description

Editor routes load files without the folder loader, so ancestor metadata stays empty. The sharing sidebar then has no ancestor IDs from which to load indirect shares.

Load ancestor metadata when the share loader runs inside an editor. The regression test covers both search and editor routes.

## Related Issue

- Fixes https://github.com/opencloud-eu/web/issues/3298

## How Has This Been Tested?

- test environment: Node 24.20.0, pnpm 11.25.0
- test case 1: FileSideBar unit suite (19 passed)
- test case 2: full unit suite (3,916 passed, 6 skipped, 11 todo)
- test case 3: type checks, lint, formatting, and production build

## Types of changes

- [x] Bugfix

### DIFF
--- a/packages/web-pkg/src/composables/shares/useLoadShares.ts
+++ b/packages/web-pkg/src/composables/shares/useLoadShares.ts
@@ -163,7 +163,11 @@ export const useLoadShares = () => {
       })
     }
 
-    if (includeInheritedShares || isLocationCommonActive(router, 'files-common-search')) {
+    if (
+      includeInheritedShares ||
+      isLocationCommonActive(router, 'files-common-search') ||
+      !resourcesStore.currentFolder
+    ) {
       yield resourcesStore.loadAncestorMetaData({
         folder: unref(resource),
         space,
--- a/packages/web-pkg/tests/unit/components/sidebar/FileSideBar.spec.ts
+++ b/packages/web-pkg/tests/unit/components/sidebar/FileSideBar.spec.ts
@@ -164,9 +164,15 @@ describe('FileSideBar', () => {
       ).toHaveBeenCalledTimes(2)
     })
 
-    it('should load ancestor meta data to get indirect shares when on search page', async () => {
+    it.each([
+      [
+        'on the search page',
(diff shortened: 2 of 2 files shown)


packages/web-pkg/src/composables/shares/useLoadShares.ts
```diff
@@ -158,7 +158,7 @@ export const useLoadShares = () => {
       })
     }
 
-    if (isLocationCommonActive(router, 'files-common-search')) {
+    if (isLocationCommonActive(router, 'files-common-search') || unref(isAppActive)) {
```
JammingBen: I think we should check for the absence of `resourcesStore.currentFolder` here. If that is missing, that's an indicator that ancestor meta data needs to be loaded.

JammingBen (changes requested):
Thanks! One nitpick from my side, plus it needs a rebase.

packages/web-pkg/src/composables/shares/useLoadShares.ts
fredrikblau: Done — the condition now checks `!resourcesStore.currentFolder` instead of the app-active flag, and the branch is rebased on current `main` (kept the new `includeInheritedShares` short-circuit in front of it).

The test now covers three cases: ancestor metadata is loaded on the search page, loaded when no current folder is set, and *not* loaded when one is. The middle case fails without the new clause.

JammingBen: You can remove the check for `isLocationCommonActive(router, 'files-common-search')` now though, it's covered by `!resourcesStore.currentFolder` already.

PR #3333, opened 2026-09-08
test(e2e): repeat the global search while waiting for an indexed result

fschade:
## Description

as the title says, it retires the search instead of waiting for it, oc has a debouncer and in some cases the search request is send before oc is done indexing.

## Types of changes

- [ ] Bugfix
- [X] Tests (adding or improving tests)

### DIFF
--- a/tests/e2e/steps/ui/resources.ts
+++ b/tests/e2e/steps/ui/resources.ts
@@ -386,6 +386,8 @@ Then(
   }
 )
 
+const searchIndexTimeout = 40 * 1000
+
 When(
   /^"([^"]*)" searches "([^"]*)" using the global search(?: and the "([^"]*)" filter)?( and presses enter)?$/,
   async (
@@ -399,6 +401,7 @@ When(
     const pressEnter = !!command && command.endsWith('presses enter')
     const { page } = world.actorsEnvironment.getActor({ key: stepUser })
     const resourceObject = new objects.applicationFiles.Resource({ page })
+    world.lastGlobalSearch[stepUser] = { keyword, filter: filter as searchFilter, pressEnter }
     await resourceObject.searchResource({
       keyword,
       filter: filter as searchFilter,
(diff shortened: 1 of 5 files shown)


tests/e2e/steps/ui/resources.ts
```diff
@@ -502,9 +511,13 @@ Then(
     const resourceObject = new objects.applicationFiles.Resource({ page })
     for (const info of stepTable.hashes()) {
       if (actionType === 'should') {
-        await expect(resourceObject.getResourceSearchItemLocator(info.resource)).toBeVisible({
-          timeout: appConfig.timeout * 1000
-        })
+        await expect(async () => {
+          const locator = resourceObject.getResourceSearchItemLocator(info.resource)
+          if (!(await locator.isVisible()) && lastGlobalSearch) {
```
v-scharf: please don't use `locator.isVisible()` https://playwright.dev/docs/api/class-locator#locator-is-visible

I think this is unnecessary here. It’s unlikely that the locator will appear immediately after the second search request.

tests/e2e/steps/ui/resources.ts
```diff
@@ -399,6 +407,7 @@ When(
     const pressEnter = !!command && command.endsWith('presses enter')
     const { page } = world.actorsEnvironment.getActor({ key: stepUser })
     const resourceObject = new objects.applicationFiles.Resource({ page })
+    lastGlobalSearch = { keyword, filter: filter as searchFilter, pressEnter }
```
v-scharf: `lastGlobalSearch` is module-level state - one value for the whole file, not scoped to the actor or the scenario. If Alice searches "foo" and then Brian searches "bar", the retry on Alice's page replays Brian's keyword. It's also never reset between scenarios, so the first retry can pick up a keyword from a previous test in the same worker.

v-scharf (changes requested):

tests/e2e/steps/ui/resources.ts
fschade: good catch, thanks

v-scharf (approved):

Issue #2875, opened 2026-07-14
[FR] Open Graph tags for previews


wo2wz:
### Initial Checklist

- [x] I understand this is a feature request and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Idea

add meta tags to applicable public share/preview links compliant with the [Open Graph Protocol](https://ogp.me/) for convenient sharing in other websites, like Nextcloud has implemented in their public shares. this is a small enhancement that, on supportive websites, makes sharing links prettier as well as giving cursory information about the share.

example:
```
<meta property="og:title" content="foo.mp4"/>
<meta property="og:description" content="OpenCloud - Excellent file sharing"/>
<meta property="og:site_name" content="OpenCloud"/>
<meta property="og:url" content="https://opencloud.example.com/s/abcxyz"/>
<meta property="og:type" content="website"/>
<meta property="og:image" content="https://opencloud.example.com/thumbnail"/>
<meta property="og:video" content="https://opencloud.example.com/video.mp4"/>
<meta property="og:video:type" content="video/mp4"/>
```

kulmann:
Great idea, thank you! We'd be happy to accept a pull request for this if we don't find the time to work on this in the mean time.

kulmann:
Turns out: this doesn't make sense without server side rendering... which we don't have at the moment. Crawlers usually don't execute JavaScript.

PR #3334, opened 2026-09-08
feat: add Open Graph metadata for public previews

tammi-23:
This PR implements https://github.com/opencloud-eu/web/issues/2875
and add Open Graph tags for previews

### DIFF
--- a/packages/web-runtime/src/helpers/meta.ts
+++ b/packages/web-runtime/src/helpers/meta.ts
@@ -13,3 +13,89 @@ export function injectGeneratorMeta(capabilityStore: CapabilityStore) {
   }
   meta.content = content
 }
+
+export interface OpenGraphMetaOptions {
+  title: string
+  siteName: string
+  url: string
+  description?: string
+  image?: string
+  imageAlt?: string
+  video?: string
+  videoType?: string
+  audio?: string
+  audioType?: string
+}
+
+function setMetaProperty(property: string, content?: string) {
+  let meta = document.querySelector<HTMLMetaElement>(`meta[property="${property}"]`)
+
+  if (!content) {
+    meta?.remove()
+    return
+  }
+
+  if (!meta) {
+    meta = document.createElement('meta')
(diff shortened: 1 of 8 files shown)


dschmidt:
Oh wow, this works? I was always convinced this would only work with server side rendering

packages/web-runtime/src/helpers/meta.ts
```diff
@@ -13,3 +13,89 @@ export function injectGeneratorMeta(capabilityStore: CapabilityStore) {
...
+      return /\.(png|jpe?g)$/i.test(new URL(url, document.baseURI).pathname)
+    } catch {
+      return false
+    }
+  })
+}
+
+/**
+ * Injects Open Graph tags (https://ogp.me) describing the currently displayed page, most notably
+ * public links. Only consumers that execute JavaScript pick these up; preview-specific tags for
```
kulmann, line 71: did you do research if crawlers nowadays **do** execute JS? Otherwise this whole PR would be kind of pointless. 🙈

dschmidt: I mean, it can probably be tested easily... just paste a link into whatsapp or any other app that renders previews

Would be good to at least find *one* that works with your approach

kulmann:
Unfortunately a dead end. For most of the (relevant) crawlers we'd need server side rendering, which we don't have.

dschmidt:
wondering if it could maybe be added to the web service in the backend? 

Would need to make the backend aware of routes in the frontend so, it definitely needs tough consideration, but it should be possible and it would be a neat feature to have

kulmann:
> wondering if it could maybe be added to the web service in the backend?
> 
> Would need to make the backend aware of routes in the frontend so, it definitely needs tough consideration, but it should be possible and it would be a neat feature to have

we were already discussion adopting the BFF pattern at some point in time, at least to move authentication flows away from the client side. I think that would be a good time to also implement this feature here.

dschmidt:
Full ack!

Issue #3110, opened 2026-08-13
Show avatars of users in the current y.js editing session


kulmann:
### Initial Checklist
- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story
> As a user, I want to see who else is editing the document right now so that I know who I am collaborating with.

### User Value
UX, collaboration awareness

### Acceptance Criteria
- Stacked avatars of the users in the current session are shown in the top right of the editor toolbar (show max 3 + n)
- Avatars (and connection status) are always visible on the right side of the toolbar, user never needs to scroll
- The list is derived from y.js awareness and updates when peers join or leave
- Clicking on the avatar-stack opens a dropdown with Avatars+display names; own user always at the top
- Each user's avatar border-colour (+ background color when falling back to initials) matches the colour of their remote cursor/selection

PR #3336, opened 2026-09-09
feat(editor): rework toolbar

### LINKED ISSUE #3183: Tiptap: horizontal overflow in toolbar not usable
### Initial Checklist

### Bug Description

When the browser window is not wide enough for all tiptap toolbar elements that we have, the editor toolbar currently becomes horizontally scrollable. This has two issues:
1. for users with a mouse that doesn't support horizontal scrolling this is not (easily) usable
2. although we have a shadow on the overflowing end, it is not that easily noticeable that the toolbar needs to be scrolled to get to the toolbar items which are currently out of reach.

### Reproduction Steps

1. Have a browser window with e.g. ~650px width
2. Open a .md or .ocnote file

### Expected Outcome

All toolbar items can be reached without thinking too hard how to get there

### Actual Outcome

Overflowing toolbar items seem to not exist

### Idea how to solve

We could collapse the overflowing toolbar items into a three-dots-button which only appears if there's not enough horizontal space. This would have another nice effect: when we have connection status and editor avatars for collaborative editing, we could keep those always visible on the very right and only collapse the toolbar items for interacting with the content into the three-dots-button.

JammingBen:
Rework the editor toolbar by putting exceeding items into a bubble menu and making the connection status always visible on the right side.

works towards https://github.com/opencloud-eu/web/issues/3110
fixes https://github.com/opencloud-eu/web/issues/3183

image


tammi-23 (commented):
1. If I´m inside "Search and replace" and click on "match case" or "whole word" the oc-drop is closed
2. Same for choosing an other emoji tab/category
3. Tabhandling is broken for submenu

JammingBen:
> * If I´m inside "Search and replace" and click on "match case" or "whole word" the oc-drop is closed
> * Same for choosing an other emoji tab/category

Good catch, fixed 👍 

> 3. Tabhandling is broken for submenu

Drop menus generally work with arrow keys instead of tab, which is also the case here. Or what exactly do you mean?

packages/design-system/src/components/OcDrop/dropRegistry.ts
kulmann, line 1: please add unit tests :-) or did I overlook them?

kulmann (approved):

PR #3364, opened 2026-09-14
feat: add tooltip for fab when creating spaces is disabled

### LINKED ISSUE #3231: Spaces: Greyed out button not work when not having creation permissions
I did not realize I cannot press this button because I lack space creation permissions.
I then tried several browsers / accounts untril I understood this.
I did not perceive the button as greyed out.

It would be nice to have one of those:
A) tooltip (_but maybe I wouldn't have noticed it)_
B) grey it out BUT on click it creates a "No Permision" message? 
C) don't grey it out, on click it creates a "No Permision" message? 

Image

AlexAndBear:
## Related Issue

- Fixes https://github.com/opencloud-eu/web/issues/3231

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bugfix


packages/web-app-files/src/extensions.ts
```diff
@@ -67,6 +69,14 @@ export const extensions = (appInfo: ApplicationInformation) => {
       type: 'floatingActionButton',
       icon: 'add',
       label: () => $gettext('New'),
+      tooltip: () => {
+        if (
+          isLocationSpacesActive(router, 'files-spaces-projects') &&
```
kulmann, line 74: didn't we want to avoid location checks completely now because of inefficiency-over-9000? 🤔 @JammingBen

AlexAndBear: there is no way around and I don't think that on additonal check will harm, since the FAB gets only rendered once instead of `n` times, as in the table.

JammingBen: Whenever possible, yes. I don't think it's possible here though :/ And it doesn't have much impact because it only gets called a few times per navigation (and not per file like a file action). Plus we now have a cache in place for it.

kulmann (approved):

PR #3369, opened 2026-09-14
fix(web-pkg): keep full folder names when file extensions are hidden

AlexAndBear:
## Description

This bugfix ensures that the "Show file extensions" preference only affects files and never truncates folder names.

Why this is important:
- Folder names are often semantic identifiers, not cosmetic labels.
- Hiding a folder suffix changes the visible identity of the folder and can make users open, move, or share the wrong target.
- In date/version naming conventions, the segment after the last dot is usually meaningful data.

Examples where full folder names must stay visible:
- `20.01.2026`
- `2026.01.Release`
- `Q4.2026.Archive`
- `Project.Alpha.v2`
- `Customer.ACME.2026`
- `Invoices.2026.01`

Technical changes:
- Apply extension split/render logic only for resources with `type === 'file'` in `ResourceName.vue`.
- Keep folder names with dot segments fully visible in table/list/tile rendering.
- Prevent trailing-dot titles for files without an extension (for example `README`, not `README.`).
- Add targeted unit tests for both behaviors in `ResourceName.spec.ts`.

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment: local development workspace on macOS, pnpm + Vitest
- test case 1: `pnpm test:unit --run packages/web-pkg/tests/unit/components/FilesList/ResourceName.spec.ts`
- test case 2: `pnpm test:unit --run packages/web-pkg/tests/unit/components/FilesList/ResourceTable.spec.ts`

## Types of changes

- [x] Bugfix


AlexAndBear:
@JammingBen @kulmann IHMO this is 'cirticial' and shouldn't be implemented that way, in the first place 🤷‍♀️

kulmann (approved):

PR #3371, opened 2026-09-14
Editor: preserve markdown when copying and pasting

### LINKED ISSUE #3337: Editor: copying and pasting markdown
### Initial Checklist

### User Story

As a user, I want to be able to seamlessly copy markdown from our editor and paste it into another markdown-compatible app so that I have an easy time exchanging text across different apps.

As a user, I want to be able to seamlessly paste markdown into our editor so that I have an easy time exchanging text across different apps.

### User Value

UX

### Acceptance Criteria

- Markdown content in our editor gets copied as valid markdown to the clipboard, without the need to switch to source mode
- Pasting valid markdown content into our editor keeps its markdown syntax and gets serialized properly by our editor to display correctly, without the need to switch to source mode

AlexAndBear:
## Description

Add Markdown-aware clipboard handling to the rendered Markdown editor view.

Users can now copy rendered Markdown and paste it into Markdown-aware tools while keeping the Markdown syntax. Pasting Markdown back into the editor also works from tools like IntelliJ, even when they provide both text and HTML clipboard content.

**For Markdown files we now prefer clipboard text over HTML. This means rich formatting copied from websites is not preserved unless it exists as Markdown syntax. Rich note editing should use `.ocnote`.**

## Related Issue

- Fixes #3337

## How Has This Been Tested?

- test environment: local macOS workspace
- test case 1: Markdown clipboard copy and paste unit tests
- test case 2: related Markdown editor, frontmatter, and code block unit tests
- test case 3: ESLint, TypeScript, and diff checks

## Types of changes

- [ ] Bugfix


packages/web-pkg/src/editor/extensions/markdownClipboard.ts
```diff
@@ -0,0 +1,53 @@
...
+              }
+
+              const content = slice.content.toJSON() as JSONContent[] | null
+              return editor.markdown?.serialize({ type: 'doc', content: content ?? [] }) ?? ''
+            },
+
+            handlePaste(view, event) {
+              // Let the original paste handler insert plain text inside code-like nodes.
+              // Parsing it as markdown would split pasted code into regular document blocks.
+              if (view.state.selection.$from.parent.type.spec.code) {
```
JammingBen, line 40: Probably makes sense to guard inline code marks as well here.

JammingBen (changes requested):
A paste of a single word or letter is always adding line breaks for some reason. E.g. have `foobar` in the editor, put the cursor after `foo` and paste a single letter. This results in 3 lines.

packages/web-pkg/src/editor/extensions/markdownClipboard.ts
AlexAndBear: It looks like this is an error in the native paste command, we already return false here, but the issue persists.
This is also the case on demo

JammingBen (approved):

AlexAndBear:
@JammingBen thanks for your review! Anyways, I decided to change the strategy entirely because we didn't use the native tip tap paste handler anymore and therefore we would need to care about a lot of quirks. So now we convert the pasted text to markdown and then let the the original paste handler take care. 

Which unfortantly means, this PR needs a new review with a fresh eye 👁️

JammingBen (approved):

PR #3373, opened 2026-09-15
perf: load the editor chunk lazily

### LINKED ISSUE #3217: Make editor component async
The editor component currently loads synchronous, causing Vite to put it in one chunk together with other things from `web-pkg`. Since the editor component is quite big (due to deps like ProseMirror and highlight.js), we want to move it to a dedicated chunk by making it load async.

If loading async doesn't work for some reason, we should use Vite's `codeSplitting` capabilities to achieve that.

The editor chunk should be `<750kb` (unzipped) ideally.

JammingBen:
The editor chunk (`tiptap`, `prosemirror`, `highlight.js`, `marked`) was loaded on first paint. It is now only fetched when an editor is actually opened, and it got slightly smaller.

Async loading:
- add `TextEditorViewer`, a read-only editor view exported from the `web-pkg` main entry as an async component. The announcement modal, the files list and space READMEs and the app store details now use it instead of importing the editor directly.
- the text editor app resolves its route component and Yjs adapter lazily
- mail loads the compose widget async, admin settings the announcement section

Bundle size:
- pin `highlight.js` to one version, lowlight pulled in a second copy

Note that the editor chunk is still around 1 MB (unzipped). We could get it down by a) separating our implementation from the vendor code and b) curating a list of languages we support code highlighting for. Currently, we pull in all common languages from `highlight.js`. See https://github.com/opencloud-eu/web/pull/3378 for this.

closes https://github.com/opencloud-eu/web/issues/3217


packages/web-app-mail/src/views/Inbox.vue
```diff
@@ -38,6 +37,9 @@ import { useLoadMailboxes } from '../composables/useLoadMailboxes'
 import { useLoadMails } from '../composables/useLoadMails'
 import { useLoadMail } from '../composables/useLoadMail'
 import { useMailCompose } from '../composables/useMailCompose'
+
+// async: the compose form pulls in the editor (tiptap & co)
+const MailWidget = defineAsyncComponent(() => import('../components/MailWidget.vue'))
```
kulmann, line 43: shouldn't this live below the imports?

packages/web-pkg/src/components/TextEditorViewer.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Starts fetching the editor chunk. Call it in parallel to loading the content
+ * so both are ready at the same time. Repeated calls reuse the same request.
+ */
+export async function preloadTextEditor(): Promise<void> {
+  try {
+    await loadEditor()
+  } catch {
+    // the component reports the failure
```
kulmann, line 16: Claude claims that you would need an `errorComponent` on the async component definition starting in line 24. Otherwise we would render nothing on failure...

packages/web-pkg/tests/unit/editor/components/TextEditorViewer.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { describe, expect, it, vi } from 'vitest'
+import { mount } from '@vue/test-utils'
```
kulmann: why not from web-test-helpers?

packages/web-pkg/tests/unit/editor/components/TextEditorViewer.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { describe, expect, it, vi } from 'vitest'
```
kulmann: I thought those are registered globally?

packages/web-pkg/src/components/TextEditorViewer.ts
JammingBen: I think it's cleaner to let the component handle this instead of having a generic error component that might not fit into the caller component. And generally, this only throws if the network or the bundle itself is broken, so the page is dead either way then.

kulmann (approved):

PR #3397, opened 2026-09-17
feat(yjs): immediately notify peers about external doc updates

### LINKED ISSUE #3239: Collaborative editing: immediately inform peers about external updates
### Initial Checklist

### User Story

As a user being in a collaborative editing session, I want to be informed about external file updates ASAP so that react quickly and don't spend too much time on a "dead" document.

### User Value

UX

### Acceptance Criteria

- an external file update immediately updates the current doc if there are no unsaved changes in the room
- an external file update immediately shows a conflict dialog if there are unsaved changes

Both can be solved via the SSE today.

JammingBen:
Notify peers in a yjs session about external doc updates via SSE. Note that this only works if at least on peer receives SSE events for updated files, which is e.g. not the case for share receivers.

closes https://github.com/opencloud-eu/web/issues/3239


AlexAndBear (approved):

AlexAndBear:
@JammingBen works great, I wonder if we should request a sse feature from the backend team, so share recievers will be notified too ? 🤔

JammingBen:
> @JammingBen works great, I wonder if we should request a sse feature from the backend team, so share recievers will be notified too ? 🤔

Good point, I will look if any issue already exists and create one if not 👍 

Edit: see https://github.com/opencloud-eu/opencloud/issues/3578

Issue #3111, opened 2026-08-13
Port the excalidraw app to web-extensions


kulmann:
### Initial Checklist
- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story
> As a user, I want to create and edit Excalidraw drawings collaboratively in OpenCloud so that I can sketch together with others without leaving the platform.

### User Value
UX, feature coverage

### Acceptance Criteria
- The Excalidraw extension from #2755 lives in web-extensions
- It registers for the Excalidraw file type and can create new files
- Editing is collaborative through the shared `yjs` service (same room/auth/save flow as the text editor)
- Users with a viewer role open the drawing read-only
- Saving persists the file client side and updates the etag as with text files
- the app has a README.md with details about how to set it up for collaboration
- (at least one simple e2e test drawing something simple if assertions are possible on the canvas at all)

IMPORTANT: publishing the excalidraw app is blocked until a) the y.js service has been published and b) the new collaborative app wrapper has been published in a rolling release. local development is already unblocked.

PR #3398, opened 2026-09-17
feat: share yjs with external apps

tammi-23:
Required for #3111 so the external Excalidraw app can consume the host-provided Yjs and Awareness modules as shared singletons. This prevents duplicate Yjs instances and allows the extension to use the same collaborative session flow as the text editor.


JammingBen (approved):
👍

PR #3404, opened 2026-09-18
Remove the vault unlock actions and hide the lock action for disabled spaces

AlexAndBear:
## Description

Entering a locked vault already routes to the scheme's unlock UI via the router guard (`setupVaultUnlockGuard`), so the separate `Unlock space` / `Unlock vault` context actions only duplicated that path - They are removed entirely: unlocking now happens by navigating into the space or folder.

What remains is the lock action, which is now hidden for disabled spaces. A disabled space cannot be entered, so offering to lock its vault led nowhere.

## Related Issue

No issue was filed for this.

- Fixes n/a

## How Has This Been Tested?

- test environment: unit tests via `pnpm test:unit --run packages/web-app-files`
- test case 1: the unlock actions and their extension registrations are gone; their tests were removed with them
- test case 2: the lock action is hidden for a disabled vault space, and still visible for an enabled, unlocked one
- test case 3: the lock actions stay hidden for an empty selection and for a multi-selection, without throwing
- test case 4: the existing lock cases (no vault, vault folders only, redirect out of a freshly locked vault, sibling-prefix guard) are unchanged
- test case 5: `pnpm check:types`, `pnpm lint` and `pnpm format:check` pass; 760 tests in `web-app-files` pass

## Types of changes

- [x] Bugfix


kulmann (approved):
Nice!
One additional thing that contextually fits in here, but I can also open a separate issue: I noticed that an unlocked space doesn't get auto-locked when disabling the space. Want me to open another issue for it?

AlexAndBear:
> Nice! One additional thing that contextually fits in here, but I can also open a separate issue: I noticed that an unlocked space doesn't get auto-locked when disabling the space. Want me to open another issue for it?

yes please

kulmann:
> yes please

https://github.com/opencloud-eu/web/issues/3407

PR #3405, opened 2026-09-18
Bring the source mode back as a read-only view for collaborative documents

AlexAndBear:
## Related Issue

No issue was filed for this.

- Fixes n/a

## How Has This Been Tested?

- test environment: unit tests via `pnpm test:unit --run packages/web-pkg` and `packages/web-app-text-editor`
- test case 1: the source mode action stays in the toolbar while a Yjs session is active, and is titled `Show source (read-only)`; without a session the title is unchanged
- test case 2: `sourceModeReadonly` is true for an active Yjs session (ydoc + configured `yjsServerUrl`) and for a read-only editor, false otherwise, and follows the `readonly` option when it flips
- test case 3: the source textarea is `readonly` in that case and swallows input; by default it stays editable and writes through to the document as before
- test case 4: the source view picks up document updates while a Yjs session is active, ignores them without one, and detaches its listener when it unmounts
- test case 5: source mode swaps the ProseMirror view for the source view and back
- test case 6: `pnpm check:types`, `pnpm lint` and `pnpm format:check` pass; 1788 tests in `web-pkg` and 21 in `web-app-text-editor` pass

Not verified by hand in a live collaborative session in the browser.

## Types of changes

- [x] Bugfix


packages/web-pkg/src/editor/components/TextEditorSourceView.vue
```diff
@@ -0,0 +1,69 @@
+<template>
+  <div class="text-editor-source-view flex size-full justify-center">
+    <textarea
+      ref="sourceTextarea"
+      :value="sourceContent"
+      :readonly="isSourceReadonly"
+      :aria-label="isSourceReadonly ? $gettext('Source view, read-only') : $gettext('Source view')"
+      class="w-full max-w-[800px] p-[1rem] resize-none border-0 focus:outline-none"
+      :class="{ 'cursor-not-allowed': isSourceReadonly }"
```
JammingBen: Hmm is `cursor-not-allowed` the correct class for selectable text? IMO it suggests that you can't interact at all with the content, no?

packages/web-pkg/src/editor/components/TextEditorSourceView.vue
```diff
@@ -0,0 +1,69 @@
...
+const { $gettext } = useGettext()
+
+const textEditor = editor || inject<TextEditorInstance>('textEditor')!
+const sourceTextareaRef = useTemplateRef<HTMLTextAreaElement>('sourceTextarea')
+const sourceContent = ref(textEditor.getContent())
+
+const isSourceReadonly = computed(() => unref(textEditor.state.sourceModeReadonly) ?? false)
+
+function syncSourceContent() {
+  sourceContent.value = textEditor.getContent()
```
JammingBen, line 34: I think we should add a debounce with `250ms` here, like the one we have in `useTextEditor.ts:222`. Otherwise we serialize and update on each keystroke per typing peer.

packages/web-pkg/src/editor/components/TextEditorSourceView.vue
```diff
@@ -0,0 +1,69 @@
...
+onMounted(() => {
+  sourceTextareaRef.value?.focus()
+  sourceTextareaRef.value?.setSelectionRange(0, 0)
+  sourceTextareaRef.value?.scrollTo(0, 0)
+
+  // A Yjs document keeps changing underneath this view while peers type, so
+  // follow it instead of showing the snapshot taken when it opened.
+  if (!unref(textEditor.yjsActive)) {
+    return
+  }
```
JammingBen, line 63: Unless I'm missing something, I think you need to make `yjsActive` publicly available on the `TextEditorState` interface. You could just have an immediate watcher on `sourceModeReadonly` here that (un-)registers the listener.

JammingBen (changes requested):
Also please don't forget to update the known limits in `dev/docs/yjs.md`.

packages/web-pkg/src/editor/components/TextEditorSourceView.vue
AlexAndBear: Hm, any better idea? I want to make this ro state very clear to the user

packages/web-pkg/src/editor/components/TextEditorSourceView.vue
AlexAndBear: yjsActive is already public — it's on TextEditorInstance

JammingBen: Ah sorry, I missed a "don't" 😅 What I was trying to say: it *doesn't* need to be public.

packages/web-pkg/src/editor/components/TextEditorSourceView.vue
JammingBen: The regular read-only mode has a default cursor that turns into the text cursor on hovering text. Shouldn't this be enough?

packages/web-pkg/src/editor/components/TextEditorSourceView.vue
AlexAndBear: :D got you, is there a specific reason, you don't want to have it public? Might come in handy with further implentations. Technically we also don't want/need to sync, when yjs is inactive and the document is read only due to sharing permissions

packages/web-pkg/src/editor/components/TextEditorSourceView.vue
AlexAndBear: yes, good idea, should be enough

packages/web-pkg/src/editor/components/TextEditorSourceView.vue
JammingBen: I don't have a hard opinion here tbh, I just thought why have it when we don't need it currently 🤷 But if you prefer it that way, also fine by me.

AlexAndBear: Technically yes, I prefer it, especially since I don't want to have the sync code running, when it's not needed

JammingBen (approved):

Issue #3381, opened 2026-02-04
Folder gets sorted as a file if it contains a dot: "."


kedanli:
I use OC 5.0.1 with collabora. 

## Describe the bug
If I create a new folder (doesn't matter where) and the folder name contains a dot (e.g. "Dr. Schmidt") The Icon changes from a folder icon to a spreadsheet icon. It still behaves like a folder (e.g. you could click on it to enter) but it does not look like a folder any more (icon wise).

gaedeke:
I can confirm this bug, with a slightly different symptom: in my case the
folder keeps its folder icon, but it is **sorted as a file**. A folder named
like `Zeugnisse 2018 2.Halbjahr` does not appear in the folder group at the
top of the list — it is sorted among the files by name. Renaming it to
`Zeugnisse_2018_2_Halbjahr` (underscore instead of dot) immediately fixes the
sorting, which confirms the dot in the name is the trigger.

This strongly suggests the frontend derives the resource type (or extension)
from the file name instead of using the actual folder/file resource type.
German folder names commonly contain dots ("2. Halbjahr", "Dr. Schmidt"),
so this hits real-world naming conventions.

<img width="480" height="559" alt="Image" src="https://github.com/user-attachments/assets/2f8d5ccb-6cd2-431b-9a97-9aa32c5a2de0" />
<img width="877" height="563" alt="Image" src="https://github.com/user-attachments/assets/c68ad9f3-a4b0-4d94-8865-f027a99a3888" />

rhafer:
AFAIK the sorting and the selection of the icon happens client side. So I am transferring this to web.

Note: On the latest code (main as of today) am not able to reproduce the sort issue, but creating a Folder with the `.txt` extension will cause it to get the same icon as a .txt file. Is that really intended?

gaedeke:
@micbar Sorry. I use the latest Docker-Compose Version OpenCloud 7.2.4 stable. Thank you for your working on this wonderfull project!
I only intend to sort first folders and second files. I have no issue with icons.

AlexAndBear:
> AFAIK the sorting and the selection of the icon happens client side. So I am transferring this to web.
> 
> Note: On the latest code (main as of today) am not able to reproduce the sort issue, but creating a Folder with the `.txt` extension will cause it to get the same icon as a .txt file. Is that really intended?

It's not, I created a bugfix pr for that (it's a regression, as @phil-davis reported this had been fixed a while a go and is an issue again)

[...]

PR #3410, opened 2026-09-19
Keep the folder icon for folders named like a file

AlexAndBear:
## Description

A folder named `lol.txt` showed the text file icon, because the extension is derived from the name alone and the file icon mapping was applied before the folder fallback.

Folders now only take an icon from an app that registered the extension with `type: 'folder'`, so app folders like `.vault` keep theirs and every other folder gets the folder icon.

## Related Issue

- Works towards https://github.com/opencloud-eu/web/issues/3381

## How Has This Been Tested?

- test environment: local, `pnpm test:unit --run`, `pnpm check:types`, `pnpm lint`, `pnpm format:check`
- test case 1: folder `lol.txt` gets the folder icon
- test case 2: folder with an extension an app registered for files gets the folder icon
- test case 3: folder `.vault` keeps the app icon
- test case 4: file `.txt` keeps the text icon

## Types of changes

- [x] Bugfix


JammingBen:
It seems you also need to wire the notebook icon `ocnotebook` up somehow (and technically also `ocnb`, though it will be removed eventually). Otherwise the [notes app](https://github.com/opencloud-eu/web-extensions/tree/main/packages/web-app-notes) doesn't render the correct icon:

image

AlexAndBear:
> It seems you also need to wire the notebook icon `ocnotebook` up somehow (and technically also `ocnb`, though it will be removed eventually). Otherwise the [notes app](https://github.com/opencloud-eu/web-extensions/tree/main/packages/web-app-notes) doesn't render the correct icon:
> 
> image

does it make sense, to proceed with that, when we actually move the notes app into the web project ? (which is on our list anyways) 🤔

JammingBen:
> does it make sense, to proceed with that, when we actually move the notes app into the web project ? (which is on our list anyways) 🤔

But this doesn't affect the icon loading, it will still be broken once we moved the app. Because it's part of `mapping.extension`, but needs to be in `mapping.folderExtension`? Haven't looked too deep, just a guess.

AlexAndBear:
When

> > does it make sense, to proceed with that, when we actually move the notes app into the web project ? (which is on our list anyways) 🤔
> 
> But this doesn't affect the icon loading, it will still be broken once we moved the app. Because it's part of `mapping.extension`, but needs to be in `mapping.folderExtension`? Haven't looked too deep, just a guess.

Afaik the folder icon will be registered via the index.ts of the individual app.

e.G

```
  extensions: [
        {
          extension: VAULT_EXTENSION,
          type: 'folder',
          icon: 'resource-type-vault',
          iconFillType: 'fill'
        }
      ]
```

So IHMO its not necessary to release a version of the notebook, just to add this, when we move it anyways, or do I miss something here 🤔  ?

JammingBen:
We could also make it work this way as well, yes. But then we should still remove these notebook icons from the icon mapping, no? Because having them in `mapping.extension` doesn't work for folders anymore.

AlexAndBear:
> We could also make it work this way as well, yes. But then we should still remove these notebook icons from the icon mapping, no? Because having them in `mapping.extension` doesn't work for folders anymore.

Yes I think we should remove them, there should be only one source of truth

JammingBen (approved):
👍

PR #3411, opened 2026-09-19
Always sort folders above files

AlexAndBear:
## Description

Folders now always come first when sorting by name. Before, the folder and file groups were concatenated in sort direction, so descending put files first and folders at the bottom. The direction now only orders within each group.

## Related Issue

- Works towards https://github.com/opencloud-eu/web/issues/3381

## How Has This Been Tested?

- test environment: local, `pnpm test:unit --run`, `pnpm check:types`, `pnpm lint`, `pnpm format:check`
- test case 1: ascending by name keeps folders on top
- test case 2: descending by name keeps folders on top, reversed within each group

## Types of changes

- [x] Bugfix


AlexAndBear:
@kulmann @tbsbdr needs PM decision

kulmann (approved):
Yep, that's a good finding. Not preserving the two groups would probably be received as a bug since there's no other file cloud which flips the two groups when reversing the sort order.

PR #3413, opened 2026-09-21
Create a space with quota, subtitle, description, image and members

### LINKED ISSUE #1188: Advanced space creation
### Initial Checklist

### User Story

As a power user, I want to create a space with all details so that I don't have to configure everything after space creation.

### User Value

Complex space creation

### Acceptance Criteria

- A space can be created by only providing the title
- The modal has an `Options` raw-button (same as for public links) which shows more fields in the space creation modal
- additional fields:
  - quota
  - subtitle (text-input)
  - description (text-area)
  - image (upload with cropping, or an emoji through the existing icon picker)
  - members
- error handling:
  - note: this can (and most likely will) happen in multiple requests
  - we don't roll back the entire space creation if something fails (like e.g. uploading the image)
  - instead we only show a summary of what failed in the space creation confirmation toast

AlexAndBear:
## Description

Create a space with everything already set, instead of configuring it afterwards. The name-only dialog stays as it was — an `Options` toggle (same pattern as public links) reveals the rest.

**Added**

- Options panel with two collapsible groups: presentation (image, subtitle, description) and expert (quota, members).
- `CreateSpaceOptions.vue`, `SpaceImagePicker.vue` (pick + crop), `SpaceMemberSelect.vue` (recipient autocomplete + role).
- `useCreateSpace` applies the options once the drive exists: `setSpaceReadme`, `setSpaceImage`, `applyMembers`.
- Per-step error handling: each step runs on its own and failures are collected into one summary toast, so a space that was created is never rolled back because e.g. its image failed.
- `useSpaceImage` — saves a cropped image on an existing space and reports the outcome.
- `useCollaboratorAutocomplete` — the debounced recipient search, shared with the sidebar invite form.

**Changed**

- `ModalWrapper` renders the whole modal stack instead of only the active one, so the cropper can open over the create-space modal without resetting it.
- `OcModal` gained `active` (only the topmost dims the background and holds the focus trap) and per-instance title ids, so stacked dialogs don't share one `aria-labelledby` target.
- `OcDrop` outranks the modal layer and hands focus back to its toggle on close.
- `InviteCollaboratorForm` uses the new autocomplete composable — behaviour unchanged, ~125 lines lighter.
- A modal's body is now the scroll container, so its header and footer stay visible with long content.
- `SpaceImageModal` and `AvatarUpload` use `@vueuse/core`'s `useObjectUrl` instead of a hand-rolled one; the space-image actions delegate their error reporting to `useSpaceImage`.
- "Edit description" no longer writes an empty readme over an existing `.space/readme.md` that the drive doesn't have registered.
- Member invites go through the shared `concurrentRequests.shares.create` queue.
- `design-system` gained `uuid` for the per-instance modal title ids.

## Related Issue

- Fixes https://github.com/opencloud-eu/web/issues/1188

## How Has This Been Tested?

- test environment: unit tests (`pnpm check:all`), 4165 passing — incl. new specs for `useSpaceImage`, `SpaceMemberSelect`, `CreateSpaceModal`, `OcDrop` and `ModalWrapper`
- test case 1: create a space with name only — unchanged behaviour
- test case 2: create a space with quota, subtitle, description, image and members set
- test case 3: a failing step (e.g. image over quota) reports in the summary toast and the space is still created
- test case 4: open the cropper from the options panel — the create-space modal keeps its input
- test case 5: pick members, enable end-to-end encryption, continue, go back — members are still shown
- test case 6: "Edit description" on a space with an unregistered `.space/readme.md` opens it instead of emptying it
- test case 7: a modal with long content scrolls its body while header and footer stay visible

## Types of changes

- [ ] Bugfix
- [x] Breaking change (a modification that affects current functionality)


JammingBen:
Love it!

A few UI things I noticed before looking into the code:

The modal for creating files has a vertical scrollbar now:

image

The modal for creating spaces now has a large margin top & bottom, especially compared to before and the folder modal:

image

packages/web-pkg/src/composables/spaces/useCreateSpace.ts
```diff
@@ -41,52 +73,261 @@ export const useCreateSpace = () => {
...
+        // The space exists and is a vault, only its secret didn't get
+        // committed - saying "failed to create" would be a lie.
+        console.error(error)
+        showErrorMessage({
+          title: $gettext('»%{space}« was created, but its password was not saved', {
+            space: name
+          }),
+          errors: [error]
+        })
+        return createdSpace
```
JammingBen: Is early returning correct here? 🤔 I think we should still add members and apply quota, which happens down below.

packages/web-runtime/src/components/ModalWrapper.vue
```diff
@@ -43,56 +45,73 @@
...
 
 const modalStore = useModals()
-const { activeModal: modal } = storeToRefs(modalStore)
+const { modals } = storeToRefs(modalStore)
 const { updateModal, removeModal } = modalStore
 
-const customComponentRef = ref<CustomModalComponentInstance>()
+// Every modal in the stack stays mounted, so the components below the topmost
+// one keep their state while it is open.
+const customComponentRefs = ref<Record<string, CustomModalComponentInstance>>({})
```
JammingBen: Can't this be a plain `Map`? Nothing renders from it, I think a ref is unnecessary.

JammingBen (approved):

